PR #4, opened 2023-02-09
[FLINK-30996] Sync Kafka 1.17 commits from apache/flink repo to flink-connector-kafka

MartijnVisser:


.github/workflows/push_pr.yml
```diff
@@ -25,4 +25,4 @@ jobs:
   compile_and_test:
     uses: apache/flink-connector-shared-utils/.github/workflows/ci.yml@ci_utils
     with:
-      flink_version: 1.16.0
+      flink_version: 1.17-SNAPSHOT
```
zentol, line 28: Does it genuinely require a snapshot version again? :cry:

flink-connector-kafka/src/test/resources/archunit.properties
```diff
@@ -17,7 +17,7 @@
 #
 
 # By default we allow removing existing violations, but fail when new violations are added.
-freeze.store.default.allowStoreUpdate=true
+#freeze.store.default.allowStoreUpdate=true
```
zentol, line 20: revert

tzulitai (commented):
I've done the following checks so far:

- Diffed all main code files (except pom.xml files) of the externalized modules in this PR, against the corresponding modules in `apache/flink:release-1.17.0-rc1`. Found only the following difference: `apache/flink:release-1.17.0-rc1` seems to contain some test classes `KafkaResource`, `KafkaResourceFactory` etc. that no longer exists in this PR. I don't think those classes are actually used anywhere, so probably is fine?

- Eyeballed the pom.xml changes, and didn't see anything out of the ordinary. Kafka version also matches with `release-1.17.0-rc1`.

- Kafka connectors didn't need special handling of the LICENSE / NOTICE files, so that looks good on this side as well.

I _think_ that covers most of the important aspects?
@MartijnVisser would you suggest me to double check anything else for us to confidently approve this?

MartijnVisser:
> @MartijnVisser would you suggest me to double check anything else for us to confidently approve this?

I think you've covered everything. We usually compared the commits to see that all code that needed to exist actually existed + the necesary LICENSE/NOTICE stuff. And of course CI needed to pass

tzulitai:
@MartijnVisser 

+1 LGTM then :)

tzulitai (approved):

PR #6, opened 2023-03-10
[FLINK-31305] Propagate producer exceptions outside of mailbox execut…

mas-chen:
…or so that checkpoints can correctly fail

-->

Fixes a regression when the checkpoint completes although there is an error. The exception is never thrown before checkpoint completing because the exception throwing logic is enqueued to separate task, during checkpointing.

## Brief change log

- Propagate exception and throw them in write/flush/close methods.

## Verifying this change

- Added unit tests and ran existing unit tests, integration tests, and e2e tests.

  - The serializers: no
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Kubernetes/Yarn, ZooKeeper: yes

## Documentation


mas-chen:
closing due to https://github.com/apache/flink/pull/22150. will sync afterwards

PR #7, opened 2023-03-12
[FLINK-31408] Add support for EOS delivery-guarantee in upsert-kafka

Gerrrr:
* add sink.delivery-guarantee and sink.transactional-id-prefix options to upsert-kafka
* fix the default isolation.level in kafka connector documentation
* add update upsert-kafka connector documentation


tzulitai:
hey @Gerrrr, thanks for opening this PR!

In addition to these changes, I think we have to let `ReducingUpsertSink` implement `TwoPhaseCommittingSink` as well.

Explanation:
`TwoPhaseCommittingSink` actually serves as a "tag" interface which the underlying operator (`SinkWriterOperator`) would check (via `instanceof`) to see if it needs to execute 2PC logic on the sink writers, such as `prepareCommit` and emitting committables to a downstream `CommitterOperator`.

Now, if sink buffering is enabled, the actual `KafkaSink` would be wrapped within a `ReducingUpsertSink`. What this means is that, if we don't implement `TwoPhaseCommittingSink` on the `ReducingUpsertSink`, then under EOS + sink buffering enabled, there will be no commits happening for the 2PC protocol because the `SinkWriterOperator` doesn't know that it is a two-phase committing sink.

tzulitai (changes requested):
hey @Gerrrr, thanks for opening this PR!

In addition to these changes, I think we have to let `ReducingUpsertSink` implement `TwoPhaseCommittingSink` as well.

Explanation:
`TwoPhaseCommittingSink` actually serves as a "tag" interface which the underlying operator (`SinkWriterOperator`) would check (via `instanceof`) to see if it needs to execute 2PC logic on the sink writers, such as `prepareCommit` and emitting committables to a downstream `CommitterOperator`. For example, take a look at [here](https://github.com/apache/flink/blob/d336dad93459622be2dd619cc1d8d3709118628b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/operators/sink/SinkWriterOperator.java#L113) and [here](https://github.com/apache/flink/blob/d336dad93459622be2dd619cc1d8d3709118628b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/operators/sink/SinkWriterOperator.java#L122) and [here](https://github.com/apache/flink/blob/d336dad93459622be2dd619cc1d8d3709118628b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/operators/sink/SinkWriterOperator.java#L187-L196).

Now, if sink buffering is enabled, the actual `KafkaSink` would be wrapped within a `ReducingUpsertSink`, which as of this PR isn't implementing `TwoPhaseCommittingSink`. What this means is that under EOS + sink buffering enabled with this PR, there will be no commits happening for the 2PC protocol simply because the `SinkWriterOperator` doesn't know that it is a two-phase committing sink. i.e. you'll never see any data downstream in `READ_COMMITTED` mode.

Gerrrr:
@tzulitai This PR is ready for review. Can you please take a look when you get a chance?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/TwoPhaseCommittingStatefulSink.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * <p>The purpose of this interface is to be able to pass an interface rather than a {@link
+ * KafkaSink} implementation into the reducing sink which simplifies unit testing.
+ *
+ * @param <InputT> The type of the sink's input
+ * @param <WriterStateT> The type of the sink writer's state
+ * @param <CommT> The type of the committables.
+ */
+@Internal
+public interface TwoPhaseCommittingStatefulSink<InputT, WriterStateT, CommT>
```
tzulitai, line 21: I'm a bit torn whether or not we should just accept this awkwardness and move on. I understand the need for this "tag" mixin interface, but it just feels really odd.

Interface mixins in Java really doesn't play well with the decorator pattern 😕

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/TwoPhaseCommittingStatefulSink.java
```diff
@@ -0,0 +1,34 @@
...
+import org.apache.flink.api.connector.sink2.TwoPhaseCommittingSink;
+
+import java.io.IOException;
+import java.util.Collection;
+
+/**
+ * A combination of {@link TwoPhaseCommittingSink} and {@link StatefulSink}.
+ *
+ * <p>The purpose of this interface is to be able to pass an interface rather than a {@link
+ * KafkaSink} implementation into the reducing sink which simplifies unit testing.
```
tzulitai, line 14: I guess it's probably not only for unit testing? Without the mixin interface the `ReducingUpsertSink` and `ReducingUpsertWriter` wouldn't be able to properly wrap the KafkaSink / KafkaWriter delegates.

tzulitai (commented):
Thanks @AlexAxeman! The changes do make sense and overall looks good, although I'm not so sure of the additional `TwoPhaseCommittingStatefulSink` interface approach to deal with the unit testing complication.

I also don't have a better idea for that at the moment, though. Please allow me to think about this a bit before we proceed to merge.

tzulitai:
@Gerrrr as discussed offline, I think there really isn't a better way around the current PR approach. I'll address the comments on the doc and proceed to merge this.

Thank you for your contribution!

PR #9, opened 2023-03-22
[FLINK-31559][connector/kafka] Update the flink version to 1.18-SNAPSHOT and drop support for 1.16

ruanhang1993:
This PR updates the flink version to 1.18-SNAPSHOT.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -26,7 +26,7 @@ jobs:
     if: github.repository_owner == 'apache'
     strategy:
       matrix:
-        flink: [1.16-SNAPSHOT, 1.17-SNAPSHOT]
+        flink: [1.16-SNAPSHOT, 1.17-SNAPSHOT, 1.18-SNAPSHOT]
     uses: apache/flink-connector-shared-utils/.github/workflows/ci.yml@ci_utils
     with:
       flink_version: ${{ matrix.flink }}
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -26,7 +26,7 @@ jobs:
     if: github.repository_owner == 'apache'
     strategy:
       matrix:
-        flink: [1.16-SNAPSHOT, 1.17-SNAPSHOT, 1.18-SNAPSHOT]
+        flink: [1.17-SNAPSHOT, 1.18-SNAPSHOT]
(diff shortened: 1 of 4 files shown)


.github/workflows/weekly.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     if: github.repository_owner == 'apache'
     strategy:
       matrix:
-        flink: [1.16-SNAPSHOT, 1.17-SNAPSHOT]
+        flink: [1.16-SNAPSHOT, 1.17-SNAPSHOT, 1.18-SNAPSHOT]
```
MartijnVisser, line 29: If we add support for 1.18, then we should drop support for 1.16 per https://cwiki.apache.org/confluence/display/FLINK/Externalized+Connector+development#ExternalizedConnectordevelopment-Flinkcompatibility

MartijnVisser (changes requested):
We should drop support for 1.16 if we add support for 1.18

.github/workflows/weekly.yml
ruanhang1993: Thanks for reminding, @MartijnVisser. Drop 1.16 now.

MartijnVisser (commented):
@ruanhang1993 The CI fails, so this can't be merged atm

ruanhang1993:
@MartijnVisser  There is some compile errors . I fixed it.

tzulitai:
I'm not so sure we should merge this.

I believe we want to always develop against the latest stable Flink version, even in `main` branch.
For example in `main` right now, we're using `1.17.0` for the Flink version.

For `1.18-SNAPSHOT`, we rely on nightly builds to ensure we're still compatible with the latest snapshot.

Not entirely sure what the consensus was here, @MartijnVisser can you confirm?

MartijnVisser:
> I believe we want to always develop against the latest stable Flink version, even in `main` branch.

@tzulitai Ideally yes, but there can be situations where we want a connector to have an implementation for a feature that isn't yet released in Flink. I think that's the case in this situation. If we don't rely on the SNAPSHOT build, we can only start developing the connector implementation when Flink and that new API is released.

MartijnVisser:
This has already been resolved, as shown in this nightly build https://github.com/apache/flink-connector-kafka/actions/runs/6486214259

PR #16, opened 2023-03-24
[FLINK-30859] Externalize confluent avro related code

mas-chen:
1st major part of FLINK-30859 to externalize confluent Avro related code to remove flink-connector-kafka code safely from apache/flink:master


flink-formats-confluent/flink-avro-confluent-registry/pom.xml
```diff
@@ -135,30 +135,36 @@ under the License.
...
 		</dependency>
 		<dependency>
 			<groupId>org.apache.flink</groupId>
 			<artifactId>flink-avro</artifactId>
-			<version>${project.version}</version>
+			<version>${flink.version}</version>
 			<scope>test</scope>
 			<type>test-jar</type>
 		</dependency>
+		<dependency>
```
mas-chen, line 162: required for unit test

mas-chen:
BTW I noticed that the NOTICE files are outdated. I hope you don't mind that I change it in the same PR here: [a554982](https://github.com/apache/flink-connector-kafka/pull/16/commits/a55498297a3d20c0ccaf078cd8629ea22e23f820)

docs/content.zh/docs/connectors/table/formats/avro-confluent.md
tzulitai, line 1: Does the "Debezium" format docs need to be migrated here as well?

flink-formats-confluent/flink-avro-confluent-registry/pom.xml
```diff
@@ -0,0 +1,184 @@
...
+		<version>4.0-SNAPSHOT</version>
+	</parent>
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>flink-avro-confluent-registry</artifactId>
+
+	<name>Flink : Formats : Avro confluent registry</name>
+
+	<properties>
+		<kafka.version>3.2.3</kafka.version>
```
tzulitai, line 35: The `main` branch recently had this upgraded to 3.4.0 for `flink-connector-kafka`.
We'd probably want this version to be shared across all modules.

flink-formats-confluent/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+  <parent>
+    <groupId>org.apache.flink</groupId>
+    <artifactId>flink-connector-kafka-parent</artifactId>
+    <version>4.0-SNAPSHOT</version>
+  </parent>
+
+  <packaging>pom</packaging>
+
+  <artifactId>flink-formats-confluent</artifactId>
+  <name>Flink : Formats : Confluent</name>
```
tzulitai, line 34: Not so sure of the naming. Should we name this `Flink : Formats : Kafka` instead?
I know that the current formats we have is all Confluent-specific, but naming it more generally might make sense if we add other Kafka-related formats down the road.

flink-formats-confluent/flink-avro-confluent-registry/src/main/java/org/apache/flink/formats/avro/registry/confluent/debezium/DebeziumAvroFormatFactory.java
tzulitai, line 1: I'm not so sure but: this makes me wonder if `debezium json` should be moved here as well.
Strictly speaking its tied to Kafka as well. But it's a bit awkward, since that format is bundled as part of `flink-json` ...

flink-formats-confluent/pom.xml
tzulitai: For example, the debezium formats are not strictly tied to Confluent concepts.

flink-formats-confluent/flink-avro-confluent-registry/src/main/java/org/apache/flink/formats/avro/registry/confluent/debezium/DebeziumAvroFormatFactory.java
mas-chen: It is awkward. I think it makes sense and the Debezium Flink documentation does covers both Avro + Json so we should port both over

flink-formats-confluent/flink-avro-confluent-registry/pom.xml
mas-chen: Good catch! Same with the Confluent version

flink-formats-confluent/flink-avro-confluent-registry/src/main/java/org/apache/flink/formats/avro/registry/confluent/debezium/DebeziumAvroFormatFactory.java
tzulitai: Do you intend to port all of `flink-json` over or just the Debezium parts?

tzulitai: 👍 makes sense

mas-chen: I took all of the Flink-json and deleted the irrelevant code, keeping only debezium related code. I found that to be an easier approach when sync'ing the commits

flink-formats-kafka/flink-json-debezium/pom.xml
```diff
@@ -0,0 +1,141 @@
...
+
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>org.apache.flink</groupId>
+		<artifactId>flink-formats-kafka</artifactId>
+		<version>4.1-SNAPSHOT</version>
+	</parent>
+
+	<artifactId>flink-json-debezium</artifactId>
```
zentol, line 31: I'm wondering if there is a significant benefit to hanving this as a separate module.

for backwards-compatibility alone we'd need to bundle it in the kafka-connector (as it was previously provided by the distribution by default), and with no additional dependencies being added I don't see a real downside.

tzulitai: re: backwards compatibility

I also just realized that the Flink SQL Client bundles flink-json as well. So we can't really separate the debezium-json formats out of flink-json ... otherwise it becomes a behavior breaking change (users suddenly have to explicitly include this new module why previously it just works out of the box).

mas-chen: Gordon and I discussed offline. For the purposes of finishing off the externalization and externalizing the required confluent avro, we will table this discussion to outside of this PR. I filed https://issues.apache.org/jira/browse/FLINK-31747 to track this

flink-confluent-schema-registry-e2e-tests/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+	</repositories>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-streaming-java</artifactId>
+			<version>${flink.version}</version>
+			<scope>provided</scope>
+		</dependency>
+		<!-- https://mvnrepository.com/artifact/org.apache.flink/flink-connector-kafka -->
```
tzulitai, line 50: This comment can be removed.

flink-formats-kafka/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+  <parent>
+    <groupId>org.apache.flink</groupId>
+    <artifactId>flink-connector-kafka-parent</artifactId>
+    <version>4.1-SNAPSHOT</version>
+  </parent>
+
+  <packaging>pom</packaging>
+
+  <artifactId>flink-formats-kafka</artifactId>
+  <name>Flink : Formats : Confluent</name>
```
tzulitai, line 34: This should be renamed to `Flink : Formats : Kafka` as well.

tzulitai (approved):
+1 LGTM 🎉 I did another pass on the pom.xmls and new module structures.

Two remaining minor comments.

PR #17, opened 2023-03-26
[FLINK-31620] Flush wrapped writer

Gerrrr:
- with the solution, the actual flushing only still happens on checkpoints
- when buffer size is full or periodically triggered, we don't actually do flushing on the wrapped writer.

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/ReducingUpsertWriter.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/ReducingUpsertWriter.java
@@ -87,7 +87,8 @@ class ReducingUpsertWriter<WriterState>
 
     @Override
     public void flush(boolean endOfInput) throws IOException, InterruptedException {
-        flush();
+        sinkBuffer();
+        wrappedWriter.flush(endOfInput);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


Gerrrr:
@tzulitai Do we need to port this patch to 1.16 and 1.17?

tzulitai:
@Gerrrr this looks like a bug, so likely yes. Lets only open the backport PRs after this is reviewed and merged.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/ReducingUpsertWriter.java
```diff
@@ -140,7 +141,7 @@ private RowData changeFlag(RowData value) {
         return value;
     }
 
-    private void flush() throws IOException, InterruptedException {
+    private void flushBuffer() throws IOException, InterruptedException {
```
tzulitai, line 144: nitpick: I wonder if `sinkReducedRecordsInBuffer` is a better name here, since we're not actually flushing the wrapped writer. Just to make it semantically clear that we're still only flushing on checkpoints.

tzulitai (commented):
Fix looks good, just one minor comment. Thanks @Gerrrr!

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/ReducingUpsertWriter.java
Gerrrr: Done in [5847885](https://github.com/apache/flink-connector-kafka/pull/17/commits/58478851fd4cf9b33ce352ea2a01d89cf995f170).

tzulitai:
+1 LGTM, merging.

PR #18, opened 2023-03-27
[FLINK-31049] [flink-connector-kafka]Add support for Kafka record headers to KafkaSink

AlexAxeman:

The default `org.apache.flink.connector.kafka.sink.KafkaSink` does not support adding Kafka record headers when using KafkaRecordSerializationSchemaBuilder, which is the most convenient way to create a Kafka sink. This PR adds support for Kafka headers to `KafkaRecordSerializationSchemaBuilder`.

## Brief change log

  - *Implemented a `HeaderProducer` that allows creating `Header`s from the input element*
  - *Added setters to `KafkaRecordSerializationSchemaBuilder` to allow setting a `HeaderProducer`*
  - *Added an optional `HeaderProducer` constructor argument to `KafkaRecordSerializationSchemaWrapper` that now uses a `ProducerRecord` constructor that includes the headers.* 

## Verifying this change

Please make sure both new and modified tests in this PR follows the conventions defined in our code quality guide: https://flink.apache.org/contributing/code-style-and-quality-common.html#testing

  - *Added tests to `KafkaRecordSerializationSchemaBuilderTest`*

  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: yes
  - The serializers: no

## Documentation

  - Does this pull request introduce a new feature? yes?
  - If yes, how is the feature documented? JavaDocs

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/HeaderProducer.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/HeaderProducer.java
@@ -21,6 +21,8 @@ package org.apache.flink.connector.kafka.sink;
 import org.apache.flink.annotation.PublicEvolving;
 
 import org.apache.kafka.common.header.Header;
+import org.apache.kafka.common.header.Headers;
+import org.apache.kafka.common.header.internals.RecordHeaders;
 
 import java.io.Serializable;
 import java.util.ArrayList;
@@ -28,7 +30,7 @@ import java.util.ArrayList;
 /** Creates an {@link Iterable} of {@link Header}s from the input element. */
 @PublicEvolving
 public interface HeaderProducer<IN> extends Serializable {
-    default Iterable<Header> produceHeaders(IN input) {
-        return new ArrayList<>();
+    default Headers produceHeaders(IN input) {
+        return new RecordHeaders(new ArrayList<>());
     }
 }
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/HeaderProducer.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/HeaderProducer.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.connector.kafka.sink;
+
+import org.apache.flink.annotation.PublicEvolving;
+
+import org.apache.kafka.common.header.Header;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+
+/** Creates an {@link Iterable} of {@link Header}s from the input element. */
+@PublicEvolving
+public interface HeaderProducer<IN> extends Serializable {
+    default Iterable<Header> produceHeaders(IN input) {
+        return new ArrayList<>();
+    }
+}
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilderTest.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilderTest.java
@@ -28,12 +28,14 @@ import org.apache.flink.shaded.guava30.com.google.common.collect.ImmutableMap;
 
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.common.Configurable;
+import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.serialization.Deserializer;
 import org.apache.kafka.common.serialization.StringDeserializer;
 import org.apache.kafka.common.serialization.StringSerializer;
 import org.junit.Before;
 import org.junit.Test;
 
+import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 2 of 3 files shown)


AlexAxeman:
@MartijnVisser @tzulitai

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/HeaderProducer.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.apache.kafka.common.header.Header;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+
+/** Creates an {@link Iterable} of {@link Header}s from the input element. */
+@PublicEvolving
+public interface HeaderProducer<IN> extends Serializable {
+    default Iterable<Header> produceHeaders(IN input) {
```
RamanVerma, line 31: I think you can use `Headers` instead of `Iterable<Header>` from `org.apache.kafka.common.header.Headers`

RamanVerma (approved):
thanks for the change, @AlexAxeman 
LGTM

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/HeaderProducer.java
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.kafka.common.header.Header;
+import org.apache.kafka.common.header.Headers;
+import org.apache.kafka.common.header.internals.RecordHeaders;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+
+/** Creates an {@link Iterable} of {@link Header}s from the input element. */
+@PublicEvolving
+public interface HeaderProducer<IN> extends Serializable {
```
tzulitai, line 32: nit: Would `HeaderAssigner` be a better name for this? `Producer` already has a meaning within the Kafka-namespace, perhaps it can cause some confusion. WDYT?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/HeaderProducer.java
```diff
@@ -0,0 +1,36 @@
...
+    default Headers produceHeaders(IN input) {
```
tzulitai, line 33: nit: if we rename the interface to `HeaderAssigner`, then this should be renamed to `assignHeaders` as well.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/HeaderProducer.java
```diff
@@ -0,0 +1,36 @@
...
+        return new RecordHeaders(new ArrayList<>());
```
tzulitai, line 34: Can we have a default implementation that doesn't create an empty `ArrayList` and `RecordHeaders` object on every input record?

tzulitai (changes requested):
Thanks for the PR! Makes total sense to add this.

Have one comment about the default behavior (no headers), and a nitpick about naming.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/HeaderProducer.java
AlexAxeman: Yeah, thought about that. So it should be a `Collections.emptyList()`?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/HeaderProducer.java
tzulitai: > it does not assign the headers, but produces them

In that case, how about `HeaderProvider`? `Provider` seems to be a more canonical name for this.

And correspondingly, the single method would be called `getHeaders(IN input)`.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/HeaderProducer.java
tzulitai: I was thinking about `Optional<Headers> getHeaders(IN input)`, and default is returning `Optional.empty()`.

I'm also wondering if it's possible / makes sense to NOT expose the Kafka `Headers` class on the public API. Although Kafka API seems to be rather stable, in general its best if we can avoid exposing our dependencies' classes on the public interface.

AlexAxeman: Hm... Personally I don't really like that. The big thing about Optionals is to force the handling of possible `null` values. You don't need that protecting while returning collections, as you can always return an empty one, which is what I did. I'd much rather go for an empty (immutable) collection.

Initially, I returned a `Iterable<Header>` and changed that to kafka API Headers upon request. I'm not really leaning to either side, so happy to change that back @RamanVerma, wdyt?

tzulitai: @AlexAxeman see my latest comment above about the default case being `headerProvider` is null.

In that case, we no longer need / want to return `Optional`.

i.e.

```
interface HeaderProvider {
    Headers getHeaders(IN input);
}
```

and in `KafkaRecordSerializationSchemaWrapper#serialize()`:
```
if (headerProvider != null) {
    return new ProducerRecord(
      ...,
      headerProvider.getHeaders(input)
    )
} else (
    return new ProducerRecord() // no headers
)
```

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilderTest.java
```diff
@@ -86,6 +89,16 @@ public void testDoNotAllowMultipleTopicSelector() {
                 .isInstanceOf(IllegalStateException.class);
     }
 
+    @Test
+    public void testDoNotAllowMultipleHeaderProducers() {
```
tzulitai, line 93: I don't think you really need this restriction for the headers.

The key / value serializer setters had the check because there were different entry points for setting the serialization schema, i.e. either via `setKafkaXSerializer` or `setKeySerializationSchema`. We don't want the user to be setting the key serializer through both entrypoints, hence the check.

This isn't a problem for the header provider setting. Same goes for the Kafka partitioner, for example - we don't check multiple sets there.

tzulitai (changes requested):

tzulitai:
Merged via https://github.com/apache/flink-connector-kafka/commit/a7785630e714af303b224c38d9a6caa89a551265

I squashed my changes along with your comment. Please let me know if you find anything odd and would like me to address. Thanks again!

AndersSpringborg:
In which release is this being published in? I cannot see it in the latest release, [3.0.0-1.17](https://mvnrepository.com/artifact/org.apache.flink/flink-connector-kafka/3.0.0-1.17)

Sandrock431:
We're eagerly awaiting this change as well, and it's not in `3.0.2-1.18` either.

AlexAxeman:
@tzulitai Do you know what happened with this change?

PR #19, opened 2023-03-28
[Sync apache/flink:master][FLINK-31305] Propagate producer exceptions outside of mailbox execut…

mas-chen:
…or so that checkpoints can correctly fail

This closes #22150.


tzulitai:
Thanks @mas-chen for porting this over.

Will merge once CI is green.

PR #20, opened 2023-03-30
[FLINK-29398][connector/kafka] Provide rack ID to Kafka Source to take advantage of Rack Awareness

jeremy-degroot:

This PR adds a new method to KafkaSourceBuilder that sets the `client.id` config for the KafkaSource to the value returned by the provided Supplier. It needs to be a Supplier because it needs to run on the TaskManager, and can't be determined at Job submit time like other configs.

## Brief change log

  - *Add setRackId to KafkaSourceBuilder*
  - *Plumb rackId into KafkaPartitionSplitReader*
  - *Add rack id tests*
  -  *Document RackId feature*

## Verifying this change
  - *Added tests for the KafkaSplitReader that verify behaviors for null rackId Supplier, null and empty return values, and provided values.*
  - *Manually verified the change by running a 3-node cluster that covered two "racks" (AWS Availability Zones) against an Amazon MSK cluster.*

  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: yes
  - The serializers: no

## Documentation

  - Does this pull request introduce a new feature? yes
  - If yes, how is the feature documented? docs/Javadocs


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -80,11 +80,20 @@ public KafkaPartitionSplitReader(
...
+            Properties props,
+            SourceReaderContext context,
+            KafkaSourceReaderMetrics kafkaSourceReaderMetrics,
+            Supplier<String> rackIdSupplier) {
         this.subtaskId = context.getIndexOfSubtask();
         this.kafkaSourceReaderMetrics = kafkaSourceReaderMetrics;
         Properties consumerProps = new Properties();
         consumerProps.putAll(props);
         consumerProps.setProperty(ConsumerConfig.CLIENT_ID_CONFIG, createConsumerClientId(props));
+        setConsumerClientRack(consumerProps, rackIdSupplier);
```
RamanVerma, line 96: Can this be resolved in the caller (KafkaSource) and `ConsumerConfig.CLIENT_RACK_CONFIG` can be set there. The benefit would be that you do not need to change this file.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReaderTest.java
```diff
@@ -319,6 +321,51 @@ public void testUsingCommittedOffsetsWithEarliestOrLatestOffsetResetStrategy(
...
+    @NullAndEmptySource
+    public void testSetConsumerClientRackIgnoresNullAndEmpty(String rackId) {
+        Properties properties = new Properties();
+        Supplier<String> rackIdSupplier = () -> rackId;
+        KafkaPartitionSplitReader reader =
+                createReader(
+                        properties,
+                        UnregisteredMetricsGroup.createSourceReaderMetricGroup(),
+                        rackIdSupplier);
+        reader.setConsumerClientRack(properties, rackIdSupplier);
```
RamanVerma, line 351: Supplier function should get called from `KafkaPartitionSplitReader` constructor. Do you need to call this method. Same for line 365

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReaderTest.java
```diff
@@ -319,6 +321,51 @@ public void testUsingCommittedOffsetsWithEarliestOrLatestOffsetResetStrategy(
...
+        assertThat(properties.containsKey(ConsumerConfig.CLIENT_RACK_CONFIG)).isFalse();
+    }
+
+    @Test
+    public void testSetConsumerClientRackUsesCorrectParameter() {
```
RamanVerma, line 356: This test can be combined with the first one if you add the assertion from line 366 to after line 338

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReaderTest.java
RamanVerma: If this call is not needed, you can make the implementation of `setConsumerClientRack` private.

RamanVerma (commented):
thanks for the change  @jeremy-degroot, I have made some minor comments

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReaderTest.java
jeremy-degroot: As far as I know, neither the KafkaConsumer nor the KafkaPartitionSplitReader offers any way to examine the final consumer Properties, so to test the behavior of setConsumerClientRack we have to test it directly.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
jeremy-degroot: I looked into this, and while it's certainly possible to do it that way the testing path is much more complex. Since the Supplier would have to be resolved in another Supplier<KafkaPartitionSplitReader> that's passed to the Reader, testing that the behavior is as expected in the actual execution path is difficult. In KafkaPartitionSplitReader, we can call the constructor directly and verify that the rackIdSupplier is called, and then also verify it does what we need it to by verifying the behavior of the helper method you noted further down.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReaderTest.java
jeremy-degroot: You're right, I'll be pushing an update to the branch to do that shortly.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSource.java
```diff
@@ -98,20 +98,24 @@
     private final KafkaRecordDeserializationSchema<OUT> deserializationSchema;
     // The configurations.
     private final Properties props;
+    // Client rackId callback
+    private final Supplier<String> rackIdSupplier;
```
mas-chen, line 102: This should be a "SerializableSupplier". A supplier may contain a non-serializable client for example

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
mas-chen: Yeah KafkaSource couldn't resolve the rack since the rack may not be the same for JM and TM. I would suggest to resolve this at the KafkaSourceReader if possible and avoid this low level change.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReaderTest.java
mas-chen: You can expose a method @VisibleForTesting or use reflection?

docs/content/docs/connectors/datastream/kafka.md
```diff
@@ -466,6 +466,28 @@ client dependencies in the job JAR, so you may need to rewrite it with the actua
 For detailed explanations of security configurations, please refer to
 <a href="https://kafka.apache.org/documentation/#security">the "Security" section in Apache Kafka documentation</a>.
 
+## Kafka Rack Awareness
+
+Kafka Rack Awareness allows flink to select and control the cloud region and availability zone configured by the use of rackId, this feature could allow a significant cost reduction in the cloud provider bill and achieve a better networking performance when connecting to closer and more reliable networks.
```
mas-chen, line 471:
```suggestion
Kafka rack awareness allows Flink to select and control the cloud region and availability zone that Kafka consumers read from, based on the Rack ID. This feature reduces network costs and latency since it allows consumers to connect to the closest Kafka brokers, possibly colocated in the same cloud region and availability zone.
```

docs/content/docs/connectors/datastream/kafka.md
```diff
@@ -466,6 +466,28 @@ client dependencies in the job JAR, so you may need to rewrite it with the actua
...
+
+### RackId
+
+setRackId is the variable where the desired or available availability zones get stored. If provided, the Supplier will be run when the consumer is set up on the Task Manager, and the consumer's client.rack configuration will be set to the value.
+
+https://kafka.apache.org/documentation/#consumerconfigs_client.rack
```
mas-chen, line 477: Consider moving this documentation earlier since it defines rack awareness

docs/content/docs/connectors/datastream/kafka.md
```diff
@@ -466,6 +466,28 @@ client dependencies in the job JAR, so you may need to rewrite it with the actua
 For detailed explanations of security configurations, please refer to
 <a href="https://kafka.apache.org/documentation/#security">the "Security" section in Apache Kafka documentation</a>.
 
+## Kafka Rack Awareness
+
+Kafka Rack Awareness allows flink to select and control the cloud region and availability zone configured by the use of rackId, this feature could allow a significant cost reduction in the cloud provider bill and achieve a better networking performance when connecting to closer and more reliable networks.
+
+### RackId
+
+setRackId is the variable where the desired or available availability zones get stored. If provided, the Supplier will be run when the consumer is set up on the Task Manager, and the consumer's client.rack configuration will be set to the value.
```
mas-chen, line 475: This section could less verbose--you can briefly describe how the rackId is set via the supplier without going too deep into the implementation and briefly mention example suppliers.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSourceBuilder.java
```diff
@@ -355,6 +360,17 @@ public KafkaSourceBuilder<OUT> setClientIdPrefix(String prefix) {
         return setProperty(KafkaSourceOptions.CLIENT_ID_PREFIX.key(), prefix);
     }
 
+    /**
+     * Set the clientRackId supplier to be passed down to the KafkaPartitionSplitReader.
+     *
+     * @param rackIdCallback callback to provide Kafka consumer client.rack
+     * @return this KafkaSourceBuilder
+     */
+    public KafkaSourceBuilder<OUT> setRackId(Supplier<String> rackIdCallback) {
```
mas-chen, line 369: How about `setRackIdSupplier`?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSource.java
jeremy-degroot: You're right about the Serializable bit, good catch.

Is there a straightforward way to do that while still allowing for Lambdas? Maybe provide one method for the basic Supplier and another for ClosableSerializableSupplier for when you need that degree of safety.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
jeremy-degroot: I don't think it can be done in KafkaSourceReader either, since that doesn't interact with the Supplier<KafkaPartitionSplitReader> at all. The alternative seems to be to resolve it in KafkaSourceFetcherManager, since that extends SingleThreadFetcherManager, where both Suppliers eventually run. But that would involve a rather messy overriding of the createSplitFetcher() method from that included class, and I don't think that's any better.

pzim:
@jeremy-degroot - this capability is one that we are currently in need of. Any idea when this might be approved/merged/released? Or are there other alternatives we might use to achieve rack awareness in flink?

pzim:
@mas-chen - any chance this will be reviewed/approved soon?

sean-scott-lr:
I need this change as well, and it appears it wont be going into any release any time soon. I tried to extend the classes to add this functionality, but due to the excessive package/private access modifiers its not possible.  Any suggestions on how I may do that without having to fork this project?

tzulitai (commented):
Looks like there's two remaining items to address before we can merge this:

1. The rack ID needs to be at least `Serializable`. `Closeable` variant can probably be added later.
2. Move the lazy configuration of rack id on client properties to be higher in the initialization call stacks - as it is right now, it's happening at too low-level (at the split reader).

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSourceBuilder.java
```diff
@@ -355,6 +360,17 @@ public KafkaSourceBuilder<OUT> setClientIdPrefix(String prefix) {
...
     }
 
+    /**
+     * Set the clientRackId supplier to be passed down to the KafkaPartitionSplitReader.
+     *
+     * @param rackIdCallback callback to provide Kafka consumer client.rack
+     * @return this KafkaSourceBuilder
+     */
+    public KafkaSourceBuilder<OUT> setRackIdSupplier(SerializableSupplier<String> rackIdCallback) {
+        this.rackIdSupplier = rackIdCallback;
```
tzulitai, line 370: nit: can we unify the naming here to just refer to this as the `rackIdSupplier`?

tzulitai (approved):
Thanks @jeremy-degroot! Just two minor nitpicks, I'll address them when I merge this.

Merging now ...

tzulitai:
Actually, @jeremy-degroot would you be able to squash all your changes into one commit that is rebased on latest `main` branch? That would be super helpful.

PR #21, opened 2023-04-04
[hotfix] Correct and/or synchronize multiple dependencies/versions

MartijnVisser:
1. Test CI builds against Flink 1.17.0 instead of Flink 1.17-SNAPSHOT
2. Run the weekly builds against 1.17-SNAPSHOT and 1.18-SNAPSHOT, while dropping 1.16-SNAPSHOT since it's incompatible with this version of the Flink Kafka connector
3. Set the correct groupId for `flink-connector-parent`
4. Remove `archunit` to avoid issues with the weekly builds, given that the version of Archunit can differ between Flink versions. Instead, it will rely on `flink-architecture-tests-test` dependencies.

### DIFF
--- a/.github/workflows/push_pr.yml
+++ b/.github/workflows/push_pr.yml
@@ -25,4 +25,4 @@ jobs:
   compile_and_test:
     uses: apache/flink-connector-shared-utils/.github/workflows/ci.yml@ci_utils
     with:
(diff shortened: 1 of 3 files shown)


tzulitai (approved):
+1 LGTM, thanks @MartijnVisser

PR #22, opened 2023-04-05
[FLINK-31740] [upsert-kafka] Allows setting boundedness for SQL Upsert Kafka Connector

tzulitai:
This is a "follow-up" to previous work in https://github.com/apache/flink/pull/21808 that added boundedness options for the SQL Kafka Connector. This PR does the same for the SQL Upsert Kafka Connector.

Please see new tests in `UpsertKafkaTableITCase` for end-to-end usage examples of the new boundedness options and the expected behaviour.

Most of the added tests in this PR is a copy of what was added for the SQL Kafka Connector in https://github.com/apache/flink/pull/21808.

## Changelog

1. Expose `scan.bounded.*` options for upsert Kafka
2. Add unit tests for source configuration, as well as end-to-end integration tests with boundedness enabled
3. Minor refactorings (tagged `[hotfix]` commits) to share test utilities across the tests for SQL Kafka Connector and SQL Upsert Kafka Connector.

## Testing

Covered in new tests in `UpsertKafkaDynamicTableFactoryTest` and `UpsertKafkaTableITCase`.

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaConnectorOptionsUtil.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaConnectorOptionsUtil.java
@@ -145,7 +145,7 @@ class KafkaConnectorOptionsUtil {
         }
     }
 
-    protected static void validateScanStartupMode(ReadableConfig tableOptions) {
+    private static void validateScanStartupMode(ReadableConfig tableOptions) {
         tableOptions
                 .getOptional(SCAN_STARTUP_MODE)
                 .ifPresent(
@@ -189,7 +189,7 @@ class KafkaConnectorOptionsUtil {
                         });
     }
 
-    protected static void validateScanBoundedMode(ReadableConfig tableOptions) {
+    static void validateScanBoundedMode(ReadableConfig tableOptions) {
         tableOptions
                 .getOptional(SCAN_BOUNDED_MODE)
                 .ifPresent(
--- a/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaConnectorOptionsUtil.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaConnectorOptionsUtil.java
@@ -145,7 +145,7 @@ class KafkaConnectorOptionsUtil {
         }
     }
 
-    private static void validateScanStartupMode(ReadableConfig tableOptions) {
+    protected static void validateScanStartupMode(ReadableConfig tableOptions) {
         tableOptions
                 .getOptional(SCAN_STARTUP_MODE)
                 .ifPresent(
@@ -189,7 +189,7 @@ class KafkaConnectorOptionsUtil {
                         });
     }
 
-    private static void validateScanBoundedMode(ReadableConfig tableOptions) {
+    protected static void validateScanBoundedMode(ReadableConfig tableOptions) {
         tableOptions
                 .getOptional(SCAN_BOUNDED_MODE)
                 .ifPresent(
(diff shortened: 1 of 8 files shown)


tzulitai:
cc @twalthr @dawidwys @Gerrrr for review, thank you :)

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/UpsertKafkaTableITCase.java
```diff
@@ -388,6 +391,201 @@ public void testKafkaSourceSinkWithKeyAndFullValue() throws Exception {
...
+                        + " (3, 102, 'payload 3'),\n"
+                        + " (1, 100, 'payload')";
+        tEnv.executeSql(insertValuesSql).await();
+
+        // results should only have records up to offset=2
+        final List<Row> results = collectAllRows(tEnv.sqlQuery("SELECT * from upsert_kafka"));
+        final List<Row> expected =
+                Arrays.asList(
+                        changelogRow("+I", 1L, 100L, "payload 1"),
+                        changelogRow("+I", 2L, 101L, "payload 2"));
```
tzulitai, line 438: Note: with boundedness set using specific offsets, would there be semantical issues if the end offset happens to be between a `-U` and `+U`?

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaConnectorOptionsUtil.java
```diff
@@ -145,7 +145,7 @@ public static void validateSinkTopic(ReadableConfig tableOptions) {
         }
     }
 
-    private static void validateScanStartupMode(ReadableConfig tableOptions) {
```
Gerrrr, line 148: nit: we can keep this `private` as this PR does not add the scan startup mode to upsert kafka.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaConnectorOptionsUtil.java
```diff
@@ -189,7 +189,7 @@ private static void validateScanStartupMode(ReadableConfig tableOptions) {
                         });
     }
 
-    private static void validateScanBoundedMode(ReadableConfig tableOptions) {
```
Gerrrr, line 192: nit: package-private is probably sufficient here

Gerrrr (approved):
LGTM!

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/UpsertKafkaTableITCase.java
Gerrrr: Doesn't upsert kafka alway insert `I` and `D` and derive `-U` and `+U` on the consumer side via a normalization node?

tzulitai: Ah, yes you're right. Another one of those mind-twists I'm still trying to wrap my head around 😅
So the case I mentioned would never happen then - if the end offset is right after a `+I` that is an update, then the consumer side would normalize it as a `-U` and `+U`. I'll update the test to cover this case.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaConnectorOptionsUtil.java
tzulitai: Changed this by mistake, will revert.

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/UpsertKafkaTableITCase.java
tzulitai: https://github.com/apache/flink-connector-kafka/pull/22/commits/61e919ecca80bd094bcc431c6f298faa52c6f5c0

this new commit answers my confusion :)

PR #23, opened 2023-04-06
[hotfix] Update NOTICE files to reflect year 2023

mas-chen:
This needs to be back ported as well


tzulitai (approved):
Thanks, merging and backporting.

PR #24, opened 2023-04-25
[FLINK-30859] Sync kafka related examples from apache/flink:master

mas-chen:
There is only one example related to Kafka -- StateMachineExample.


flink-examples-kafka/flink-examples-streaming-state-machine-build-helper/src/main/resources/META-INF/NOTICE
```diff
@@ -0,0 +1,9 @@
+flink-examples-streaming-state-machine
+Copyright 2014-2023 The Apache Software Foundation
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+This project bundles the following dependencies under the Apache Software License 2.0. (http://www.apache.org/licenses/LICENSE-2.0.txt)
+
+- org.apache.kafka:kafka-clients:3.2.3
```
tzulitai, line 9: the depended Kafka version is now 3.4.0

tzulitai (commented):
I'm not entirely sure we should move the whole state machine example as is to the externalized Kafka connector repo.

My thinking is the following:
1. The state machine example, for the most part, demonstrates state transitions and Flink managed state.
2. The actual Kafka read / write part of the example almost only feels like a cherry-on-top.

@mas-chen Do you think it would make more sense if we do the following instead?:
1. Modify the state machine example in `apache/flink` to not depend on Kafka.
2. Move only part of the examples that write to Kafka and read from Kafka to `apache/flink-connector-kafka`.

mas-chen:
Discussed offline to refactor to remove Kafka dependency

PR #25, opened 2023-04-27
[FLINK-30859] Sync flink sql client test

mas-chen:
final code to move, before Kafka code can be deleted from master branch

### DIFF
--- a/flink-connector-kafka-e2e-tests/flink-sql-client-test/pom.xml
+++ b/flink-connector-kafka-e2e-tests/flink-sql-client-test/pom.xml
@@ -22,7 +22,7 @@ under the License.
 	<parent>
 		<artifactId>flink-end-to-end-tests</artifactId>
 		<groupId>org.apache.flink</groupId>
-		<version>1.11-SNAPSHOT</version>
+		<version>1.12-SNAPSHOT</version>
 	</parent>
 	<modelVersion>4.0.0</modelVersion>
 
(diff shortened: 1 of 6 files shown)


flink-connector-kafka-e2e-tests/flink-sql-client-test/pom.xml
```diff
@@ -0,0 +1,129 @@
...
+			<artifactId>flink-end-to-end-tests-common-kafka</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+
+		<!-- The following dependencies are for connector/format sql-jars that
+			we copy using the maven-dependency-plugin. When extending the test
+ 			to cover more connectors/formats, add a dependency here and an entry
+			to the dependency-plugin configuration below.
+			This ensures that all modules we actually need (as defined by the
+ 			dependency-plugin configuration) are built before this module. -->
```
tzulitai, line 52: Is this comment still relevant?

tzulitai (approved):
+1, LGTM. Thanks @mas-chen!

tzulitai:
On a second look, it looks like this test does far more stuff than just testing the SQL Kafka Connector ...

Again, like the state machine example PR, I'm not sure we really should be moving the whole e2e test as is, but rather:

1. modify the original E2E in `apache/flink` to not use Kafka
2. Move over only the parts related to Kafka in `apache/flink-connector-kafka`

tzulitai (commented):
On a second look, it looks like this test does far more stuff than just testing the SQL Kafka Connector ...

Again, like the state machine example PR, I'm not sure we really should be moving the whole e2e test as is, but rather:

modify the original E2E in apache/flink to not use Kafka
Move over only the parts related to Kafka in apache/flink-connector-kafka

mas-chen:
Discussed offline to allow pinning versions with test scope

PR #28, opened 2023-05-08
[FLINK-32019][Connector/Kafka]  EARLIEST offset strategy for partitions discoveried later based on FLIP-288

loserwang1024:

As described in [[FLIP-288](https://cwiki.apache.org/confluence/display/FLINK/FLIP-288%3A+Enable+Dynamic+Partition+Discovery+by+Default+in+Kafka+Source)](https://cwiki.apache.org/confluence/display/FLINK/FLIP-288%3A+Enable+Dynamic+Partition+Discovery+by+Default+in+Kafka+Source), the strategy used for new partitions is the same as the initial offset strategy, which is not reasonable.

According to the semantics, if the startup strategy is latest, the consumed data should include all data from the moment of startup, which also includes all messages from new created partitions. However, the latest strategy currently maybe used for new partitions, leading to the loss of some data (thinking a new partition is created and might be discovered by Kafka source several minutes later, and the message produced into the partition within the gap might be dropped if we use for example "latest" as the initial offset strategy).if the data from all new partitions is not read, it does not meet the user's expectations.

Other ploblems see final Section: `User specifies OffsetsInitializer for new partition` .

Therefore,  it’s better to provide an **EARLIEST** strategy for later discovered partitions.

### Brief change log

1. Expand `KafkaSourceEnumState` with `TopicPartitionWithAssignStatus` to distinguish between initial partitions and newly discovered partitions. `TopicPartitionWithAssignStatus` is also better for future expansion, as new statuses can be added without changing the state results.
2. Add a `newDiscoveryOffsetsInitializer`(EARLIEST) to get offsets for newly discovered partitions. 
3. Modify `kafkaSourceEnumStateSerializer` to handle the expanded `KafkaSourceEnumState`.

### Verifying this change

1. Test the backward compatibility of state when deserializing in `KafkaSourceEnumStateSerializerTest`.
2. Expand `KafkaEnumeratorTest#testSnapshotState` method to test snapshot state in more scenarios:
    1. Before first discovery, so the state should be empty
    2. First partition discovery after start, but no assignments to readers
    3. Assign partials partitions to readers
    4. Assign all partitions to readers
3. Expand `KafkaEnumeratorTest#testDiscoverPartitionsPeriodically` method to test whether new partitions use EARLIEST offset while initial partitions use specified offset strategy.

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumState.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumState.java
@@ -22,18 +22,72 @@ import org.apache.flink.annotation.Internal;
 
 import org.apache.kafka.common.TopicPartition;
 
+import java.util.HashSet;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /** The state of Kafka source enumerator. */
 @Internal
 public class KafkaSourceEnumState {
-    private final Set<TopicPartition> assignedPartitions;
+    /** Partitions with status: ASSIGNED or UNASSIGNED_INITIAL. */
+    private final Set<TopicPartitionAndAssignmentStatus> partitions;
+    /**
+     * this flag will be marked as true if inital partitions are discovered after enumerator starts.
+     */
+    private final boolean initialDiscoveryFinished;
 
-    KafkaSourceEnumState(Set<TopicPartition> assignedPartitions) {
-        this.assignedPartitions = assignedPartitions;
+    public KafkaSourceEnumState(
+            Set<TopicPartitionAndAssignmentStatus> partitions, boolean initialDiscoveryFinished) {
+        this.partitions = partitions;
+        this.initialDiscoveryFinished = initialDiscoveryFinished;
+    }
+
+    KafkaSourceEnumState(
+            Set<TopicPartition> assignPartitions,
+            Set<TopicPartition> unassignedInitialPartitions,
+            boolean initialDiscoveryFinished) {
+        this.partitions = new HashSet<>();
+        partitions.addAll(
+                assignPartitions.stream()
+                        .map(
+                                topicPartition ->
+                                        new TopicPartitionAndAssignmentStatus(
+                                                topicPartition, AssignmentStatus.ASSIGNED))
+                        .collect(Collectors.toSet()));
+        partitions.addAll(
+                unassignedInitialPartitions.stream()
+                        .map(
+                                topicPartition ->
+                                        new TopicPartitionAndAssignmentStatus(
+                                                topicPartition,
+                                                AssignmentStatus.UNASSIGNED_INITIAL))
+                        .collect(Collectors.toSet()));
+        this.initialDiscoveryFinished = initialDiscoveryFinished;
+    }
+
+    public Set<TopicPartitionAndAssignmentStatus> partitions() {
+        return partitions;
     }
 
     public Set<TopicPartition> assignedPartitions() {
-        return assignedPartitions;
+        return filterPartitionsByAssignmentStatus(AssignmentStatus.ASSIGNED);
+    }
+
+    public Set<TopicPartition> unassignedInitialPartitions() {
+        return filterPartitionsByAssignmentStatus(AssignmentStatus.UNASSIGNED_INITIAL);
+    }
+
+    public boolean initialDiscoveryFinished() {
+        return initialDiscoveryFinished;
+    }
+
+    private Set<TopicPartition> filterPartitionsByAssignmentStatus(
+            AssignmentStatus assignmentStatus) {
+        return partitions.stream()
+                .filter(
+                        partitionWithStatus ->
+                                partitionWithStatus.assignmentStatus().equals(assignmentStatus))
+                .map(TopicPartitionAndAssignmentStatus::topicPartition)
+                .collect(Collectors.toSet());
     }
 }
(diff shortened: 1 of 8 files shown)


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumState.java
```diff
@@ -22,18 +22,69 @@
...
+    private final Set<TopicPartitionWithAssignStatus> partitions;
+    /**
+     * this flag will be marked as true if inital partitions are discovered after enumerator starts.
+     */
+    private final boolean initialDiscoveryFinished;
 
-    KafkaSourceEnumState(Set<TopicPartition> assignedPartitions) {
-        this.assignedPartitions = assignedPartitions;
+    KafkaSourceEnumState(
+            Set<TopicPartition> assignPartitions,
```
RamanVerma, line 40: please change the parameter names to `assignedPartitions` and `unassignedInitialPartitions`

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/TopicPartitionWithAssignStatus.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.apache.flink.annotation.Internal;
+
+import org.apache.kafka.common.TopicPartition;
+
+/** Kafka partition with assign status. */
+@Internal
+public class TopicPartitionWithAssignStatus {
+    private final TopicPartition topicPartition;
+    private final long assignStatus;
```
RamanVerma, line 29: `assignmentStatus` would convey the meaning better than `assignStatus`
Also, I would prefer `TopicPartitionAndAssignmentStatus` over `TopicPartitionWithAssignStatus`

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumState.java
```diff
@@ -22,18 +22,69 @@
...
+        this.initialDiscoveryFinished = initialDiscoveryFinished;
+    }
+
+    public Set<TopicPartitionWithAssignStatus> partitions() {
+        return partitions;
     }
 
     public Set<TopicPartition> assignedPartitions() {
-        return assignedPartitions;
+        return partitions.stream()
```
RamanVerma, line 68: Lines 68-74 and 78-84 duplicate the code a bit. 
Maybe you can define a private method to abstract the common code and call it from `assignedPartitions()` and `unassignedPartitions()`

So, something like this
```
private Set<TopicPartition> filterPartitions(long assignmentStatus);
```

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
```diff
@@ -113,10 +124,13 @@ public KafkaSourceEnumerator(
             Properties properties,
             SplitEnumeratorContext<KafkaPartitionSplit> context,
             Boundedness boundedness,
-            Set<TopicPartition> assignedPartitions) {
+            Set<TopicPartition> assignedPartitions,
```
RamanVerma, line 127: It will be better to pass the `KafkaSourceEnumState` object in this constructor to limit the number of arguments.

RamanVerma (commented):
Thanks for the changes @loserwang1024 
I have left some initial comments.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumState.java
loserwang1024: @RamanVerma Thanks for your advice. Would you like to code review again?

loserwang1024:
Now that @RamanVerma is busy, could anyone else help me? CC, @PatrickRen

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
```diff
@@ -266,10 +286,18 @@ private void checkPartitionChanges(Set<TopicPartition> fetchedPartitions, Throwa
     private PartitionSplitChange initializePartitionSplits(PartitionChange partitionChange) {
         Set<TopicPartition> newPartitions =
                 Collections.unmodifiableSet(partitionChange.getNewPartitions());
+
         OffsetsInitializer.PartitionOffsetsRetriever offsetsRetriever = getOffsetsRetriever();
+        // intial partitions use OffsetsInitializer specified by the user while new partitions use
```
PatrickRen, line 291: nit: `intial` -> `initial`

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/enumerator/KafkaEnumeratorTest.java
```diff
@@ -632,6 +702,31 @@ private void verifySplitAssignmentWithPartitions(
         assertThat(actualTopicPartitions).isEqualTo(allTopicPartitionsFromAssignment);
     }
 
+    /**
+     * get all assigned partition splits of topics.
+     *
+     * @param context
+     * @param topics
+     * @return
```
PatrickRen, line 710: Uncompleted JavaDocs

PatrickRen (changes requested):
@loserwang1024 Thanks for the PR! The overall logic looks good to me. I left some comments about naming and code style issues.

loserwang1024:
@PatrickRen CC

PatrickRen (approved):
@loserwang1024 Thanks for the update. LGTM.

Could you squash all commits into one and rebase the latest master? I triggered a CI just now and let's wait for the result.

PR #29, opened 2023-05-08
[FLINK-32021][Connectors/Kafka] Improvement the Javadoc for SpecifiedOffsetsInitializer and TimestampOffsetsInitializer

loserwang1024:

As described in [[FLIP-288](https://cwiki.apache.org/confluence/display/FLINK/FLIP-288%3A+Enable+Dynamic+Partition+Discovery+by+Default+in+Kafka+Source)](https://cwiki.apache.org/confluence/display/FLINK/FLIP-288%3A+Enable+Dynamic+Partition+Discovery+by+Default+in+Kafka+Source), Current JavaDoc does not fully explain the behavior of OffsetsInitializers. When the partition does not meet the condition, there will be a different offset strategy. 

This may lead to misunderstandings in the design and usage.

### Brief change log

Add to SpecifiedOffsetsInitializer: "Use Specified offset for specified partitions while use commit offset or Earliest for unspecified partitions. Specified partition offset should be less than the latest offset, otherwise it will start from the earliest."

Add to TimestampOffsetsInitializer:Initialize the offsets based on a timestamp. If the message meeting the requirement of the timestamp have not been produced to Kafka yet, just use the latest offset.

### Verifying this change

no code changes.

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/initializer/SpecifiedOffsetsInitializer.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/initializer/SpecifiedOffsetsInitializer.java
@@ -38,6 +38,11 @@ import static org.apache.flink.util.Preconditions.checkState;
  * An implementation of {@link OffsetsInitializer} which initializes the offsets of the partition
  * according to the user specified offsets.
  *
+ * <p>Use specified offsets for specified partitions while use commit offsets or offsetResetStrategy
+ * for unspecified partitions. Specified partition's offset should be less than its latest offset,
+ * otherwise it will start from the offsetResetStrategy. The default value of offsetResetStrategy is
+ * earliest.
+ *
  * <p>Package private and should be instantiated via {@link OffsetsInitializer}.
  */
 class SpecifiedOffsetsInitializer implements OffsetsInitializer, OffsetsInitializerValidator {
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/initializer/TimestampOffsetsInitializer.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/initializer/TimestampOffsetsInitializer.java
@@ -28,6 +28,8 @@ import java.util.Map;
 
 /**
  * An implementation of {@link OffsetsInitializer} to initialize the offsets based on a timestamp.
+ * If the message meeting the requirement of the timestamp have not been produced to Kafka yet, just
+ * use the latest offset.
  *
  * <p>Package private and should be instantiated via {@link OffsetsInitializer}.
  */


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/initializer/SpecifiedOffsetsInitializer.java
```diff
@@ -38,6 +38,10 @@
  * An implementation of {@link OffsetsInitializer} which initializes the offsets of the partition
  * according to the user specified offsets.
  *
+ * <p>Use specified offsets for specified partitions while use commit offsets or earliest for
```
RamanVerma, line 41: @loserwang1024 I have a question.
When you mention `specified offsets`, do you mean the ones provided in the map `initialOffsets`.
If that is true, these unspecified partitions are initialized to be consumed according to `offsetResetStrategy`, which can be `earliest` or `latest`. So, not necessarily `earliest`

Also, could you point me to the code (or code change) that explains this
> Specified partition's offset should be less than its latest offset,
  otherwise it will start from the earliest.

loserwang1024: > These unspecified partitions are initialized to be consumed according to offsetResetStrategy
> 

yes, howerver, OffsetResetStrategy is set to earliest in OffsetsInitializer#offsets and `KafkaDynamicSource` .

Maybe I shoud give more desciptions like “use offsetResetStrategy which is default earliest”.

Yes, however, the OffsetResetStrategy is set to earliest in both OffsetsInitializer#offsets and KafkaDynamicSource. **Maybe I should add more details such as "use the default earliest offsetResetStrategy"**  .

> Specified partition's offset should be less than its latest offset, otherwise it will start from the earliest(maybe offsetResetStrategy).
> 

this code is in org.apache.kafka.clients.consumer.internals.Fetcher#handleOffsetOutOfRange, which use offsetResetStrategy for offsetOutOfRange.

You can see the last section of FLIP-288, where I provide a detailed explanation.

loserwang1024: @RamanVerma 
> These unspecified partitions are initialized to be consumed according to offsetResetStrategy
> 

yes, howerver, OffsetResetStrategy is set to earliest in OffsetsInitializer#offsets and `KafkaDynamicSource` .

Maybe I shoud give more desciptions like “use offsetResetStrategy which is default earliest”.

Yes, however, the OffsetResetStrategy is set to earliest in both OffsetsInitializer#offsets and KafkaDynamicSource. **Maybe I should add more details such as "use the default earliest offsetResetStrategy"**  .

> Specified partition's offset should be less than its latest offset, otherwise it will start from the earliest(maybe offsetResetStrategy).
> 

this code is in org.apache.kafka.clients.consumer.internals.Fetcher#handleOffsetOutOfRange, which use offsetResetStrategy for offsetOutOfRange.

You can see the last section of FLIP-288, where I provide a detailed explanation.

RamanVerma: thanks for the explanation @loserwang1024 
As you said, you can change that comment to "use offsetResetStrategy which is default earliest”.

RamanVerma (approved):
Thanks for the PR, @loserwang1024 
Please improve the documentation as you have mentioned in your comment, and then merge.
LGTM

tzulitai:
Thanks for the PR @loserwang1024 and @RamanVerma for the reviews!

Merging this.

leonardBang (approved):
Thanks @loserwang1024 for the contribution, LGTM

PR #31, opened 2023-06-01
[FLINK-31747] Remove Confluent Avro formats from externalized Kafka connector repo

tzulitai:
This PR removes again the Confluent Avro related formats from the externalized Kafka connector repo.

This was a result of a discussion with @twalthr. The formats will remain in the core `apache/flink` repo, as they can be commonly used by various different connectors and not just the Kafka connector.


mas-chen (commented):
Can we also remove the flink-confluent-schema-registry-e2e-tests since they are more directly related to the format? I guess we will have to pin the test version of the Kafka connector in `apache/flink`

tzulitai:
Good point @mas-chen. Following along the same logic for how other tests in `apache/flink` that depend on the Kafka connector handle this (i.e. pinning to a specific Kafka connector version), I guess we would do the same here for [flink-confluent-schema-registry-e2e-tests](https://issues.apache.org/jira/browse/FLINK-confluent-schema-registry-e2e-tests).

mas-chen:
Looks good aside from the CI error

tzulitai:
Seems to be a flaky test. I'll re-run to be sure and open a ticket for that.

PR #37, opened 2023-06-30
[WIP][FLINK-32522] Add commons-collections in kafka-sql-connector shade jar

chucheng92:

Apache commons-collections 3.x is a Java 1.3 compatible version, and it does not use Java 5 generics. Apache commons-collections4 4.4 is an upgraded version of commons-collections and it built by Java 8. So we bump commons-collections. this pr is related with https://github.com/apache/flink/pull/21442

## Brief change log

update pom

## Verifying this change

This change is a trivial rework / code cleanup.  Verifying by current total cases.

- Dependencies (does it add or upgrade a dependency): yes
- The serializers: no

## Documentation

- Does this pull request introduce a new feature?  no

### DIFF
--- a/flink-connector-kafka/pom.xml
+++ b/flink-connector-kafka/pom.xml
@@ -72,6 +72,13 @@ under the License.
             <version>${kafka.version}</version>
         </dependency>
 
+        <!-- When flink-core bump to commons-collections4, we can remove it.-->
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-collections4</artifactId>
+            <version>4.4</version>
+        </dependency>
+
         <!-- Tests -->
 
         <dependency>
--- a/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaConsumerBase.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaConsumerBase.java
@@ -59,7 +59,7 @@ import org.apache.flink.streaming.runtime.operators.util.AssignerWithPunctuatedW
 import org.apache.flink.util.ExceptionUtils;
 import org.apache.flink.util.SerializedValue;
 
-import org.apache.commons.collections.map.LinkedMap;
+import org.apache.commons.collections4.map.LinkedMap;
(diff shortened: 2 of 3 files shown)


flink-connector-kafka/pom.xml
```diff
@@ -72,6 +72,13 @@ under the License.
             <version>${kafka.version}</version>
         </dependency>
 
+        <!-- When flink-core bump to commons-collections4, we can remove it.-->
```
MartijnVisser, line 75: With the externalization of connectors, we shouldn't have the connector depend on an artifact that may or may not be bundled with Flink.

MartijnVisser (changes requested):
@chucheng92 I'm not sure this is the best solution. Ideally, all connector doesn't need to rely on a dependency to may or may not be available in Flink. Especially with connectors that will be compatible with multiple versions of Flink.

flink-connector-kafka/pom.xml
chucheng92: @MartijnVisser it will be removed when flink update finished. Now it'a circular.

flink-python test case in flink reference flink-sql-connector-kafka（not uber jar, without commons-collections class）, so search it in flink and cause ClassNotFoundException. 

And this kafka connector also reference flink-core. 

Must break circular references from side. Is there a better way？

chucheng92: Another way is flink keep commons-collections 3.x remained and add commons-collections4.
then kafka connector in this repo update to commons-collections4, and flink remove commons-collections 3.x.  it's flink-side way. what do u think?

chucheng92: > Another way is flink keep commons-collections 3.x remained and add commons-collections4. then kafka connector in this repo update to commons-collections4 (just replace package name, no dependency import), and flink remove commons-collections 3.x. it's [flink-side](https://issues.apache.org/jira/browse/FLINK-side) solution. what do u think?

@MartijnVisser Whether we can keep these 2 version in flink temporarily? 


In fact, spark also encounters similar problems (but not reference loop), and currently two versions coexist.


https://github.com/apache/spark/pull/35257
https://github.com/apache/spark/pull/36009/files

chucheng92:
@MartijnVisser This kind of similar upgrade may appear in the future, let's see if there is a better solution together, looking forward to your opinion. 😄

PR #38, opened 2023-07-03
[FLINK-32522] Add commons-collections in kafka-sql-connector shade jar

chucheng92:

Currently, externalized sql connector rely on flink main repo. but flink main repo has many test cases(especially in flink python) reference flink sql kafka connector.

If we change the dependencies(e.g. commons-collections) in flink main repo, it will cause exception.

We should let externalized flink connectors depend on flink main, not the other way around. So we need to add this dependency in shade jar.

Otherwise, it will cause external components block the upgrade of flink main.

## Brief change log

update pom

## Verifying this change

This change is a trivial rework / code cleanup.  Verifying by current total cases.

- Dependencies (does it add or upgrade a dependency): yes
- The serializers: no

## Documentation

- Does this pull request introduce a new feature?  no


chucheng92:
@MartijnVisser Hi, Martijn. We need to let connectors not rely on dependencies that may or may not be available in Flink. PTAL. thanks

chucheng92:
@MartijnVisser hi, Martijn. Can you help to review it?

MartijnVisser (changes requested):
@chucheng92 Looking at the dependency tree, there's still commons-collections coming in via Flink itself. If we want to remove the dependency, we should also make sure that it's not included from there

chucheng92:
> @chucheng92 Looking at the dependency tree, there's still commons-collections coming in via Flink itself. If we want to remove the dependency, we should also make sure that it's not included from there

Yes. I have updated the PR. 

commons-collections in kafka connector is inherited from flink (so we will see it in kafka). This is a circular dependency and we should break it.

What I am thinking about now is to change the kafka connector to commons-collections4. When the kafka connector is released, upgrade the kafka-connector that Python depends on to this version on the flink side. In this way, it is decoupled, and the flink side can also be upgraded to commons-collections4. So the https://github.com/apache/flink/pull/21442 can be resolved. And it will remove all the old commons-collections from flink-side.

chucheng92:
> @chucheng92 Looking at the dependency tree, there's still commons-collections coming in via Flink itself. If we want to remove the dependency, we should also make sure that it's not included from there

Of course, the more important point is that pyflink should not rely on connectors, even for integration tests

PR #39, opened 2023-07-04
[FLINK-32453] Make main and v3.0 branch build against Flink 1.18-SNAPSHOT

tzulitai:
This PR contains two commits that allows the Kafka connector to be built against Flink `1.18-SNAPSHOT`.

- cc684bb19e36a5c41f0a8225c865a063397c3ddb: Copy over `TypeSerializerUpgradeTestBase` due to a breaking change introduced in [FLINK-27518](https://issues.apache.org/jira/browse/FLINK-27518) across Flink `1.17.x` and `1.18.x`.
- a0f588eeebce13e32e2232bfb9fd39f459b886da Misc hotfix for ambiguous method call due to multiple-matching overloading signatures

For the former, the long-term proper fix would be to properly introduce a public-facing replacement test utility for the {{TypeSerializerUpgradeTestBase}}, and have the Flink Kafka connector test code gradually move to that instead.


gaoyunhaii (approved):
LGTM

XComp (commented):
I'm wondering how we should handle that in the future. So far, only Kafka connector utilized this test base. But if we, for instance, externalize the formats as well, we would run into the same issue with Avro. And in general, it might be a framework for other connectors to use in the future.

Would it make sense to create a new testbase class for connectors in `flink-connector-common` that implements the test data generation and mark it as @PublicEvolving? To make it more obvious that it's used in the external repositories?

Additionally, where is the release documentation kept for the Kafka connector. Having this test in the external repo would mean that the release manager would have to generate the test data manually. But I couldn't find any docs on that one. :thinking:

tzulitai:
re: @XComp 

> Would it make sense to create a new testbase class for connectors in flink-connector-common that implements the test data generation and mark it as @PublicEvolving? To make it more obvious that it's used in the external repositories?

I was thinking along the same lines, although it should just be a public-facing test utility that Flink distributes. I can imagine any user implementing implementing their own `TypeSerializer` who might want to write migration tests for them using the base class.

> Additionally, where is the release documentation kept for the Kafka connector.

There's this: https://cwiki.apache.org/confluence/display/FLINK/Creating+a+flink-connector+release

tzulitai:
@XComp there's isn't a Kafka-connector specific release documentation, as far as I'm aware of. So far, the generic doc was good enough at least for the first `v3.0,0` release of the connector. Adding steps for generating the snapshot test files to the readme sounds good for now.

XComp:
I created FLINK-32582 as a follow-up ticket.

PR #40, opened 2023-07-07
[FLINK-32020] Enable Dynamic Partition Discovery by Default in Kafka Source based on FLIP-288

loserwang1024:
As described in [FLIP-288](https://cwiki.apache.org/confluence/display/FLINK/FLIP-288%3A+Enable+Dynamic+Partition+Discovery+by+Default+in+Kafka+Source),  dynamic partition discovery is disabled by default, and users have to specify the interval of discovery in order to turn it on. It will be better if enable partition discovery by default.

### Brief change log
- Kafka table source: "scan.topic-partition-discovery.interval" will be set to 5 minutes by default, aligned with the default value of ["metadata.max.age.ms" in Kafka consumer.](https://kafka.apache.org/documentation/#consumerconfigs_metadata.max.age.ms)
- Kafka stream source: "partition.discovery.interval.ms"  will be set to 5 minutes by default, with discovery interval set to 5 minutes.

### Verifying this change
- See test unit.
-  Start flink with kafka source connector by default, then add kafka partition after a period of time to see if the new partition can be read.

### DIFF
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/enumerator/KafkaEnumeratorTest.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/enumerator/KafkaEnumeratorTest.java
@@ -519,6 +519,26 @@ public class KafkaEnumeratorTest {
                 .isEqualTo(KafkaSourceOptions.PARTITION_DISCOVERY_INTERVAL_MS.defaultValue());
     }
 
+    @Test
+    public void testDisablePartitionDiscovery() {
+        Properties props = new Properties();
+        props.setProperty(
+                KafkaSourceOptions.PARTITION_DISCOVERY_INTERVAL_MS.key(), String.valueOf(0));
+        KafkaSourceEnumerator enumerator =
+                new KafkaSourceEnumerator(
+                        KafkaSubscriber.getTopicListSubscriber(Arrays.asList(TOPIC1, TOPIC2)),
+                        OffsetsInitializer.earliest(),
+                        new NoStoppingOffsetsInitializer(),
+                        props,
+                        new MockSplitEnumeratorContext<>(NUM_SUBTASKS),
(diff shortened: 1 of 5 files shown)


PatrickRen (commented):
@loserwang1024 Thanks for the PR! The logic looks good to me overall. Could you add another test case that validates disabling dynamic partition discovery?

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/enumerator/KafkaEnumeratorTest.java
```diff
@@ -501,6 +501,44 @@ public void testPartitionChangeChecking() throws Throwable {
...
+                        OffsetsInitializer.earliest(),
+                        new NoStoppingOffsetsInitializer(),
+                        props,
+                        new MockSplitEnumeratorContext<>(NUM_SUBTASKS),
+                        Boundedness.CONTINUOUS_UNBOUNDED,
+                        new KafkaSourceEnumState(
+                                Collections.emptySet(), Collections.emptySet(), false));
+        long partitionDiscoveryIntervalMs =
+                (long) Whitebox.getInternalState(enumerator, "partitionDiscoveryIntervalMs");
+        assertThat(partitionDiscoveryIntervalMs).isEqualTo(0);
```
PatrickRen, line 539: If we only check the value of `partitionDiscoveryIntervalMs` here, this test case only guards the logic of parsing properties and setting up the value of corresponding variable. I think we can dig deeper, to check if setting partition discovery interval to 0 does disable the partition discovery, by adding another assertion:
```
assertThat(context.periodicCallables()).isEmpty();
```
where `context` is the `MockSplitEnumeratorContext` that passed into the enumerator.

PatrickRen (commented):
@loserwang1024 Thanks for the update. I left a tiny comment about the new test case you added.

PR #45, opened 2023-08-23
[FLINK-32893][Connectors/Kafka] Allow the clientid to be fully configurable

bdbene:
[FLINK-32893](https://issues.apache.org/jira/browse/FLINK-32893)
There's some use cases to use the same clientID across all of the low level Kafka consumers to
1. Simplify how quotas are applied
2. Lower the cardinality of metrics reported by the Kafka brokers

## Brief Change Log
Support an optional configuration for the exact clientID to run Kafka consumers with

## Verifying this change
It's difficult to write unit tests for this which is probably why there's no existing unit tests for the client ID. However I was able to run this locally and could see the client ID was logged exactly as I specified it (no prefix or suffix).


MartijnVisser:
@bdbene Can you please rebase your PR?

bdbene:
Done :)

MartijnVisser:
@bdbene Please rebase instead of merging, since we can't merge in merge commits

bdbene:
Oops, fixed that

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSourceOptions.java
```diff
@@ -36,6 +36,14 @@ public class KafkaSourceOptions {
                     .noDefaultValue()
                     .withDescription("The prefix to use for the Kafka consumers.");
 
+    public static final ConfigOption<String> COMPLETE_CLIENT_ID =
+            ConfigOptions.key("client.id.complete")
```
jychen7, line 40:
```suggestion
            ConfigOptions.key("client.id")
```
shall we just use the same config as kafka clients?

I agree this would NOT be backwards compatible in Flink source, but more align with Kafka clients

PR #48, opened 2023-09-13
[FLINK-28758] Fix stop-with-savepoint for FlinkKafkaConsumer

pnowojski:
https://issues.apache.org/jira/browse/FLINK-28758?focusedCommentId=17700124&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17700124

PR #49, opened 2023-09-15
[FLINK-28758] Fix stop-with-savepoint for FlinkKafkaConsumer

pnowojski:
this is a backport of https://github.com/apache/flink-connector-kafka/pull/48


ngannt1710:
I use FlinkKafkaConsumer with two consumer(2 nodes) and two partitions. But the two consumers receive the same message regardless of which partitions it is pushed to.

PR #52, opened 2023-09-26
[FLINK-28303] Support LatestOffsetsInitializer to avoid latest-offset strategy lose data

Tan-JiaLiang:

Currently, enumerator would send the LATEST_OFFSET marker into reader, when no data is consumed the currentOffset will always equal to -1. This causes the saved currentOffset to be -1 at checkpoint/savepoint, meaning that data is lost between the time the task stops and the time it resumes.

For example:
Assume topic has 3 partitions, A/B/C
1. Start job with `latest-offsets` offsets initializer. A/B/C's `currentOffset` is -1, which means that it will call `seekToEnd` in `KafkaSourceReader`.
2. Write a record to partition A and B, C is empty. The `currentOffset` for three partitions is A=1, B=1, C=-1.
3. Stop the job with savepoint. It will save the `currentOffset` A=1, B=1, C=-1 into savepoint.
4. Write a record for each partition. All partitions get a new record. The `endOffset` for each partition is A=2, B=2, C=1 after written.
5. Restore the job from savepoint. After restore, partition A's and B's consume is ok, but because of the partition C's `currentOffset` is -1, it will call the `seekToEnd` API to restore consume, so data offset=1 lost.

## Brief change log

  - *Implement the LatestOffsetsInitializer for latest-offset strategy.*
  - *Remove KafkaPartitionSplit.LATEST_OFFSET and its dependency.*
  - *Modify the corresponding test case above.*

## Verifying this change

Please make sure both new and modified tests in this PR follows the conventions defined in our code quality guide: https://flink.apache.org/contributing/code-style-and-quality-common.html#testing

This change is already covered by existing tests, such as
  - *KafkaTableITCase#testLatestOffsetStrategyResume.*
  - *OffsetsInitializerTest#testLatestOffsetsInitializer.*
  - *KafkaDynamicTableFactoryTest#testBoundedLatestOffset.*
  - *UpsertKafkaDynamicTableFactoryTest#testBoundedLatestOffset.*

  - Dependencies (does it add or upgrade a dependency): No
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: No
  - The serializers: No
  - The runtime per-record code paths (performance sensitive): No
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Kubernetes/Yarn, ZooKeeper: No
  - The S3 file system connector: No

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
@@ -201,8 +201,7 @@ public class KafkaPartitionSplitReader
 
         // Seek on the newly assigned partitions to their stating offsets.
         seekToStartingOffsets(
-                partitionsStartingFromEarliest,
-                partitionsStartingFromSpecifiedOffsets);
+                partitionsStartingFromEarliest, partitionsStartingFromSpecifiedOffsets);
         // Setup the stopping offsets.
         acquireAndSetStoppingOffsets(partitionsStoppingAtCommitted);
 
@@ -268,8 +267,7 @@ public class KafkaPartitionSplitReader
     }
 
     private void parseStoppingOffsets(
-            KafkaPartitionSplit split,
-            Set<TopicPartition> partitionsStoppingAtCommitted) {
+            KafkaPartitionSplit split, Set<TopicPartition> partitionsStoppingAtCommitted) {
         TopicPartition tp = split.getTopicPartition();
         split.getStoppingOffset()
                 .ifPresent(
@@ -305,8 +303,7 @@ public class KafkaPartitionSplitReader
         }
     }
 
-    private void acquireAndSetStoppingOffsets(
-            Set<TopicPartition> partitionsStoppingAtCommitted) {
+    private void acquireAndSetStoppingOffsets(Set<TopicPartition> partitionsStoppingAtCommitted) {
         if (!partitionsStoppingAtCommitted.isEmpty()) {
             retryOnWakeup(
                             () -> consumer.committed(partitionsStoppingAtCommitted),
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
@@ -177,7 +177,9 @@ public class KafkaPartitionSplitReader
         // Starting offsets.
         Map<TopicPartition, Long> partitionsStartingFromSpecifiedOffsets = new HashMap<>();
         List<TopicPartition> partitionsStartingFromEarliest = new ArrayList<>();
+        List<TopicPartition> partitionsStartingFromLatest = new ArrayList<>();
         // Stopping offsets.
+        List<TopicPartition> partitionsStoppingAtLatest = new ArrayList<>();
         Set<TopicPartition> partitionsStoppingAtCommitted = new HashSet<>();
 
         // Parse the starting and stopping offsets.
@@ -189,8 +191,10 @@ public class KafkaPartitionSplitReader
                             parseStartingOffsets(
                                     s,
                                     partitionsStartingFromEarliest,
+                                    partitionsStartingFromLatest,
                                     partitionsStartingFromSpecifiedOffsets);
-                            parseStoppingOffsets(s, partitionsStoppingAtCommitted);
+                            parseStoppingOffsets(
+                                    s, partitionsStoppingAtLatest, partitionsStoppingAtCommitted);
                             // Track the new topic partition in metrics
                             kafkaSourceReaderMetrics.registerTopicPartition(s.getTopicPartition());
                         });
@@ -201,9 +205,11 @@ public class KafkaPartitionSplitReader
 
         // Seek on the newly assigned partitions to their stating offsets.
         seekToStartingOffsets(
-                partitionsStartingFromEarliest, partitionsStartingFromSpecifiedOffsets);
+                partitionsStartingFromEarliest,
+                partitionsStartingFromLatest,
+                partitionsStartingFromSpecifiedOffsets);
         // Setup the stopping offsets.
-        acquireAndSetStoppingOffsets(partitionsStoppingAtCommitted);
+        acquireAndSetStoppingOffsets(partitionsStoppingAtLatest, partitionsStoppingAtCommitted);
 
         // After acquiring the starting and stopping offsets, remove the empty splits if necessary.
         removeEmptySplits();
@@ -253,11 +259,14 @@ public class KafkaPartitionSplitReader
     private void parseStartingOffsets(
             KafkaPartitionSplit split,
             List<TopicPartition> partitionsStartingFromEarliest,
+            List<TopicPartition> partitionsStartingFromLatest,
             Map<TopicPartition, Long> partitionsStartingFromSpecifiedOffsets) {
         TopicPartition tp = split.getTopicPartition();
         // Parse starting offsets.
         if (split.getStartingOffset() == KafkaPartitionSplit.EARLIEST_OFFSET) {
             partitionsStartingFromEarliest.add(tp);
+        } else if (split.getStartingOffset() == KafkaPartitionSplit.LATEST_OFFSET) {
+            partitionsStartingFromLatest.add(tp);
         } else if (split.getStartingOffset() == KafkaPartitionSplit.COMMITTED_OFFSET) {
             // Do nothing here, the consumer will first try to get the committed offsets of
             // these partitions by default.
@@ -267,13 +276,17 @@ public class KafkaPartitionSplitReader
     }
 
     private void parseStoppingOffsets(
-            KafkaPartitionSplit split, Set<TopicPartition> partitionsStoppingAtCommitted) {
+            KafkaPartitionSplit split,
+            List<TopicPartition> partitionsStoppingAtLatest,
+            Set<TopicPartition> partitionsStoppingAtCommitted) {
         TopicPartition tp = split.getTopicPartition();
         split.getStoppingOffset()
                 .ifPresent(
                         stoppingOffset -> {
                             if (stoppingOffset >= 0) {
                                 stoppingOffsets.put(tp, stoppingOffset);
+                            } else if (stoppingOffset == KafkaPartitionSplit.LATEST_OFFSET) {
+                                partitionsStoppingAtLatest.add(tp);
                             } else if (stoppingOffset == KafkaPartitionSplit.COMMITTED_OFFSET) {
                                 partitionsStoppingAtCommitted.add(tp);
                             } else {
@@ -288,11 +301,17 @@ public class KafkaPartitionSplitReader
 
     private void seekToStartingOffsets(
             List<TopicPartition> partitionsStartingFromEarliest,
+            List<TopicPartition> partitionsStartingFromLatest,
             Map<TopicPartition, Long> partitionsStartingFromSpecifiedOffsets) {
 
         if (!partitionsStartingFromEarliest.isEmpty()) {
             LOG.trace("Seeking starting offsets to beginning: {}", partitionsStartingFromEarliest);
             consumer.seekToBeginning(partitionsStartingFromEarliest);
+        }
+
+        if (!partitionsStartingFromLatest.isEmpty()) {
+            LOG.trace("Seeking starting offsets to end: {}", partitionsStartingFromLatest);
+            consumer.seekToEnd(partitionsStartingFromLatest);
         }
 
         if (!partitionsStartingFromSpecifiedOffsets.isEmpty()) {
@@ -303,7 +322,11 @@ public class KafkaPartitionSplitReader
         }
     }
 
-    private void acquireAndSetStoppingOffsets(Set<TopicPartition> partitionsStoppingAtCommitted) {
+    private void acquireAndSetStoppingOffsets(
+            List<TopicPartition> partitionsStoppingAtLatest,
+            Set<TopicPartition> partitionsStoppingAtCommitted) {
+        Map<TopicPartition, Long> endOffset = consumer.endOffsets(partitionsStoppingAtLatest);
+        stoppingOffsets.putAll(endOffset);
         if (!partitionsStoppingAtCommitted.isEmpty()) {
             retryOnWakeup(
                             () -> consumer.committed(partitionsStoppingAtCommitted),
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
@@ -177,9 +177,7 @@ public class KafkaPartitionSplitReader
         // Starting offsets.
         Map<TopicPartition, Long> partitionsStartingFromSpecifiedOffsets = new HashMap<>();
         List<TopicPartition> partitionsStartingFromEarliest = new ArrayList<>();
-        List<TopicPartition> partitionsStartingFromLatest = new ArrayList<>();
         // Stopping offsets.
-        List<TopicPartition> partitionsStoppingAtLatest = new ArrayList<>();
         Set<TopicPartition> partitionsStoppingAtCommitted = new HashSet<>();
 
         // Parse the starting and stopping offsets.
@@ -191,10 +189,8 @@ public class KafkaPartitionSplitReader
                             parseStartingOffsets(
                                     s,
                                     partitionsStartingFromEarliest,
-                                    partitionsStartingFromLatest,
                                     partitionsStartingFromSpecifiedOffsets);
-                            parseStoppingOffsets(
-                                    s, partitionsStoppingAtLatest, partitionsStoppingAtCommitted);
+                            parseStoppingOffsets(s, partitionsStoppingAtCommitted);
                             // Track the new topic partition in metrics
                             kafkaSourceReaderMetrics.registerTopicPartition(s.getTopicPartition());
                         });
@@ -206,10 +202,9 @@ public class KafkaPartitionSplitReader
         // Seek on the newly assigned partitions to their stating offsets.
         seekToStartingOffsets(
                 partitionsStartingFromEarliest,
-                partitionsStartingFromLatest,
                 partitionsStartingFromSpecifiedOffsets);
         // Setup the stopping offsets.
-        acquireAndSetStoppingOffsets(partitionsStoppingAtLatest, partitionsStoppingAtCommitted);
+        acquireAndSetStoppingOffsets(partitionsStoppingAtCommitted);
 
         // After acquiring the starting and stopping offsets, remove the empty splits if necessary.
         removeEmptySplits();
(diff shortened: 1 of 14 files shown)


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -205,11 +201,9 @@ public void handleSplitsChanges(SplitsChange<KafkaPartitionSplit> splitsChange)
 
         // Seek on the newly assigned partitions to their stating offsets.
         seekToStartingOffsets(
-                partitionsStartingFromEarliest,
-                partitionsStartingFromLatest,
-                partitionsStartingFromSpecifiedOffsets);
+                partitionsStartingFromEarliest, partitionsStartingFromSpecifiedOffsets);
         // Setup the stopping offsets.
-        acquireAndSetStoppingOffsets(partitionsStoppingAtLatest, partitionsStoppingAtCommitted);
+        acquireAndSetStoppingOffsets(partitionsStoppingAtCommitted);
```
Gerrrr, line 206: nit: with this method only acquiring committed offsets, we should change its name.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/enumerator/KafkaEnumeratorTest.java
```diff
@@ -300,7 +300,7 @@ public void testDiscoverPartitionsPeriodically() throws Throwable {
                     getAllAssignSplits(context, PRE_EXISTING_TOPICS);
             assertThat(initialPartitionAssign)
                     .extracting(KafkaPartitionSplit::getStartingOffset)
-                    .containsOnly(KafkaPartitionSplit.LATEST_OFFSET);
+                    .containsOnly((long) KafkaSourceTestEnv.NUM_RECORDS_PER_PARTITION);
```
Gerrrr, line 303: nit: redundant cast

Gerrrr (commented):
Thanks for the patch! I do think that we should add an e2e test, say to `KafkaTableITCase` that follows the repro instructions in the issue:
```
1. Set up a Kafka topic with 10 partitions
2. produce records 0-9 into the topic
3. take a savepoint and stop the job
4. produce records 10-19 into the topic
5. restore the job from the savepoint.
```

I think that we should add the test because otherwise there is no automatic way to verify that we do not accidentally reintroduce that issue again some time in the future. It will also speed up the review process since it will be easy to verify that the problem is fixed. For example, the e2e test fails on `main` and passes on this branch.

Please let me know if you need a hand with developing that test.

Tan-JiaLiang:
> Thanks for the patch! I do think that we should add an e2e test, say to `KafkaTableITCase` that follows the repro instructions in the issue:
> 
> ```
> 1. Set up a Kafka topic with 10 partitions
> 2. produce records 0-9 into the topic
> 3. take a savepoint and stop the job
> 4. produce records 10-19 into the topic
> 5. restore the job from the savepoint.
> ```
> 
> I think that we should add the test because otherwise there is no automatic way to verify that we do not accidentally reintroduce that issue again some time in the future. It will also speed up the review process since it will be easy to verify that the problem is fixed. For example, the e2e test fails on `main` and passes on this branch.
> 
> Please let me know if you need a hand with developing that test.

@Gerrrr Thanks for the code review, very useful suggestions. 
I have add test case in KafkaTableITCase#testLatestOffsetStrategyResume, please help to check.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/enumerator/KafkaEnumeratorTest.java
Tan-JiaLiang: I had been double check, removing it will results in a syntax error.

Gerrrr (approved):
I checked that the test passes on this branch and fails on `main` with 

```
java.util.concurrent.TimeoutException: Can not get the expected result.

	at org.apache.flink.core.testutils.CommonTestUtils.waitUtil(CommonTestUtils.java:214)
	at org.apache.flink.core.testutils.CommonTestUtils.waitUtil(CommonTestUtils.java:230)
	at org.apache.flink.streaming.connectors.kafka.table.KafkaTableTestUtils.waitingExpectedResults(KafkaTableTestUtils.java:101)
	at org.apache.flink.streaming.connectors.kafka.table.KafkaTableITCase.testLatestOffsetStrategyResume(KafkaTableITCase.java:970)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.apache.flink.util.TestNameProvider$1.evaluate(TestNameProvider.java:45)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.testcontainers.containers.FailureDetectingExternalResource$1.evaluate(FailureDetectingExternalResource.java:30)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:54)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:54)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:69)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater$1.execute(IdeaTestRunner.java:38)
	at com.intellij.rt.execution.junit.TestsRepeater.repeat(TestsRepeater.java:11)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:35)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:232)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:55)
```

I also double-checked that on `main` this test fails exactly in the way described in the issue; there will be one or more missing rows among produced before resuming the job.

@tzulitai it looks like this PR needs a manual intervention to run CI. Can you please kick it off?

tzulitai:
Thanks for the fix and review, @Tan-JiaLiang @Gerrrr.
I'll do one last pass and then hopefully merge this ASAP.

tzulitai:
@Tan-JiaLiang there's something that doesn't quite add up here that I'm trying to understand.

The scenario that is being fixed is the following, if I'm understanding correctly (assume topic has 3 partitions, A / B / C):

t=0:  Partitions A, B are written some data. C is empty.

t=1: Start job with `latest-offsets` offsets initializer. `currentOffsets` for partitions A and B will be some meaningful value, while for partition C we have `-1`.

t=2: Take savepoint.

t=3: Add more records to topic. All of A / B / C get new records.

t=4: Restore from savepoint. KafkaSource only reads the new records from partitions A + B, while for C we don't see any records?

If that's the case, shouldn't partition C be started from earliest offset at t=4 instead of from latest offset, which seems to be what this PR change does?

Tan-JiaLiang:
@tzulitai Sorry for late. 
In your example, all of partitions's `currentOffset` is -1 at tp=1 before this PR changes. After this PR changes, all the partitions's `currentOffset` will be assign a meaningful value at tp=1.

Let's take the test case `KafkaTableITCase#testLatestOffsetStrategyResume` as an example.
Assume topic has 3 partitions, A/B/C
tp=0: Start job with `latest-offsets` offsets initializer.
tp=1: Write a record to partition A and B, C is empty.
tp=2: Stop the job with savepoint.
tp=3: Write a record to each partitions. All of partitions get a new record.
tp=4: Restore the job from savepoint.

Before this PR changes:
tp=0: A/B/C `currentOffset` is -1. The `currentOffset` is -1 means that it will call `seekToEnd` in `KafkaSourceReader`.
tp=1: The `currentOffset` for three partitions is A=1, B=1, C=-1.
tp=2: We save the `currentOffset` A=1, B=1, C=-1 into savepoint.
tp=3: Partitions's `endOffset` is A=2, B=2, C=1 after written.
tp=4: After restore, partition A and B's consume is ok, but because of `currentOffset` C=-1, it will consume with `seekToEnd`, so data offset=1 lost.

After this PR changes:
tp=0: A/B/C `currentOffset` is 0. I query the end offset in `KafkaSourceEnumerator` instead of call `seekToEnd` in `KafkaSourceReader`.
tp=1: The `currentOffset` for three partitions is A=1, B=1, C=0.
tp=2: We save the `currentOffset` A=1, B=1, C=0 into savepoint.
tp=3: Partitions's `endOffset` is A=2, B=2, C=1 after written.
tp=4: After restore, partition A/B/C's consume is ok because of `currentOffset` C=0, so no data lost.

tzulitai:
Thanks for the explanation @Tan-JiaLiang. This is aligned with @Gerrrr and my understanding as well after we discussed this offline.

I'm going to do one more pass now and try to merge this ASAP. We should get out a new release for the connector soon once this is merged.

tzulitai:
One more comment:
It seems like the old `FlinkKafkaConsumer` (on the deprecated `SourceFunction` interface) was handling this case correctly.
Apparently, this corner case was overlooked when implement the newer `KafkaSource` on the new `Source` interface :-(

tzulitai:
I thought about this more closely, and while the current fix will work, I'm not so sure it's actually how we want to do things. My main issue with this PR is the following behavioral change: with this PR, we now eagerly query Kafka for the latest offsets on the JM, instead of only lazily realizing the actual offsets on the TM (which was the whole point for the `ReaderHandledOffsetsInitializer` class).

So - I'm wondering if the correct fix is to actually, at time of `snapshotState` on the TMs, IFF the `currentOffset` of a snapshotted `KafkaPartitionSplitState` is still the `LATEST_OFFSET` marker (i.e. `-1`), we should replace it with `EARLIEST_OFFSET` marker (i.e. `-2`) on the spot and store that into the snapshots.

Semantically we get the same result - at restore time, partitions that were previously empty will be read from the beginning, instead of seeked to end at restore time and potentially skipping records. But we still also retain the nice property of lazily realizing offsets only on TMs.

tzulitai:
@Tan-JiaLiang @Gerrrr I'm going to open a new alternative PR tonight that addresses the problem based on my latest proposal, and will cherry pick over all new tests added already here. Will ping once ready.

Tan-JiaLiang:
@tzulitai Maybe this will lead to a new bug.
Assume topic has 3 partitions, A/B/C
tp=0: Write a record to each partitions. All of partitions get a new record. The offset of A=1,B=1,C=1.
tp=1: Start job with latest-offsets offsets initializer.
tp=2: Write a record to partition A and B, C is empty.
tp=3: Stop the job with savepoint. We save the `currentOffset` A=2,B=2,C=-1 into savepoint.
tp=4: Restore the job from savepoint. After restore, partition A and B's consume is ok, but because of currentOffset C=-1, it will consume with `seekToBeginning`, so data offset=0 will be consumed, which should not to be.

But after your proposal, I think i'm not considering the compatibility of the state, `currentOffset` will error in restore when it equals to -1, which will cause confused for users after they upgrade the connector.
I will refine the above test case and state compatibility problem if need after we discuss a result.

tzulitai:
@Tan-JiaLiang you are right, that's why it's taking me a while to open the new PR. The problem with my approach is that it can potentially read more records when restoring.

Essentially what we need to do, is that at _some_ point before the first snapshot is taken, we need to actually look up the offsets for empty partitions and replace any kind of marker. Your PR does this very early on the JMs, and I'm trying to figure out if there's any way to postpone this as much as possible to the TMs.

tzulitai:
@Tan-JiaLiang ok - lets try to reach consensus here.

It does seem like that my approach can lead to other problems, e.g. bombarding Kafka brokers from multiple end-offset lookups across multiple TMs. This might turn out worse than doing just one more extra query on the JM.

And since JMs right already are assumed to have access to Kafka brokers already anyways, we should be breaking anything for users.

So overall - lets proceed with this PR's approach. Sorry for the back and forth @Tan-JiaLiang, just wanted to make sure that we're fixing this properly.

Regarding the restore issue you mentioned: yes, please address that, and then I think we can merge this.

Tan-JiaLiang:
@tzulitai I found it difficult to get the end-offset and set into the state when TMs restore. Because of this problem doesn't come up very often and once the partition fetch records, that will not a problem. So i would like to still call the `seekToEnd` API when currentOffset is -1 to compatibility with the history state.

I rollback the `KafkaPartitionSplitReader` changes and refine the test case, can you please have a look? Let's finish it in this week.

Tan-JiaLiang:
@mas-chen Hi Mason, thanks for your feedback.

I think empty splits may not be the point in this problem. The point is we should not store the marker (i.e. -1) into state.

Currently we handle the empty splits on TMs. If we handle on the enumerator:
1. We need to store the UNASIGNED partition's offset into state in JM when checkpoint/savepoint.
2. We need to consider compatibility with the history state.
3. We need to reconsider the bounded mode and unbounded mode cases.

I don't want to make too many changes because current connector seems still healthy, although it has some problem. I'd like to fix it instead of changing it.

tzulitai:
@Tan-JiaLiang @mas-chen 

I do agree that changing the split assignment logic in the enumerator right now is too big of a change - although I do see the merit of doing that.

> this can happen with any offset initializer strategy (Flink provided ones, custom implementations). 

I disagree here.

The `LATEST_OFFSET` initializer strategy is particularly interesting specifically because `LATEST` means different things depending on when the offset is physically resolved, whereas for the end user, when using `LATEST` their intention is for this to loosely correlate to "start progress from the time of job submission". Therefore, if we don't eagerly resolve `LATEST` markers prior to first checkpoint occurring, from the users' perspective we can be skipping records if.

Other strategies don't have this issue:
1. For `EARLIEST`, if there were no records initially by first snapshot, on restore we still attempt to read from the earliest position.
2. For `TIMESTAMP`, if there were no records initially by first snapshot, on restore we still attempt to read from said timestamp.
3. For `GROUP_OFFSETS`, if there were no records initially by first snapshot, on restore we still attempt to read from whatever group offsets were indicated by Kafka.
4. For `SPECIFIC_OFFSETS`, if there were no records initially by first snapshot, on restore we still attempt to read from the specified offsets.

tzulitai:
I'm taking another pass on the PR now @Tan-JiaLiang.

mas-chen:
Thanks for explaining, the PR description was confusing and I thought the issue was that the Kafka API was returning -1 offset for empty partition. 

So basically, with this change, we should be storing the initial offset of 0 in these cases (would be good to add unit test with the new offset initializer that verifies this).

I thought of an alternative solution on the reader side to set offsets to `consumer.position(tp) - 1` (I also realized this is how the old implementation handles it). But I guess you could have an edge case, in which your checkpoint starts before offsets are resolved by the reader...

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/split/KafkaPartitionSplit.java
```diff
@@ -35,17 +35,18 @@
 public class KafkaPartitionSplit implements SourceSplit {
     public static final long NO_STOPPING_OFFSET = Long.MIN_VALUE;
     // Indicating the split should consume from the latest.
-    public static final long LATEST_OFFSET = -1;
+    // @deprecated Only be used for compatibility with the history state, see FLINK-28303
+    @Deprecated public static final long LATEST_OFFSET = -1;
```
mas-chen, line 39: Simiarly, we could remove it from Javadocs in `ReaderHandledOffsetsInitializer` (the javadocs describes how it supports latest which we want to avoid using).

Tan-JiaLiang:
> the PR description was confusing and I thought the issue was that the Kafka API was returning -1 offset for empty partition.

Sorry for that, I rewrite already.


> So basically, with this change, we should be storing the initial offset of 0 in these cases (would be good to add unit test with the new offset initializer that verifies this).

Agree, and the test case was already added. 
See `org.apache.flink.streaming.connectors.kafka.table.KafkaTableITCase#testLatestOffsetStrategyResume`


> I thought of an alternative solution on the reader side to set offsets to consumer.position(tp) - 1 (I also realized this is how the old implementation handles it). But I guess you could have an edge case, in which your checkpoint starts before offsets are resolved by the reader...

I took a look at the old implementation, maybe different from the current.

The old implementation can position the offset, replace the startup marker(e.g LATEST_OFFSET/EARLIEST_OFFSET) and set it into the `KafkaTopicPartitionState`, It can do it all in `KafkaConsumerThread`.

The new implementation can position the offset in `KafkaPartitionSplitReader`, but there is no way to set it into the `KafkaPartitionSplitState`. The `KafkaPartitionSplitState#currentOffset` can only call in `KafkaRecordEmitter` in current design.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/split/KafkaPartitionSplit.java
Tan-JiaLiang: I was negligent. Fixed.

tzulitai (approved):
Thanks @Tan-JiaLiang, latest iteration LGTM.
Merging this now.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/split/KafkaPartitionSplit.java
```diff
@@ -35,17 +35,18 @@
...
     // Indicating the split should consume from the last committed offset.
     public static final long COMMITTED_OFFSET = -3;
 
     // Valid special starting offsets
     public static final Set<Long> VALID_STARTING_OFFSET_MARKERS =
-            new HashSet<>(Arrays.asList(EARLIEST_OFFSET, LATEST_OFFSET, COMMITTED_OFFSET));
+            new HashSet<>(Arrays.asList(EARLIEST_OFFSET, COMMITTED_OFFSET));
     public static final Set<Long> VALID_STOPPING_OFFSET_MARKERS =
-            new HashSet<>(Arrays.asList(LATEST_OFFSET, COMMITTED_OFFSET, NO_STOPPING_OFFSET));
+            new HashSet<>(Arrays.asList(COMMITTED_OFFSET, NO_STOPPING_OFFSET));
```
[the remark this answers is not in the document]
tzulitai: I'll apply a hotfix.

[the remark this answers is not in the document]
Tan-JiaLiang: I notice it's not merged yet, so I fixed it.

PR #56, opened 2023-10-09
[FLINK-33219][connector/kafka] Add new archunit violation messages introduced by FLINK-31804

PatrickRen:
The reason we add new violation messages instead of update existing ones is that the patch of FLINK-31804 is only applied after Flink 1.18. We need to make sure the CI could run successfully for Flink versions before and after that.

If Kafka connector decides to drop support for versions before 1.18 in the future, please re-freeze the violations then.

### DIFF
--- a/.github/workflows/push_pr.yml
+++ b/.github/workflows/push_pr.yml
@@ -23,6 +23,10 @@ concurrency:
   cancel-in-progress: true
 jobs:
   compile_and_test:
+    strategy:
+      matrix:
+        flink: [ 1.17-SNAPSHOT, 1.18-SNAPSHOT ]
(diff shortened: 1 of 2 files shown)


.github/workflows/push_pr.yml
```diff
@@ -23,6 +23,10 @@ concurrency:
   cancel-in-progress: true
 jobs:
   compile_and_test:
+    strategy:
+      matrix:
+        flink: [ 1.17-SNAPSHOT, 1.18-SNAPSHOT ]
```
MartijnVisser, line 28:
```suggestion
        flink: [ 1.17.1, 1.18-SNAPSHOT ]
```

MartijnVisser (approved):

PR #57, opened 2023-10-09
[FLINK-33191][Connector/Kafka] Remove dependency on Flink Shaded

MartijnVisser:


flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/deserializer/KafkaRecordDeserializationSchemaTest.java
```diff
@@ -82,16 +81,24 @@ public void testKafkaDeserializationSchemaWrapper() throws Exception {
     @Test
     public void testKafkaValueDeserializationSchemaWrapper() throws Exception {
         final ConsumerRecord<byte[], byte[]> consumerRecord = getConsumerRecord();
-        KafkaRecordDeserializationSchema<ObjectNode> schema =
-                KafkaRecordDeserializationSchema.valueOnly(
-                        new JsonDeserializationSchema<>(ObjectNode.class));
+        KafkaRecordDeserializationSchema<
+                        org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.node
```
tzulitai, line 85: I'm a bit lost here. Why do these test code still need to use the shaded version?

MartijnVisser: > Why do these test code still need to use the shaded version?

It tests `flink-shaded-jackson` ObjectNodes that are coming from Flink's JSON format implementation. I talked about it with @zentol and we think it's a reasonable state to be in (until connectors live in their own plugin folder/classpath). If a user relies on `flink-json` they are locking themselves in to that specific Flink version, and therefore `flink-shaded-jackson`. If they use the JSON support from the connector itself, then they get vanilla jackson and are independent of Flink.

tzulitai (approved):

tzulitai:
I'm merging this now to `main` and `v3.0`

PR #58, opened 2023-10-10
[FLINK-33231] [source] Properly evict offsetsToCommit cache on checkpoint complete if no offsets exist

tzulitai:
Prior to this fix, if the offsets to commit for a given checkpoint is empty, which can be the case if no starting offsets were retrieved from Kafka yet, then on checkpoint completion the cache is not properly evicted up to the given checkpoint.

This change fixes this such that in notifyOnCheckpointComplete, we shortcut the method execution to not need to try to commit the offsets since its empty anyways, and always remember to evict the cache up to the completed checkpoint.

## Testing

I've modified the existing `KafkaSourceReaderTest#testCommitEmptyOffsets()` test to fail if the cache eviction fix was not applied. With this PR, that test now passes.


MartijnVisser (approved):
LGTM

lauri-suurvali-cko:
LGTM!

PR #60, opened 2023-10-13
[hotfix][docs] Fix scan.bounded.mode missing forward options

Tan-JiaLiang:
1. add `scan.bounded.mode`, `scan.startup.specific-offsets`, `scan.bounded.specific-offsets` into forward options.
2. fix `scan.bounded.mode` docs.


Tan-JiaLiang:
@MartijnVisser  Sorry late.

> Do we have tests for them?

To be honest, I checked all connectors(e.g. jdbc/hbase/filesystem...) but there is no test case for forward options.

Forward options is a set of ConfigOption that are directly forwarded to the runtime implementation but don't affect the final execution topology. And according to the design of FLINK-24456, `scan.bounded.mode`, `scan.bounded.specific-offsets`, `scan.bounded.timestamp-millis` should be forward options.

> And can you also check the Chinese docs?

Yes, I filled in the gaps in the Chinese documentation and translated the content.

Finally, the `sink.parallelism` option should not be a forward option, as it can lead to inconsistent parallelism between upstream and downstream operators and change the execution topology.

PR #64, opened 2023-10-31
[FLINK-33401][connectors/kafka] Fix the wrong download link and version in the connector doc

TanYuxin-tyx:
Currently, Kafka Connector official documentation has a bug in the download link and the version number.
This is the fix for https://issues.apache.org/jira/browse/FLINK-33401. 

The doc link with the bug is 
https://nightlies.apache.org/flink/flink-docs-release-1.18/docs/connectors/datastream/kafka/

### DIFF
--- a/docs/content.zh/docs/connectors/datastream/kafka.md
+++ b/docs/content.zh/docs/connectors/datastream/kafka.md
@@ -36,16 +36,12 @@ Apache Flink 集成了通用的 Kafka 连接器，它会尽力与 Kafka client �
 当前 Kafka client 向后兼容 0.10.0 或更高版本的 Kafka broker。
 有关 Kafka 兼容性的更多细节，请参考 [Kafka 官方文档](https://kafka.apache.org/protocol.html#protocol_compatibility)。
 
-{{< connector_artifact flink-connector-kafka 3.0.0 >}}
-
-如果使用 Kafka source，```flink-connector-base``` 也需要包含在依赖中：
-
-{{< artifact flink-connector-base >}}
+{{< connector_artifact flink-connector-kafka kafka >}}
 
 Flink 目前的流连接器还不是二进制发行版的一部分。
 [在此处]({{< ref "docs/dev/configuration/overview" >}})可以了解到如何链接它们，从而在集群中运行。
 
-{{< py_download_link "kafka" >}}
+{{< py_connector_download_link "kafka" >}}
 
 ## Kafka Source
 {{< hint info >}}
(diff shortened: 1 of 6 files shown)


TanYuxin-tyx:
@MartijnVisser Hi, Martijn, could you please help review this?

TanYuxin-tyx:
@MartijnVisser, Thanks for helping review.

This fix aims to resolve the issue of displaying the incorrect version number (-1.18) for the Kafka connector.

Currently, the download link and version are not accessible until the new connector is released. This is expected, as it is similar to other connectors such as the Pulsar connector (https://nightlies.apache.org/flink/flink-docs-release-1.18/docs/connectors/datastream/pulsar/), Kinesis connector (https://nightlies.apache.org/flink/flink-docs-release-1.18/docs/connectors/datastream/kinesis/), Elasticsearch connector (https://nightlies.apache.org/flink/flink-docs-release-1.18/docs/connectors/datastream/elasticsearch/), and Google PubSub dependency (https://nightlies.apache.org/flink/flink-docs-release-1.18/docs/connectors/datastream/pubsub/), and so on.

Upon visiting these links, you will notice that the download links and dependencies are not available, resulting in a 404 error or the inability to find the dependency. However, their displayed versions are accurate. 

Therefore, this fix aims to ensure that the Kafka connector also displays the correct version and link. Once the new connector is released, these links will naturally become accessible.

MartijnVisser:
> This fix aims to resolve the issue of displaying the incorrect version number (-1.18) for the Kafka connector.

How it should work, is that in the right branch for the Flink docs (see https://github.com/apache/flink/blob/master/docs/setup_docs.sh#L46-L56 for the setup in Flink's `master` branch) the correct branch/tag from the externalized connector should be included. If there's no connector available for that version (which is the case for everything that's displayed on `master`), we should not display a download link at all.

When a connector version becomes available, the release branch for Flink should be updated to point to the correct branch/tag from the externalized connector

TanYuxin-tyx:
@MartijnVisser Thanks. It's my fault to use the wrong target branch. I have updated the target branch to v3.0. 

The reason why it can fix the incorrect -1.18 display is that the version number is derived from the following URL: https://github.com/apache/flink/blob/dc86179e4626ff9efacc637b85a487dd16869e05/docs/layouts/shortcodes/connector_artifact.html#L26.

The original code is `connector_artifact flink-connector-kafka 3.0.0`. However, this code fails to assist Flink shortcode (flink/docs/layouts/shortcodes/connector_artifact.html) in constructing the correct version number due to the following two lines:
```
{{ $connector := index .Site.Data $name }}
{{ $version := $connector.version  }}
```
(Here $name is from the second field of the args, so it is 3.0.0, and .Site.Data cannot be found.)

By modifying it to `connector_artifact flink-connector-kafka kafka`, it becomes capable of generating the correct version based on `$connector.version` from the .Site.Data of `themes/connectors/data/kafka.yml`.

MartijnVisser (approved):
LGTM

PR #67, opened 2023-11-10
[hotfix] fix_kafka_sasl_ssl_example in the docs 

531651225:
- Removing unnecessary transition symbols'/'in  the document 

`  'properties.sasl.jaas.config' = 'org.apache.kafka.common.security.scram.ScramLoginModule required username=\"username\" password=\"password\";'`

图片

- using transition characters can cause "Value not specified for key 'username' in JAAS config" error,The examples in the document can mislead users
图片


MartijnVisser (changes requested):
@531651225 Can you please make sure to update it for both English and Chinese docs?

PR #70, opened 2023-11-30
[FLINK-33545][Connectors/Kafka] KafkaSink implementation can cause dataloss during broker issue when not using EXACTLY_ONCE if there's any batching

aeolus811tw:

To address the current flow of KafkaProducer having exact same behavior for `DeliveryGuarantee.NONE` and `DeliveryGuarantee.AT_LEAST_ONCE`

It is based on the understanding that the existing flush performed on producer via `prepareSnapshotPreBarrier` and the actual checkpoint completion on commit has a very rare race condition where there could be data being invoked via processElement after the PreBarrier flush, and if KafkaProducer is having retry on a batched data that has yet thrown any error, upon job failure (caused by broker) will cause the batched data to never be committed, and since checkpoint was successful, these data will be lost.

This PR address the issue by enabling AT_LEAST_ONCE to have an opportunity to flush again when commit is happening when needed to avoid this issue. This is to ensure at the end of the checkpoint cycle, producer will definitely have no data left in its buffer.

Please comment or verify on the above understanding.

## Brief change log

 - *add variable `hasRecordsInBuffer` to `FlinkKafkaInternalProducer` and will be set when send/flush/close are called
 - *add variable `transactional` to `KafkaCommittable` to track whether a committable is transactional or not
 - *add new constructor to `KafkaCommittable` for Unit Test backward compatibility
 - *have `prepareCommit` also return list of commitable for DeliveryGuarantee.AT_LEAST_ONCE in `prepareCommit()`
 - *have `KafkaCommitter` check the new transactional value on `KafkaCommittable` to perform `commitTransaction()` to preserve original EXACTLY_ONCE pathway

## Verifying this change

 - *Added second flush test to test this particular special case (where producer.send was invoked after a flush)*
 - *Manually verified the job runs correctly with this change in existing cluster*

  - The serializers: no

## Documentation


MartijnVisser:
@hhktseng Can you rebase your PR?

hhktseng:
> @hhktseng Can you rebase your PR?

@MartijnVisser can you point me to which commit to rebase onto?

thanks

MartijnVisser:
> @MartijnVisser can you point me to which commit to rebase onto?

@hhktseng On the latest changes from `main` please

hhktseng:
created patch and applied after syncing to latest commit, then replaced forked branch with latest sync + patch

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducer.java
```diff
@@ -74,6 +76,10 @@ private static Properties withTransactionalId(
 
     @Override
     public Future<RecordMetadata> send(ProducerRecord<K, V> record, Callback callback) {
+        if (flushGated) {
```
mas-chen, line 79: I don't think this helps because during a SinkWriter#flush(), a SinkWriter#write() cannot be invoked. So this boolean condition never occurs. For your issue, we are looking to replicate this behavior https://github.com/apache/flink-connector-kafka/blob/15f2662eccf461d9d539ed87a78c9851cd17fa43/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer.java#L1107 to ensure that all records are flushed before completing the checkpoint, even in the case of Kafka client timeouts/retries.

hhktseng: reworked to follow [flink-connector-kafka/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer.java](https://github.com/apache/flink-connector-kafka/blob/15f2662eccf461d9d539ed87a78c9851cd17fa43/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer.java#L1107)

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducer.java
```diff
@@ -396,8 +408,27 @@ public String toString() {
...
                 + "', inTransaction="
                 + inTransaction
+                + ", pendingRecords="
+                + pendingRecords.get()
                 + ", closed="
                 + closed
                 + '}';
     }
+
+    public class TrackingCallback implements Callback {
```
mas-chen, line 418: nit: private

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducer.java
```diff
@@ -396,8 +408,27 @@ public String toString() {
...
+
+        private final Callback actualCallback;
+
+        public TrackingCallback(final Callback actualCallback) {
+            this.actualCallback = actualCallback;
+        }
+
+        @Override
+        public void onCompletion(final RecordMetadata recordMetadata, final Exception e) {
+            pendingRecords.decrementAndGet();
```
mas-chen, line 428: Do we want to decrement after the callback is completed? What's the best approach semantically?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducer.java
```diff
@@ -86,6 +93,11 @@ public void flush() {
         if (inTransaction) {
             flushNewPartitions();
         }
+        final long pendingRecordsCount = pendingRecords.get();
+        if (pendingRecordsCount != 0) {
+            throw new IllegalStateException(
+                    "Pending record count must be zero at this point: " + pendingRecordsCount);
```
mas-chen, line 99: nit: how about this message like this:

`n pending records after flush. There must be no pending records left.`

mas-chen:
@hhktseng were you able to test that this change mitigates your original issue? Is there a way to repro in the tests?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducer.java
tweise: Since it already happened, we should probably decrement as soon as possible?

tweise:
@hhktseng thanks for working on this. Can you please address the review comments?

AHeise:
Please check my comment here. https://issues.apache.org/jira/browse/FLINK-33545?focusedCommentId=17863737&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17863737 
If everyone is sure that the current fix is addressing the actual issue, please go ahead.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducer.java
```diff
@@ -72,12 +74,17 @@ private static Properties withTransactionalId(
...
+    }
+
     @Override
     public Future<RecordMetadata> send(ProducerRecord<K, V> record, Callback callback) {
         if (inTransaction) {
             hasRecordsInTransaction = true;
         }
-        return super.send(record, callback);
+        pendingRecords.incrementAndGet();
+        return super.send(record, new TrackingCallback(callback));
```
AHeise, line 87: This change creates a new callback with every `send`. Since the callback being passed in our codebase is mostly constant, we should add a simple cache like `new LRUMap(3);`. The number is kind of arbitrary and 1 should work already. The most important part is that it shouldn't grow boundless or we get the next memory leak if I overlooked a dynamic usage ;).

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducer.java
AHeise: I'd improve the error message as follows:

`Some records have not been fully persisted in Kafka. As a precaution, Flink will restart to resume from previous checkpoint. Please report this issue with logs on https://issues.apache.org/jira/browse/FLINK-33545`.

AHeise (commented):
Thank you very much for your contribution. Changes look mostly good to me. I have added two small suggestions. 

I'm also wondering if we should add a property around this new behavior, so folks can turn off the check if it has unintended side-effects (performance degration or it's interfering with DeliveryGuarantee.NONE)

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducer.java
hhktseng: just want to check, are we proposing putting instance of TrackingCallback into the rotating cache? wouldn't that caused the previous callback that might not have been invoked to be ignored?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducer.java
hhktseng: i think having reference to the issue reported might allow us better tracking the potential problem

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducer.java
AHeise: I can't quite follow. I was proposing to use

`return super.send(record, callbackCache.computeIfAbsent(callback, TrackingCallback::new));`

So we have 3 cases:
- New callback, wrap in `TackingCallback` and cache.
- Existing callback (common case), retrieve existing callback and use it.
- Remove existing `TackingCallback` from cache if full.

In all cases, both the TackingCallback and the original callback will be invoked. The only difference to the code without cache is that we avoiding creating extra TrackingCallback instances around the same original callback.

PR #74, opened 2023-12-13
make sure that super.close() must be executed!

ustctozm:
make sure that super.close() must be executed!  #33692


MartijnVisser:
@ustctozm Please rebase your PR

MartijnVisser:
@mas-chen Any thoughts on this? Is this indeed a bug?

mas-chen (commented):
@ustctozm @MartijnVisser I think this is fine because the finally block addresses the exceptional condition--if the close fails due to exception in the abort, it would try closing the current producer. If closing the producer causes exception, we want to rethrow it. Given this, I think the current code is fine. Unless you have more logs of the error you are encountering?

PR #77, opened 2024-01-15
[FLINK-34089] Verify that the subscribed topics match the assigned topics

Tan-JiaLiang:

When we unsubscribe a topic and still restore from the old state, the job will still read data from the unsubscribed topic. I think we should check if that the subscribed topic partitions match the assigned partitions, and throw an error so the user can handle it.

## Brief change log

  - *Verify that the subscribed topics match the assigned topics.*

## Verifying this change

This change is already covered by existing tests, such as
  - *org.apache.flink.streaming.connectors.kafka.table.KafkaTableITCase#testChangingSubscribedTopicsBeforeResuming.*

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
@@ -162,6 +162,8 @@ public class KafkaSourceEnumerator
     @Override
     public void start() {
         adminClient = getKafkaAdminClient();
+        validateTopicPartitions(
+                subscriber.getSubscribedTopicPartitions(adminClient), assignedPartitions);
         if (partitionDiscoveryIntervalMs > 0) {
             LOG.info(
                     "Starting the KafkaSourceEnumerator for consumer group {} "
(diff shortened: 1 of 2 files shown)


Tan-JiaLiang:
@tzulitai Hi gordon, can you take a look when you are free?

MartijnVisser:
> When we unsubscribe a topic and still restore from the old state, the job will still read data from the unsubscribed topic.

Isn't this by design? If we would change this, we would rely on Kafka for failure recovery, which is something that we don't want to do.

Tan-JiaLiang:
@MartijnVisser 
Suppose there is a flink job that subscribes two topics in a source at the same time, TopicA and TopicB.

* Write some data to both TopicA and TopicB.
* Stop the job with savepoint, the offset of TopicA and TopicB will be saved in savepoint.
* Now I change the subscribe topic list, subscribe only the TopicB, and restore from the last savepoint.
* The Flink job will still consume the TopicA record, which will make user feel confused.

So I think it is better to clearly tell the user that we cannot restore from the savepoint because the subscribe list has been changed.

Tan-JiaLiang:
But now I am thinking more carefully, maybe just removing the topics from state that are not in the current subscription list is a better solution.

@MartijnVisser @tzulitai WDYT?

MartijnVisser:
> But now I am thinking more carefully, maybe just removing the topics from state that are not in the current subscription list is a better solution.

That's what a user indeed should do at this point. Just removing an input topic has implications on your pipeline, and users shouldn't just do that. If users have a requirement to (dynamically) add or remove Kafka topics, they should use the solution that has been brought forward with https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=217389320

Tan-JiaLiang:
But users need to know that they should not use checkpoint & savepoint to restore when they change the subscription topic list. Should we add a validation to prevent users from making mistakes?

MartijnVisser:
> But users need to know that they should not use checkpoint & savepoint to restore when they change the subscription topic list.

Users still can, but they have to take into consideration the same thing as if you would delete an operator from their job. That's already documented at https://nightlies.apache.org/flink/flink-docs-master/docs/ops/state/savepoints/#what-happens-if-i-delete-an-operator-that-has-state-from-my-job

PR #78, opened 2024-01-16
[hotfix] Synchronize CI pipeline setup

MartijnVisser:
The Flink Kafka connector now tests:
- All PRs against `main` are tested against Flink `1.17.2` and `1.18.1`, instead of `1.17.1` and `1.18.0`
- The weekly run tests -- `main` against `1.17-SNAPSHOT, `1.18-SNAPSHOT` and `1.19-SNAPSHOT` (1.19 is newly added) -- `v3.0` against `1.17-SNAPSHOT` and `1.18-SNAPSHOT` (as was before)

### DIFF
--- a/.github/workflows/push_pr.yml
+++ b/.github/workflows/push_pr.yml
@@ -25,19 +25,12 @@ jobs:
   compile_and_test:
     strategy:
       matrix:
-        flink: [ 1.17.1, 1.18.0 ]
-        jdk: [ 8, 11, 17 ]
-        exclude:
-          - jdk: 17
-            flink: 1.17.1
+        flink: [ 1.17.2 ]
+        jdk: [ '8, 11' ]
+        include:
+          - flink: 1.18.1
+            jdk: '8, 11, 17'
     uses: apache/flink-connector-shared-utils/.github/workflows/ci.yml@ci_utils
     with:
       flink_version: ${{ matrix.flink }}
       jdk_version: ${{ matrix.jdk }}
-  python_test:
-    strategy:
-      matrix:
-        flink: [ 1.17.1, 1.18.0 ]
-    uses: apache/flink-connector-shared-utils/.github/workflows/python_ci.yml@ci_utils
-    with:
-      flink_version: ${{ matrix.flink }}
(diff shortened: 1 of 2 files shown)


.github/workflows/push_pr.yml
```diff
@@ -25,19 +25,12 @@ jobs:
...
     with:
       flink_version: ${{ matrix.flink }}
-      jdk_version: ${{ matrix.jdk }}
-  python_test:
-    strategy:
-      matrix:
-        flink: [ 1.17.1, 1.18.0 ]
-    uses: apache/flink-connector-shared-utils/.github/workflows/python_ci.yml@ci_utils
-    with:
-      flink_version: ${{ matrix.flink }}
```
snuyanzin, line 43: should we remove ci for python part?

IIRC currently it is done separately

MartijnVisser: Oh wow I completely missed the Python part. Let me check

MartijnVisser: I think we should keep them in for PRs, and actually we should also have them for weekly runs.

snuyanzin:
since it is touched a couple of more ideas
1. by default python tests are running with jdk8, after https://github.com/apache/flink-connector-shared-utils/pull/30
it could be also customized for puthon tests, so may be would make sense to run python ci with different jdk  as well
2. it seems python ci is not included in weekly, I wonder whether we should add it there

.github/workflows/push_pr.yml
snuyanzin: +1

MartijnVisser:
> 1. it could be also customized for python tests

I think we should mirror what we do with Flink, so I think we should indeed also run them for JDK11 and JDK17? We could argue that JDK11/JDK17 would only be run during a weekly, to avoid overly hitting the servers. I believe we also run those tests during the nightly builds for Flink

> 2. it seems python ci is not included in weekly, I wonder whether we should add it there

We should indeed

snuyanzin:
yeah... initially there were already 2 jdks (8, 11) in ci for PRs and that's why I've just added other jdks there...
Now amount of ci jobs per is growing together with number of jdks + possibly twice because of python...

it would make sense to follow something similar we have in Flink main repo where everything is built with jdk8 for PRs and all other jdks are participating only during nightlies

snuyanzin (approved):

PR #79, opened 2024-01-17
[FLINK-32416] Fix flaky tests by ensuring test utilities produce records w…

mas-chen:
…ith consistency and cleanup notify no more splits to ensure it is sent

I have spent the last few days troubleshooting this and found that most issues are related to the bounded mode tests. These problems don't surface locally and I suspect that they are more likely to occur with smaller/busy CI machines. Issues:
1. Test records were not being produced to Kafka. I added the Kafka settings and refactored to the API to ensure all records are sent to all brokers.
2. When I fixed this, I ran into some tests that would exceed the 50 minute CI. This is because the the no more splits signal was not sent properly. I refactored the code a bit to make it clearer.

At this commit, the tests are quite stable. 6/7 CI runs pass.
Screenshot 2024-01-16 at 10 13 43 PM

In one CI run that failed, I did see that there was a Flink error, indicating a lost source event. In separate commits, there were other instances that exceeded 50 minutes but this can be solved separately, see https://issues.apache.org/jira/browse/FLINK-34127.

Original issue that Max found after #44 was merged: https://github.com/apache/flink-connector-kafka/actions/runs/7504290046/job/20433204351

cc: @mxm @tzulitai @MartijnVisser

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
@@ -96,6 +96,7 @@ public class DynamicKafkaSourceEnumerator
     private int kafkaMetadataServiceDiscoveryFailureCount;
     private Map<String, Set<String>> latestClusterTopicsMap;
     private Set<KafkaStream> latestKafkaStreams;
+    private boolean firstDiscoveryComplete;
 
     public DynamicKafkaSourceEnumerator(
             KafkaStreamSubscriber kafkaStreamSubscriber,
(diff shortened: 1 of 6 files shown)


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -212,32 +214,27 @@ public void start() {
 
     private void handleNoMoreSplits() {
         if (Boundedness.BOUNDED.equals(boundedness)) {
-            enumContext.runInCoordinatorThread(
```
mas-chen, line 215: No need to use this API. The code already runs in coordinator thread

MartijnVisser:
The only thing I'm wondering about is that we actually should file a Jira for this; otherwise LGTM

MartijnVisser:
> Can we put this under the original JIRA issue? I think that should be ok since we haven't release in the meantime.

Yeah that makes sense for me too, as a follow-up to actually complete the Jira

mxm (approved):

MartijnVisser (approved):
LGTM: I'll fix the commit message when merging and add it to the original Jira
Edit: I see that @mas-chen still wants to fix one thing, so I'll postpone merging this until he's done that himself :)

mas-chen:
@MartijnVisser which thing to fix are you referring to? I think this PR is ready as-is if it blocking your other work. The ticket I'll need a few days to look into, probably get to it at the end of the week.

MartijnVisser:
> which thing to fix are you referring to?

I misread https://github.com/apache/flink-connector-kafka/pull/79#discussion_r1454686838 as that you still wanted to fix something 🙈 

I'll merge it after this CI run, thanks :)

PR #88, opened 2024-03-27
FLINK-31361 - Added the right shaded dependency for SQL client connector for kafka - Updated kafka.md

diptimanr:
Modified the documentation on using the right dependency for 'properties.sasl.jaas.config'. When using 'flink-sql-connector-kafka.jar', existing document doesn't use the shaded dependency.
Also added the name of the jar file (flink-sql-connector-kafka-x.xx.x.jar).

### DIFF
--- a/docs/content/docs/connectors/table/kafka.md
+++ b/docs/content/docs/connectors/table/kafka.md
@@ -629,7 +629,7 @@ for more details.
 ### Security
 In order to enable security configurations including encryption and authentication, you just need to setup security
 configurations with "properties." prefix in table options. The code snippet below shows configuring Kafka table to
-use PLAIN as SASL mechanism and provide JAAS configuration when using SQL client JAR :
+use PLAIN as SASL mechanism and provide JAAS configuration when using SQL client JAR (flink-sql-connector-kafka-x.xx.x.jar) :
 ```sql
 CREATE TABLE KafkaTable (
   `user_id` BIGINT,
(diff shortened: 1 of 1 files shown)


docs/content/docs/connectors/table/kafka.md
```diff
@@ -674,7 +674,7 @@ Please note that the class path of the login module in `sasl.jaas.config` might
 client dependencies, so you may need to rewrite it with the actual class path of the module in the JAR.
 SQL client JAR has relocated Kafka client dependencies to `org.apache.flink.kafka.shaded.org.apache.kafka`,
 then the path of plain login module in code snippets above need to be
-`org.apache.flink.kafka.shaded.org.apache.kafka.common.security.plain.PlainLoginModule` when using SQL client JAR.
+`org.apache.flink.kafka.shaded.org.apache.kafka.common.security.plain.PlainLoginModule` when using SQL client JAR(flink-sql-connector-kafka-x.xx.x.jar).
```
alpinegizmo, line 677:
```suggestion
`org.apache.flink.kafka.shaded.org.apache.kafka.common.security.plain.PlainLoginModule` when using SQL client JAR (flink-sql-connector-kafka-x.xx.x.jar).
```

alpinegizmo (changes requested):
@diptimanr Thank you for your contribution. 

Aside from one small nit that I've suggested you fix, the basic content of the PR looks good. But would you please clean up the PR? In this case there should be just one commit, and more importantly, the commit message should be like this, with both the JIRA ticket and component clearly identified:

[FLINK-31361][docs] Added the right shaded dependency for SQL client connector for kafka

Also, please make the same change in the chinese language docs.

PR #90, opened 2024-04-04
[FLINK-35007] Add support for Flink 1.19

MartijnVisser:
This also includes dropping the weekly tests for the `v3.0` branch, since `v3.1` has been released and that's our main version going forward.

### DIFF
--- a/.github/workflows/push_pr.yml
+++ b/.github/workflows/push_pr.yml
@@ -30,7 +30,7 @@ jobs:
         include:
           - flink: 1.18.1
             jdk: '8, 11, 17'
-          - flink: 1.19-SNAPSHOT
+          - flink: 1.19.0
             jdk: '8, 11, 17, 21'
     uses: apache/flink-connector-shared-utils/.github/workflows/ci.yml@ci_utils
     with:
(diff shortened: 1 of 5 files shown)


flink-python/pyflink/pyflink_gateway_server.py
```diff
@@ -71,7 +71,7 @@ def read_from_config(key, default_value, flink_conf_file):
 def find_java_executable():
     java_executable = "java.exe" if on_windows() else "java"
     flink_home = _find_flink_home()
-    flink_conf_file = os.path.join(flink_home, "conf", "flink-conf.yaml")
+    flink_conf_file = os.path.join(flink_home, "conf", "config.yaml")
```
uce, line 74: We need a fallback to `flink-conf.yaml` for the older versions. Otherwise, this LGTM.

uce (changes requested):
LGTM modulo the fallback for `flink-conf.yaml`.

flink-python/pyflink/pyflink_gateway_server.py
MartijnVisser: Yes, I've decided to not do this change, but use an old `flink-conf.yaml` file and copy that to all locations where a `config.yaml` exists. Since that's the backwards guarantee for all Flink 1.* releases, I think this is a safe way for us to run all the Python tests

uce (approved):
LGTM. Fantastic!

PR #92, opened 2024-04-08
[FLINK-34961] Use dedicated CI name for Kafka connector to differentiate it in infra-reports

snuyanzin:
The PR will allow to differentiate between Kafka connector statistics and others with name ci


MartijnVisser (approved):
LGTM - Thanks for this.

PR #98, opened 2024-04-26
[FLINK-34127] Upgrade to flink-connector-parent 1.1.0 to exclude runn…

mas-chen:
…ing nested unit tests

The root cause of the duplicate nested test execution is noted in this github issue: https://github.com/junit-team/junit5/issues/3782. TLDR, the outer test runs once (running all its nested tests), then the nested tests run again due to the include test regex.

Surefire also excludes nested tests by default. https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html

I have tested this approach in https://github.com/mas-chen/flink-connector-kafka/pull/8 and realized that excluding tests was only available from 1.1.0 of the parent pom.


mas-chen:
This shaves 1-2 minutes from CI

PR #99, opened 2024-04-26
FLINK-35311 Kafka part required for Flink Apicurio Avro support. Prototype for review 

davidradl:


flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/internals/KafkaDeserializationSchemaWrapper.java
```diff
@@ -36,10 +42,36 @@ public class KafkaDeserializationSchemaWrapper<T> implements KafkaDeserializatio
 
     private static final long serialVersionUID = 2651665280744549932L;
 
+    private static Method deserializeWithAdditionalPropertiesMethod = null;
+
+    protected static final String IS_KEY = "IS_KEY";
+
+    protected static final String HEADERS = "HEADERS";
+
+    static {
```
nictownsend, line 51: Why is https://github.com/apache/flink-connector-kafka/pull/99/files?file-filters%5B%5D=.java&show-viewed-files=true#diff-0a1bb10356e97bd2de4e3e8a257cf8c94290dd0aadb133d9288b6d1bc60b5994R73 different in terms of the `static` block?

You use a static block here, but a static method in DynamicKafkaDeserializationSchema.

However, I like the idea of an interface method `initializeMethod` that returns `Method<DeserializationSchema>` rather than `void` - the typing gives safety that `intializeMethod` returns the Method from a class structure that has Deserialization as the parent.

I do appreciate that the typing doesn't guarantee the method returned is one with the correct parameters - but I think catching the runtime exception is as good as trying to validate the method before using it https://www.tutorialspoint.com/javareflect/javareflect_method_getparametertypes.htm

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchema.java
```diff
@@ -74,25 +113,63 @@ class DynamicKafkaRecordSerializationSchema implements KafkaRecordSerializationS
...
-                    topic,
-                    extractPartition(
-                            consumedRow,
-                            null,
-                            valueSerialized,
-                            context.getPartitionsForTopic(topic)),
-                    null,
-                    valueSerialized);
+            byte[] valueSerialized = getSerialized(consumedRow, inputMap, outputValueMap, false);
+            if (serializeWithAdditionalPropertiesMethod == null) {
```
nictownsend, line 132: This logic seems odd - you check for `serializeWithAdditionalPropertiesMethod` but then you do not use it in the `else` block.

It could be re-written to be cleaner regardless - the logic is only about adding headers, so you can put the header generation in an if  block and then create ProducerRecord later in both cases.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchema.java
```diff
@@ -74,25 +113,63 @@ class DynamicKafkaRecordSerializationSchema implements KafkaRecordSerializationS
...
+                        extractPartition(
+                                consumedRow,
+                                null,
+                                valueSerialized,
+                                context.getPartitionsForTopic(topic)),
+                        null, // timestamp will be current time
+                        null,
+                        valueSerialized,
+                        headers);
+            }
```
nictownsend, line 165:
```suggestion

                List<Header> headers = new ArrayList<>();
            if (serializeWithAdditionalPropertiesMethod != null) {
               Map<String, Object> headersToAddForValue =
                        (Map<String, Object>) outputValueMap.get(HEADERS);
                if (!MapUtils.isEmpty(headersToAddForValue)) {
                    for (String headerKey : headersToAddForValue.keySet()) {
                        KafkaDynamicSink.KafkaHeader kafkaHeader =
                                new KafkaDynamicSink.KafkaHeader(
                                        headerKey, (byte[]) headersToAddForValue.get(headerKey));
                        headers.add(kafkaHeader);
                    }
                }
               
                return new ProducerRecord<>(
                        topic,
                        extractPartition(
                                consumedRow,
                                null,
                                valueSerialized,
                                context.getPartitionsForTopic(topic)),
                        null, // timestamp will be current time
                        null,
                        valueSerialized,
                        headers);
            }
```

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaDeserializationSchemaTest.java
```diff
@@ -0,0 +1,199 @@
...
+                        testSpec.valueOnly ? 0 : 1,
+                        1,
+                        key,
+                        value,
+                        new RecordHeaders(),
+                        Optional.empty());
+        try {
+            dynamicKafkaRecordDeserializationSchema.deserialize(message, null);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
nictownsend, line 92: You can use https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#fail-java.lang.Throwable- `fail(cause)` to fail the junit test - rather than wrapping and rethrowing

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaDeserializationSchemaTest.java
```diff
@@ -0,0 +1,199 @@
...
+                            value,
+                            new RecordHeaders(),
+                            Optional.empty());
+            try {
+                dynamicKafkaRecordDeserializationSchema.deserialize(message, null);
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+        } finally {
+            DynamicKafkaRecordSerializationSchema.initializeMethod();
```
nictownsend, line 149: Why are you using `finally` here?

This looks like a side effect on a static object for the test - would it not be better under a `beforeEach()`?

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchemaTest.java
```diff
@@ -0,0 +1,230 @@
...
+
+            Long timestamp = null;
+            final ProducerRecord<byte[], byte[]> producerRecord =
+                    dynamicKafkaRecordSerializationSchema.serialize(
+                            consumedRow, context, timestamp);
+            Headers headers = producerRecord.headers();
+            assertThat(headers).isEmpty();
+        } finally {
+            // ensure the method is present after this test.
+            DynamicKafkaRecordSerializationSchema.initializeMethod();
```
nictownsend, line 105: Similar, I don't think it's good practice to use `finally` as a `beforeEach/afterEach`

nictownsend (changes requested):

PR #100, opened 2024-04-29
[FLINK-34470][Connectors/Kafka] Fix indefinite blocking by adjusting stopping condition in split reader

dongwoo6kim:
## Problem 
When using the flink kafka connector in batch scenarios, consuming transactional messages can cause indefinite hanging.
This issue can be easily reproduced with following steps. 
1. Produce transactional messages and commit them.
2. Configure `scan.bounded.mode` to `latest-offset` and run consumer using flink kafka connector

## Cause
The previous stopping condition in the `KafkaPartitionSplitReader` compared the offset of the last record with the `stoppingOffset`. This approach works for streaming use cases and batch processing of non-transactional messages. However, in scenarios involving transactional messages, this is insufficient.      
[Control messages](https://kafka.apache.org/documentation/#controlbatch), which are not visible to clients, can occupy the entire range between the last record's offset and the stoppingOffset which leads to indefinite blocking.

## Workaround
I've modified the stopping condition to use `consumer.position(tp)`, which effectively skips any control messages present in the current poll, pointing directly to the next record's offset. 
To handle edge cases, particularly when `properties.max.poll.records` is set to `1`, I've adjusted the fetch method to always check all assigned partitions, even if no records are returned in a poll.

#### Edge case example 
Consider partition `0`, where offsets `13` and `14` are valid records and `15` is a control record. If `stoppingOffset` is set to 15 for partition `0`and `properties.max.poll.records` is configured to `1`, checking only partitions that return records would miss offset 15. By consistently reviewing all assigned partitions, the consumer’s position jumps control record in the subsequent poll, allowing the system to escape.

## Discussion
To address the metric issue in [FLINK-33484](https://issues.apache.org/jira/browse/FLINK-33484), I think we need to make wrapper class of  `ConsumerRecord` for example `ConsumerRecordWithOffsetJump`.
```java
public ConsumerRecordWithOffsetJump(ConsumerRecord<K, V> record, long offsetJump) {
        this.record = record;
        this.offsetJump = offsetJump;
    }
```
And we may need new `KafkaPartitionSplitReader` that implements     
`SplitReader<ConsumerRecordWithOffsetJump<byte[], byte[]>, KafkaPartitionSplit>`.     
So when record is emitted it should set current offset not just `record.offset()+1` but     
`record.offset() + record.jumpValue` in [here](https://github.com/apache/flink-connector-kafka/blob/369e7be46a70fd50d68746498aed82105741e7d6/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaRecordEmitter.java#L54).  
  `jumpValue` is typically 1, except for the last record of each poll where it's calculated as     
`consumer.position() - lastRecord.offset()`.       
If this sounds good to everyone, I'm happy to work on this.

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
@@ -122,31 +122,31 @@ public class KafkaPartitionSplitReader
         KafkaPartitionSplitRecords recordsBySplits =
                 new KafkaPartitionSplitRecords(consumerRecords, kafkaSourceReaderMetrics);
         List<TopicPartition> finishedPartitions = new ArrayList<>();
-        for (TopicPartition tp : consumerRecords.partitions()) {
+        for (TopicPartition tp : consumer.assignment()) {
             long stoppingOffset = getStoppingOffset(tp);
-            final List<ConsumerRecord<byte[], byte[]>> recordsFromPartition =
-                    consumerRecords.records(tp);
-
-            if (recordsFromPartition.size() > 0) {
-                final ConsumerRecord<byte[], byte[]> lastRecord =
-                        recordsFromPartition.get(recordsFromPartition.size() - 1);
-
-                // After processing a record with offset of "stoppingOffset - 1", the split reader
-                // should not continue fetching because the record with stoppingOffset may not
-                // exist. Keep polling will just block forever.
-                if (lastRecord.offset() >= stoppingOffset - 1) {
-                    recordsBySplits.setPartitionStoppingOffset(tp, stoppingOffset);
-                    finishSplitAtRecord(
-                            tp,
-                            stoppingOffset,
-                            lastRecord.offset(),
-                            finishedPartitions,
-                            recordsBySplits);
-                }
+            long consumerPosition = consumer.position(tp);
+            // Stop fetching when the consumer's position reaches the stoppingOffset.
+            // Control messages may follow the last record; therefore, using the last record's
+            // offset as a stopping condition could result in indefinite blocking.
+            if (consumerPosition >= stoppingOffset) {
+                LOG.debug(
+                        "Position of {}: {}, has reached stopping offset: {}",
+                        tp,
+                        consumerPosition,
+                        stoppingOffset);
+                recordsBySplits.setPartitionStoppingOffset(tp, stoppingOffset);
+                finishSplitAtRecord(
+                        tp, stoppingOffset, consumerPosition, finishedPartitions, recordsBySplits);
             }
-            // Track this partition's record lag if it never appears before
-            kafkaSourceReaderMetrics.maybeAddRecordsLagMetric(consumer, tp);
         }
+
+        // Only track non-empty partition's record lag if it never appears before
+        consumerRecords
+                .partitions()
+                .forEach(
+                        trackTp -> {
+                            kafkaSourceReaderMetrics.maybeAddRecordsLagMetric(consumer, trackTp);
+                        });
 
         markEmptySplitsAsFinished(recordsBySplits);
 
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java
@@ -80,6 +80,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.apache.flink.connector.kafka.testutils.KafkaSourceExternalContext.SplitMappingMode.PARTITION;
 import static org.apache.flink.connector.kafka.testutils.KafkaSourceExternalContext.SplitMappingMode.TOPIC;
+import static org.apache.flink.streaming.connectors.kafka.KafkaTestBase.kafkaServer;
 import static org.assertj.core.api.Assertions.assertThat;
 
 /** Unite test class for {@link KafkaSource}. */
@@ -368,6 +369,40 @@ public class KafkaSourceITCase {
                             source,
                             WatermarkStrategy.noWatermarks(),
                             "testConsumingTopicWithEmptyPartitions"));
+        }
+
+        @Test
+        public void testConsumingTransactionalMessage() throws Throwable {
+            String transactionalTopic = "transactionalTopic-" + UUID.randomUUID();
+            KafkaSourceTestEnv.createTestTopic(
+                    transactionalTopic, KafkaSourceTestEnv.NUM_PARTITIONS, 1);
+            List<ProducerRecord<String, Integer>> records =
+                    KafkaSourceTestEnv.getRecordsForTopic(transactionalTopic);
+            KafkaSourceTestEnv.produceToKafka(
+                    records, kafkaServer.getTransactionalProducerConfig());
+            // After running KafkaSourceTestEnv.setupEarliestOffsets(transactionalTopic):
+            // - For each partition, records with offsets before partition number P are deleted.
+            //   - Partition 0: offset 0 is earliest
+            //   - Partition 5: offset 5 is earliest, 0-4 are deleted.
+            //   - Partition 9: offset 9 is earliest, 0-8 are deleted.
+            KafkaSourceTestEnv.setupEarliestOffsets(transactionalTopic);
+            KafkaSource<PartitionAndValue> source =
+                    KafkaSource.<PartitionAndValue>builder()
+                            .setBootstrapServers(KafkaSourceTestEnv.brokerConnectionStrings)
+                            .setTopics(transactionalTopic)
+                            .setGroupId("topic-with-transactional-message-test")
+                            .setDeserializer(new TestingKafkaRecordDeserializationSchema(false))
+                            .setStartingOffsets(OffsetsInitializer.earliest())
+                            .setBounded(OffsetsInitializer.latest())
+                            .build();
+            StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
+            env.setParallelism(1);
+            executeAndVerify(
+                    env,
+                    env.fromSource(
+                            source,
+                            WatermarkStrategy.noWatermarks(),
+                            "testConsumingTransactionalMessage"));
         }
     }
 
--- a/flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaTestBase.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaTestBase.java
@@ -46,6 +46,8 @@ import org.junit.rules.TemporaryFolder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nullable;
+
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -242,11 +244,15 @@ public abstract class KafkaTestBase extends TestLogger {
             Collection<ProducerRecord<K, V>> records,
             Class<? extends org.apache.kafka.common.serialization.Serializer<K>> keySerializerClass,
             Class<? extends org.apache.kafka.common.serialization.Serializer<V>>
-                    valueSerializerClass)
+                    valueSerializerClass,
+            @Nullable Properties extraProps)
             throws Throwable {
         Properties props = new Properties();
         props.putAll(standardProps);
         props.putAll(kafkaServer.getIdempotentProducerConfig());
+        if (extraProps != null) {
+            props.putAll(extraProps);
+        }
         props.setProperty(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, keySerializerClass.getName());
         props.setProperty(
                 ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, valueSerializerClass.getName());
@@ -261,8 +267,15 @@ public abstract class KafkaTestBase extends TestLogger {
                     }
                 };
         try (KafkaProducer<K, V> producer = new KafkaProducer<>(props)) {
+            if (props.containsKey(ProducerConfig.TRANSACTIONAL_ID_CONFIG)) {
+                producer.initTransactions();
+                producer.beginTransaction();
+            }
             for (ProducerRecord<K, V> record : records) {
                 producer.send(record, callback);
+            }
+            if (props.containsKey(ProducerConfig.TRANSACTIONAL_ID_CONFIG)) {
+                producer.commitTransaction();
             }
         }
         if (sendingError.get() != null) {
--- a/flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableITCase.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableITCase.java
@@ -54,6 +54,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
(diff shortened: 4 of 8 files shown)


LinMingQiang:
its work , i had try.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -122,27 +122,21 @@ public RecordsWithSplitIds<ConsumerRecord<byte[], byte[]>> fetch() throws IOExce
...
+                        "Position of {}: {}, has reached stopping offset: {}",
+                        tp,
+                        consumerPosition,
+                        stoppingOffset);
+                recordsBySplits.setPartitionStoppingOffset(tp, stoppingOffset);
+                finishSplitAtRecord(
+                        tp, stoppingOffset, consumerPosition, finishedPartitions, recordsBySplits);
             }
             // Track this partition's record lag if it never appears before
             kafkaSourceReaderMetrics.maybeAddRecordsLagMetric(consumer, tp);
```
LinMingQiang, line 142: consumerRecords.partitions().forEach(trackTp -> {
                // Track this partition's record lag if it never appears before
                kafkaSourceReaderMetrics.maybeAddRecordsLagMetric(consumer, trackTp);
            });

LinMingQiang: we do not need to track tp when consumerRecords is empty.

dongwoo6kim: @LinMingQiang, thanks for the review. 
I've changed to track tp, only when records for that tp is not empty.

morazow (approved):
Hey @dongwoo6kim, overall seems okay from my side. But would be possible to add a integration-test ITCase for this case

dongwoo6kim:
Hello @morazow I've added ITCase for this case. 
This case fails for current main branch due to timeout and works okay with fixed code.

morazow (approved):

morazow:
Thanks @dongwoo6kim ,

Tests looks good from my side 👍 

(_Recently I faced similar issue which maybe related, when running batch mode with setting `startingOffsets`. The change should solve that issue. But we may create issue for it_)

dongwoo6kim:
Thanks for confirming @morazow, 
Please feel free to provide any additional advice before merging this fix. 
It would be also helpful if you could elaborate more on the issue you mentioned and consider adding relevant test code for it.

morazow:
Hey @dongwoo6kim, we created another [issue](https://issues.apache.org/jira/browse/FLINK-35565) for it, the solution seems to be similar but let's discuss it again once this PR is merged.

dongwoo6kim:
Hello @morazow, I've added [test code](https://github.com/apache/flink-connector-kafka/blob/f3c9666caa0ef1577402a5c16d57664bf06f51ef/flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableITCase.java#L194) for the mentioned issue, please take a look. 
Test passes with this fixed code and on the latest main branch, it timeouts due to indefinite blocking.

morazow:
Thanks @dongwoo6kim, Looks good!

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -122,32 +122,32 @@ public RecordsWithSplitIds<ConsumerRecord<byte[], byte[]>> fetch() throws IOExce
...
-                            tp,
-                            stoppingOffset,
-                            lastRecord.offset(),
-                            finishedPartitions,
-                            recordsBySplits);
-                }
+            long consumerPosition = consumer.position(tp);
+            // Stop fetching when the consumer's position reaches the stoppingOffset.
+            // Control messages may follow the last record; therefore, using the last record's
+            // offset as a stopping condition could result in indefinite blocking.
```
AHeise, line 130: This is not really clear to me. Are you discussing `if (consumerPosition > stoppingOffset) {`? 

Is the stop offset inclusive or exclusive? It looks like exclusive both in the old and the new version. But that's a bit unintuitive if I would like to read all data currently present in the Kafka topic and this approach discards it.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -122,32 +122,32 @@ public RecordsWithSplitIds<ConsumerRecord<byte[], byte[]>> fetch() throws IOExce
...
+            if (consumerPosition >= stoppingOffset) {
```
AHeise, line 131: Afaik this is equivalent to the old `if (lastRecord.offset() >= stoppingOffset - 1) {`. This code gave the offset of the last method and your new version is using the offset of the next record.

Is this what you wanted? I'm not sure if this is fixing anything then except for the first fix above.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java
```diff
@@ -369,6 +370,38 @@ public void testConsumingTopicWithEmptyPartitions() throws Throwable {
                             WatermarkStrategy.noWatermarks(),
                             "testConsumingTopicWithEmptyPartitions"));
         }
+
+        @Test
+        @Timeout(value = 10)
```
AHeise, line 375: Why are you using Timeout here? We usually discourage the usage as any infra hiccup will result in test failure. Instead we rely on a global timeout where small infra hiccups are absorbed.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java
```diff
@@ -369,6 +370,38 @@ public void testConsumingTopicWithEmptyPartitions() throws Throwable {
...
+        public void testConsumingTransactionalMessage() throws Throwable {
+            String transactionalTopic = "transactionalTopic-" + UUID.randomUUID();
+            KafkaSourceTestEnv.createTestTopic(
+                    transactionalTopic, KafkaSourceTestEnv.NUM_PARTITIONS, 1);
+            List<ProducerRecord<String, Integer>> records =
+                    KafkaSourceTestEnv.getRecordsForTopic(transactionalTopic);
+            // Prepare records for executeAndVerify method
+            records.removeIf(record -> record.partition() > record.value());
```
AHeise, line 383: Why is this necessary?

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java
```diff
@@ -369,6 +370,38 @@ public void testConsumingTopicWithEmptyPartitions() throws Throwable {
...
+                            .setBounded(OffsetsInitializer.latest())
+                            .build();
+            StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
+            env.setParallelism(1);
+            executeAndVerify(
+                    env,
+                    env.fromSource(
+                            source,
+                            WatermarkStrategy.noWatermarks(),
+                            "testConsumingTransactionalMessage"));
```
AHeise, line 403: Is this failing with the old code?

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaTestBase.java
```diff
@@ -270,6 +270,42 @@ public static <K, V> void produceToKafka(
         }
     }
 
+    public static <K, V> void produceTransactionalMessageToKafka(
```
AHeise, line 273: Instead of code replication, how about patching `produceToKafka` to create transactions if transaction.id is set?

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableITCase.java
```diff
@@ -188,6 +190,66 @@ public void testKafkaSourceSink() throws Exception {
...
+        List<Row> values =
+                Arrays.asList(
+                        Row.of(1, 1102, "behavior 1"),
+                        Row.of(2, 1103, "behavior 2"),
+                        Row.of(3, 1104, "behavior 3"));
+        tEnv.fromValues(values).insertInto("kafka").execute().await();
+        // ---------- Delete events from Kafka -------------------
+        Map<Integer, Long> partitionOffsetsToDelete = new HashMap<>();
+        partitionOffsetsToDelete.put(0, 3L);
+        deleteRecords(topic, partitionOffsetsToDelete);
```
AHeise, line 235: Is it the deletion marker that would cause a hang?

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableITCase.java
```diff
@@ -188,6 +190,66 @@ public void testKafkaSourceSink() throws Exception {
...
+        // ---------- Consume stream from Kafka -------------------
+        List<Row> results = new ArrayList<>();
+        assertTimeoutPreemptively(
```
AHeise, line 238: Again please remove all custom timeout logic.

AHeise (commented):
@dongwoo6kim thank you very much for your contribution. I have a couple of remarks.

If you don't have time to fix it, I can also take over because we also hitting that on production.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
dongwoo6kim: As you mentioned, `consumerPosition >= stoppingOffset` is generally equivalent to `lastRecord.offset() >= stoppingOffset - 1`, but there are some edge cases where the new condition works better.

### Edge case
#### Producer setup
Let’s say we are producing transactional messages, and the `lastRecord's offset` is `100`. 
After producing the lastRecord, we commit the transaction, which writes a `marker message` at offset `101`. 
Now, the `nextRecord's offset` becomes `102` (one more than the last message including marker message).

#### Consumer setup
When the flink consumer is reading with `scan.bounded.mode` set to `latest-offset`, the `stoppingOffset` is set to the `nextRecord's offset`, which in this case is `102`.

#### Comparison
In the old condition `(lastRecord.offset() >= stoppingOffset - 1)`, it would compare `100 >= 102 - 1`, which would never be met, leading to indefinite blocking.
However, with the new condition `(consumerPosition >= stoppingOffset)`, it compares `102 >= 102`, which is immediately satisfied. 
This happens because consumer.position() skips the marker message. 
Using lastRecord.offset() in the stopping condition is insufficient in transactional case (when a marker message follows the last record).


| Condition       | As-Is (Old Condition)                             | To-Be (New Condition)                             |
|-----------------|---------------------------------------------------|---------------------------------------------------|
| Expression      | `lastRecord.offset() >= stoppingOffset - 1`        | `consumerPosition >= stoppingOffset`              |
| Example Comparison | Compares `100 >= 102 - 1`, which never meets     | Compares `102 >= 102`, which is immediately satisfied |
| Reason          | The condition fails because it checks the last data record's offset, not accounting for the marker message | The condition succeeds because `consumer.position()` skips the marker message and correctly evaluates the stopping condition |
| Issue           | Leads to indefinite polling                       | Stops polling immediately

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java
dongwoo6kim: I used the Timeout to prevent infinite blocking and fail the test if it enters an infinite loop. 
Since I'm not familiar with the project's global timeout — could you elaborate more about it? 
Where is it configured? 
If global timeout is set then it should be enough but when I tested this method on the current main branch without the Timeout, it didn't fail but just hung.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java
dongwoo6kim: This is preprocessing step before using [executeAndVerify](https://github.com/apache/flink-connector-kafka/blob/7929b16dcfe648da30b6cc9755f63de2ed3d5319/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java#L482) method to verify test result. 

This method expects records from `partition P` should be an integer sequence from `P` to `NUM_RECORDS_PER_PARTITION`. So I deleted records where the value is less than the partition number.

Similar approach to [here](https://github.com/apache/flink-connector-kafka/blob/7929b16dcfe648da30b6cc9755f63de2ed3d5319/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java#L348)

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java
dongwoo6kim: Yes this test fails with the current main branch due to timeout.

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableITCase.java
dongwoo6kim: Yes by deleting message in this step we can cause hanging. 
This test code is for issue [here](https://issues.apache.org/jira/browse/FLINK-35565)

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaTestBase.java
dongwoo6kim: I agree, I've applied your idea. Thanks for the input

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
dongwoo6kim: It is inclusive `(consumerPosition >= stoppingOffset)`. I can't understand about discarding issue, could you elaborate more?

dongwoo6kim:
@AHeise thanks for the feedback! 
I've addressed your comments and applied the suggested changes. When you have a moment, please take a look. Thanks!

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java
AHeise: Could you change the comment to reflect how the data looks afterwards? I'm assuming this will preserve one record per partition?
I'm also curios why we need a specific data layout for this test to work? Or rephrased: what would happen if we retain all records originally generated? Wouldn't the test still assert similar things?

AHeise (commented):
Thank you very much for the swift response and action. Changes look good to me. I will approve and merge once CI passes.

My only concern is that I don't fully understand why we need to modify the data in the test case. Shouldn't a transaction marker always be at the end? What did I miss?

Also if possible could you squash your commits and extend the commit message to include some of the information of the PR description? Basically state the problem and the briefly summarize the solution. Your commit message and PR title is already very descriptive, so it's not necessary for merging and it would just be a nice-to-have.

AHeise:
Seems like the deleted arch unit rule was still needed. What was your intent when you deleted it?

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java
AHeise: The global timeout is applied during CI only. So after X minutes, the CI build will fail. If you test it locally, you have to abort manually (or you use the scripts to run the build as CI does).

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -122,32 +122,32 @@ public RecordsWithSplitIds<ConsumerRecord<byte[], byte[]>> fetch() throws IOExce
...
-                if (lastRecord.offset() >= stoppingOffset - 1) {
-                    recordsBySplits.setPartitionStoppingOffset(tp, stoppingOffset);
-                    finishSplitAtRecord(
-                            tp,
-                            stoppingOffset,
-                            lastRecord.offset(),
-                            finishedPartitions,
-                            recordsBySplits);
-                }
+            long consumerPosition = consumer.position(tp);
```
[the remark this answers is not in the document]
AHeise: (This was just remark for other viewers of this PR).

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
AHeise: I think I understand now: 
* If I use stopping offset `endOffset` and it picks let's say 100, it would abort reading at 100 and sets the stopping offset of the partition to 100.
* The record 100 has been read and is part of the batch. It will not be filtered out.
* So it indeed appears as the last record of the batch.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
AHeise: Thank you very much for the detailed explanation.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java
dongwoo6kim: Got it, thanks for the explanation. I've deleted all custom timeout logic in test codes

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableITCase.java
dongwoo6kim: I've deleted timeout logic thanks for the feedback

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java
dongwoo6kim: I've made two changes. 
1. Replaced `records.removeIf(record -> record.partition() > record.value())` with `KafkaSourceTestEnv.setupEarliestOffsets(transactionalTopic)`, as it serves a similar purpose.
2. Added a comment explaining how the data looks after the setup.

> I'm assuming this will preserve one record per partition?

After data modification, each partition p will contain records from p to NUM_RECORDS_PER_PARTITION (which is 10). For example, partition 1 has records 1 to 10, and partition 5 has records 5 to 10.
 
> what would happen if we retain all records originally generated

The main purpose of this data modification setup is to reuse `executeAndVerify` method. 
When you look at [here](https://github.com/apache/flink-connector-kafka/blob/7929b16dcfe648da30b6cc9755f63de2ed3d5319/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java#L505) `executeAndVerify` method expects the input data to be modified like this way. 

I intended to reuse existing test util functions and follow the test code convention but if you think it is causing unnecessary confusion I can change the test code to have custom assertion logic.

dongwoo6kim:
> Seems like the deleted arch unit rule was still needed. What was your intent when you deleted it?

It was automatically deleted after running `mvn clean verify` locally. I manually rolled back the archunit changes.

dongwoo6kim:
@AHeise Thanks for the feedback. I've left some comments and made updates. Please have a look.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java
AHeise: If it's needed for verification, then all good. We should probably revise the setup then at a later point.

AHeise:
Changes all LGTM. I'm running CI and merge when it's green. 
Thank you very much for this much needed fix!

PR #101, opened 2024-05-02
PoC for supporting unique Kafka producer client ids

francis-a:
This PR came out of debuging a warning we’re seeing in our Flink logs. We’re running Flink 1.18 and have an application that uses Kafka topics as a source and a sink. We’re running with several tasks. The warning we’re seeing in the logs is:

```
WARN org.apache.kafka.common.utils.AppInfoParser - Error registering AppInfo mbean
javax.management.InstanceAlreadyExistsException: kafka.producer:type=app-info,id=kafka producer client id
```

I’ve spent a bit of time debugging, and it looks like the root cause of this warning is the Flink `KafkaSink` creating multiple `KafkaWriter`s that, in turn, create multiple `KafkaProducer`s with the same Kafka producer `client.id`. Since the value for `client.id` is used when registering the `AppInfo` MBean — when multiple `KafkaProducer`s with the same `client.id` are registered we get the above `InstanceAlreadyExistsException`. Since we’re running with several tasks and we get a Kafka producer per task this duplicate registration exception makes sense to me.

This PR proposes a fix that would update the `KafkaSink.builder` by adding a `setClientIdPrefix` method, similar to what we have already on the `KafkaSource.builder`.

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ClientIdFactory.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ClientIdFactory.java
@@ -1,0 +1,14 @@
+package org.apache.flink.connector.kafka.sink;
+
+public class ClientIdFactory {
+    private static final String CLIENT_ID_DELIMITER = "-";
+
+    public static String buildClientId(
+            String clientIdPrefix,
+            int subtaskId
+    ) {
+        return clientIdPrefix
+                + CLIENT_ID_DELIMITER
+                + subtaskId;
+    }
+}
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaSinkBuilder.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaSinkBuilder.java
@@ -71,6 +71,7 @@ public class KafkaSinkBuilder<IN> {
 
     private DeliveryGuarantee deliveryGuarantee = DeliveryGuarantee.NONE;
     private String transactionalIdPrefix = "kafka-sink";
+    private String clientIdPrefix = null;
 
     private final Properties kafkaProducerConfig;
     private KafkaRecordSerializationSchema<IN> recordSerializer;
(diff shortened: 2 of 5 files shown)


SoberChina:
I think this is very necessary. We can start multiple parallelism in the same JVM

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ClientIdFactory.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Construct a Kafka client id in the following format
+     * {@code clientIdPrefix-subtaskId}.
+     *
+     * @param clientIdPrefix prefix for the id
+     * @param subtaskId id of the Kafka producer subtask
+     * @return clientId
+     */
+    public static String buildClientId(
```
AHeise, line 35: Do we need to include checkpoint id? At any given time we have at least 2 producers open.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaSinkBuilder.java
```diff
@@ -202,6 +217,25 @@ private void sanityCheck() {
...
+        String userValue = properties.getProperty(key);
+        if (userValue != null) {
+            LOG.warn(
+                    String.format(
+                            "Property %s is provided but will be overridden from %s to %s",
+                            key, userValue, value));
+        }
+        properties.setProperty(key, value);
+    }
+
```
AHeise, line 238: Is this still needed? looks dead to me.

AHeise (commented):
Thank you very much for your contribution. Changes mostly LGTM. 

I'm wondering if there is a way to test the change. I think a unit test is trivial and doesn't contribute to anything. In the end, we would like to assert that the exception is never thrown but I don't think we can come up with a non-hacky solution. So in this case, I'd add a manual verification in your stead (=I checked the logs of the run and the exception doesn't occur anymore).

AHeise:
QQ: do we need to sanatize/validate the prefix? I don't think we currently do it if you manually set the prefix, so it's probably not needed.

PR #107, opened 2024-07-05
[FLINK-35749] Kafka sink component will lose data when kafka cluster is unavailable for a while

JimmyZZZ:

Fix the bug for losing data during kafka cluster is unavailable for a while.
The related issue: https://issues.apache.org/jira/browse/FLINK-35749 tells how to reproduce this bug.

## Brief change log

The key problem is in WriterCallback#onCompletion of KafkaWriter:
  ```java
                  mailboxExecutor.submit(
                        () -> {
                            // Checking for exceptions from previous writes
                            checkAsyncException();
                        },
                        "Update error metric");
  ```
'mailboxExecutor.submit' without getting future back will not throw exception, which causes the 'asyncProducerException' is assign to null but the job seems like nothing happened. The fix is using 'mailboxExecutor.execute' instead of 'mailboxExecutor.submit'.

## Verifying this change

Added new test case KafkaWriterFaultToleranceITCase to cover this bug.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriter.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriter.java
@@ -449,7 +449,9 @@ class KafkaWriter<IN>
                 }
 
                 // Checking for exceptions from previous writes
-                mailboxExecutor.submit(
+                // Notice: throwing exception in mailboxExecutor thread is not safe enough for
+                // triggering global fail over, which has been fixed in [FLINK-31305].
+                mailboxExecutor.execute(
                         () -> {
                             // Checking for exceptions from previous writes
                             checkAsyncException();
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaWriterFaultToleranceITCase.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaWriterFaultToleranceITCase.java
@@ -37,9 +37,9 @@ import static org.assertj.core.api.AssertionsForClassTypes.assertThatCode;
 @ExtendWith(TestLoggerExtension.class)
 public class KafkaWriterFaultToleranceITCase extends KafkaWriterTestBase {
     private static final String INIT_KAFKA_RETRIES = "0";
-    private static final String INIT_KAFKA_REQUEST_TIMEOUT_MS = "2000";
-    private static final String INIT_KAFKA_MAX_BLOCK_MS = "3000";
-    private static final String INIT_KAFKA_DELIVERY_TIMEOUT_MS = "4000";
+    private static final String INIT_KAFKA_REQUEST_TIMEOUT_MS = "1000";
+    private static final String INIT_KAFKA_MAX_BLOCK_MS = "2000";
+    private static final String INIT_KAFKA_DELIVERY_TIMEOUT_MS = "3000";
 
     @BeforeAll
     public static void beforeAll() {
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaWriterFaultToleranceITCase.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaWriterFaultToleranceITCase.java
@@ -36,7 +36,7 @@ import static org.assertj.core.api.AssertionsForClassTypes.assertThatCode;
 /** Tests for the standalone KafkaWriter in case of fault tolerance. */
 @ExtendWith(TestLoggerExtension.class)
 public class KafkaWriterFaultToleranceITCase extends KafkaWriterTestBase {
-    private static final String INIT_KAFKA_RETRIES = "1";
+    private static final String INIT_KAFKA_RETRIES = "0";
     private static final String INIT_KAFKA_REQUEST_TIMEOUT_MS = "2000";
     private static final String INIT_KAFKA_MAX_BLOCK_MS = "3000";
     private static final String INIT_KAFKA_DELIVERY_TIMEOUT_MS = "4000";
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaWriterFaultToleranceITCase.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaWriterFaultToleranceITCase.java
@@ -21,6 +21,7 @@ import org.apache.flink.connector.base.DeliveryGuarantee;
 import org.apache.flink.metrics.groups.SinkWriterMetricGroup;
 import org.apache.flink.util.TestLoggerExtension;
 
+import org.apache.kafka.common.errors.NetworkException;
 import org.apache.kafka.common.errors.TimeoutException;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
@@ -38,8 +39,8 @@ import static org.assertj.core.api.AssertionsForClassTypes.assertThatCode;
 public class KafkaWriterFaultToleranceITCase extends KafkaWriterTestBase {
     private static final String INIT_KAFKA_RETRIES = "0";
     private static final String INIT_KAFKA_REQUEST_TIMEOUT_MS = "1000";
-    private static final String INIT_KAFKA_MAX_BLOCK_MS = "2000";
-    private static final String INIT_KAFKA_DELIVERY_TIMEOUT_MS = "3000";
+    private static final String INIT_KAFKA_MAX_BLOCK_MS = "1000";
+    private static final String INIT_KAFKA_DELIVERY_TIMEOUT_MS = "1000";
 
     @BeforeAll
     public static void beforeAll() {
@@ -64,7 +65,7 @@ public class KafkaWriterFaultToleranceITCase extends KafkaWriterTestBase {
 
         final KafkaWriter<Integer> writer =
                 createWriterWithConfiguration(
-                        properties, DeliveryGuarantee.EXACTLY_ONCE, metricGroup);
+                        properties, DeliveryGuarantee.AT_LEAST_ONCE, metricGroup);
 
         writer.write(1, SINK_WRITER_CONTEXT);
 
@@ -73,7 +74,7 @@ public class KafkaWriterFaultToleranceITCase extends KafkaWriterTestBase {
         try {
             writer.getCurrentProducer().flush();
             assertThatCode(() -> writer.write(1, SINK_WRITER_CONTEXT))
-                    .hasRootCauseExactlyInstanceOf(TimeoutException.class);
+                    .hasRootCauseExactlyInstanceOf(NetworkException.class);
         } finally {
             KAFKA_CONTAINER.start();
         }
@@ -87,13 +88,13 @@ public class KafkaWriterFaultToleranceITCase extends KafkaWriterTestBase {
 
         final KafkaWriter<Integer> writer =
                 createWriterWithConfiguration(
-                        properties, DeliveryGuarantee.EXACTLY_ONCE, metricGroup);
+                        properties, DeliveryGuarantee.AT_LEAST_ONCE, metricGroup);
         writer.write(1, SINK_WRITER_CONTEXT);
 
         KAFKA_CONTAINER.stop();
         try {
             assertThatCode(() -> writer.flush(false))
-                    .hasRootCauseExactlyInstanceOf(TimeoutException.class);
+                    .hasRootCauseExactlyInstanceOf(NetworkException.class);
         } finally {
             KAFKA_CONTAINER.start();
         }
@@ -107,7 +108,7 @@ public class KafkaWriterFaultToleranceITCase extends KafkaWriterTestBase {
 
         final KafkaWriter<Integer> writer =
                 createWriterWithConfiguration(
-                        properties, DeliveryGuarantee.EXACTLY_ONCE, metricGroup);
+                        properties, DeliveryGuarantee.AT_LEAST_ONCE, metricGroup);
 
         writer.write(1, SINK_WRITER_CONTEXT);
 
@@ -117,11 +118,7 @@ public class KafkaWriterFaultToleranceITCase extends KafkaWriterTestBase {
             writer.getCurrentProducer().flush();
             // closing producer resource throws exception first
             assertThatCode(() -> writer.close())
-                    .hasNoCause()
-                    .hasMessage(
-                            String.format(
-                                    "Timeout expired after %sms while awaiting " + "InitProducerId",
-                                    INIT_KAFKA_MAX_BLOCK_MS));
+                    .hasRootCauseExactlyInstanceOf(NetworkException.class);
         } finally {
             KAFKA_CONTAINER.start();
         }
@@ -135,7 +132,7 @@ public class KafkaWriterFaultToleranceITCase extends KafkaWriterTestBase {
 
         final KafkaWriter<Integer> writer =
                 createWriterWithConfiguration(
-                        properties, DeliveryGuarantee.EXACTLY_ONCE, sinkInitContext);
+                        properties, DeliveryGuarantee.AT_LEAST_ONCE, sinkInitContext);
 
         KAFKA_CONTAINER.stop();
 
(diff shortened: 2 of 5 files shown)


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriter.java
```diff
@@ -449,12 +460,17 @@ public void onCompletion(RecordMetadata metadata, Exception exception) {
                 }
 
                 // Checking for exceptions from previous writes
-                mailboxExecutor.submit(
+                // Notice: throwing exception in mailboxExecutor thread is not safe enough for
+                // triggering global
+                // fail over, which has been fixed in [FLINK-31305]. And using
+                // mailboxExecutor.execute() is better than
+                // mailboxExecutor.submit() since submit will swallow exceptions inside.
+                mailboxExecutor.execute(
```
AHeise, line 468: Yes, this is the actual fix. Using execute instead of submit. This bug got introduced through FLINK-31305.

Note: `submit` does not swallow exceptions. It's an anti-pattern to use `submit` without looking at the Future. The future holds the result/exception to be handled async somewhere. If we bubble up the exception the task dies without the call-site being able to properly react to it. 
So please remove the second part of your comment from "And using ...".

I will amend the javadoc of the mailbox executor to make it clear that the executor indeed does not bubble up exception in `submit`.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriter.java
```diff
@@ -411,11 +411,22 @@ private void registerMetricSync() {
...
+            // chance to make
+            // asyncProducerException be null before the function flush() and write() seeing the
+            // exception.
+            // After that, the flush and writer functions can continue to write next message batch
+            // to kafka, the old
+            // in-flight message batch with exceptions get lost.
+            // We hope that once some exceptions thrown during sending kafka, KafkaWriter could be
+            // fast-failed and
+            // trigger global restarting, then asyncProducerException will be initialized as null
+            // again.
```
AHeise, line 429: I don't think this change is necessary at all. You can see that we now have inflated error count metrics in all the tests that you adjusted (and should be reverted as well).

The issue is not that we reset the exception. It's that it doesn't bubble up (see below). We actually have a clear contract on the volatile `asyncProducerException`: it's set in the `onCompletion` and read+reset in the main thread (which is the mailbox thread). So we never have a chance to concurrently reset the exception and lose it somehow.

AHeise (changes requested):
Thank you very much for your contribution. The actual fix looks good but you have some additional changes that are not necessary imho.

We also need to add a test case for the fix. In your description, you said it's easy to simulate when you set retries to a low value (don't use 0) and then Kafka becomes unavailable. So please add a test to `KafkaWriterITCase` where you `KAFKA_CONTAINER.stop()` and check for the result. Don't forget to `start` it again in a finally block though. It might even be safer to create a new ITCase like KafkaWriterFaultToleranceITCase where you have more control over the container.

AHeise:
Please also add some more details to the PR description and commit message in accordance with the [guidlines](https://flink.apache.org/how-to-contribute/reviewing-prs/). In particular, please use the proper Github PR template.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriter.java
JimmyZZZ: I agree. My description is not exactly, if we use submit and get the future returned, the exception will also be thrown. 

I'll remove this part: "And using ...".

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriter.java
JimmyZZZ: Thanks for explaining, and I want to discuss this part with you for more detail.

As we can see, the read-reset-throw logic is in checkAsyncException(), and there 4 parent method may invoke checkAsyncException():
1. writer()  2. flush()  3. close()  4. WriterCallback#onCompletion

I agree with "we never have a chance to concurrently reset the exception ", but 1/2/3 have the chance to reset the exception  before 4, since WriterCallback#onCompletion is in asynchronous threads and assigning exception to asyncProducerException is not in the mailboxExecutor, which will expose the exception to 1/2/3 before mailboxExecutor.execute.

JimmyZZZ: As you mentioned in another comment: " If we bubble up the exception the task dies without the call-site being able to properly react to it."

I'm a little confused: 
If we hope to bubble up the exception only in main thread with mailboxExecutor.execute, the problem of [FLINK-31305](https://issues.apache.org/jira/browse/FLINK-31305) will occur again: 
in-flight checkpoint will be completed, once the mailboxExecutor.execute throw the exception a little late, then the job restarts and recovers from this error checkpoint

Throwing exception depending on mailboxExecutor.execute is not safe enough since mailboxExecutor thread and checkpoint thread are different. So I agree with the fix of FLINK-31305 bubbling up the exception in 1.writer() 2. flush() 3. close(), which can  trigger global fail over ASAP. But in FLINK-31305, the exception may be reset by mailboxExecutor.execute, 1/2/3 can not see the exception and the in-flight checkpoint with kafka sending problems may be completed. That's  why I remove the logic of resetting to null.

If there is any mistake in my understanding, please point it out to me, Thanks.  @AHeise

JimmyZZZ: Thanks for remind, I need some time to think over about this part

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriter.java
JimmyZZZ: @AHeise  removed now

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriter.java
JimmyZZZ: @AHeise  I've remove the modification before about the reset logic in checkAsyncException(), and i understand what you mean about "we never have a chance to concurrently reset the exception" and agree with you.

Btw, I'm still little curious: if we don't reset the asyncProducerException to null in checkAsyncException(), what problems we will meet? As the comment from [FLINK-31305] : Propagate the first exception since amount of exceptions could be large.  Is this the only reason that we throw the Exception for only one time?  This seems not so persuasive for me.

JimmyZZZ:
> Please also add some more details to the PR description and commit message in accordance with the [guidlines](https://flink.apache.org/how-to-contribute/reviewing-prs/). In particular, please use the proper Github PR template.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriter.java
JimmyZZZ: If we keep the asyncProducerException until KafkaWriter task is restarted and assign to asyncProducerException null again, that seems safer if someone modifies the related logic later...

JimmyZZZ:
@AHeise  added new test case KafkaWriterFaultToleranceITCase and do some refactor to extract some public things for KafkaWriterFaultToleranceITCase and KafkaWriterITCase, and also provided more details in this pr as Github PR template.

Pls help to review again, thanks very much.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriter.java
mas-chen: Good catch, thanks for fixing this!

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriter.java
AHeise: It's a good point to raise if we need to reset at all or not. I'm assuming @mas-chen did it to emulate the previous behavior.

From what I can see, the only real downside is that we rethrow the exception in close (which is always called also in case of errors). That could prevent some proper cleanup and will count exceptions twice in most cases, which will inflate the metrics. I think the latter point has quite a bit of side-effects:

I'm expecting most customers to have some alerts on the exceptions metrics, which could mean that it's a behavior change that would alert more users than in previous versions. So in the worst case, some poor data engineer will be paged in the middle of the night because of that. And I would expect a few alerts to be changed to be less noisy.

All in all, I think that resetting the exception would be good IF we are all sure that there won't be any exception missed. Being sure that we can't have corrupted state is of course more important than having accurate metrics but ideally we have both.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaWriterFaultToleranceITCase.java
```diff
@@ -0,0 +1,170 @@
...
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThatCode;
+
+/** Tests for the standalone KafkaWriter in case of fault tolerance. */
+@ExtendWith(TestLoggerExtension.class)
+public class KafkaWriterFaultToleranceITCase extends KafkaWriterTestBase {
+    private static final String INIT_KAFKA_RETRIES = "1";
+    private static final String INIT_KAFKA_REQUEST_TIMEOUT_MS = "2000";
+    private static final String INIT_KAFKA_MAX_BLOCK_MS = "3000";
+    private static final String INIT_KAFKA_DELIVERY_TIMEOUT_MS = "4000";
```
AHeise, line 42: Tests look good. I'm wondering if we can/need to tweak these numbers to make the test hopefully not flaky.

We have to make good tradeoffs here:
- The overall test shouldn't take too long and we are relying on timeouts here.
- Timeouts shouldn't be too small to compensate for infra hiccups.
- Retries should be non-zero.

On first glance, the values all look good in that regard.

However, now at a second thought: do we ever want to have a successful request in this test? Maybe we could set really low timeouts and retry = 0 and provoke the failures even quicker?

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaWriterTestBase.java
```diff
@@ -0,0 +1,288 @@
...
+import java.util.PriorityQueue;
+import java.util.Properties;
+import java.util.concurrent.ScheduledFuture;
+import java.util.function.Consumer;
+
+import static org.apache.flink.connector.kafka.testutils.DockerImageVersions.KAFKA;
+import static org.apache.flink.connector.kafka.testutils.KafkaUtil.createKafkaContainer;
+
+/** Test base for KafkaWriter. */
+public abstract class KafkaWriterTestBase {
```
AHeise, line 62: Thank you very much for avoiding duplicate code.

AHeise (approved):
LAGTM. Production code is good. Test code is exhaustive and still minimal. I just added a remark on the producer settings in the test. We may be able to speed up the tests significantly. PTAL.

Do you need a committer that pushes the code for you? LMK if that's the case. I still need to get back into the community to understand who is committer and who is contributor ;).

JimmyZZZ:
> LAGTM. Production code is good. Test code is exhaustive and still minimal. I just added a remark on the producer settings in the test. We may be able to speed up the tests significantly. PTAL.
> 
> Do you need a committer that pushes the code for you? LMK if that's the case. I still need to get back into the community to understand who is committer and who is contributor ;).

@AHeise  Yes, I need your help to reach to someone can pushes the code since I'm also not very sure who can do it. Thanks a lot.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaWriterFaultToleranceITCase.java
JimmyZZZ: Since we want to mock the scene of kafka unavailable, I'm worrying about the bad network of the test machine will also cause timeout too, so I keep other timeout thresholds at least 1 second here

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriter.java
JimmyZZZ: Very good consideration for "data engineer will be paged in the middle of the night" once the alerts on the exceptions metrics changes. Then I think we need to be very cautious for the following modification about the exception logic. And the new test KafkaWriterFaultToleranceITCase can play a role for this concerns.

AHeise:
There is an arch unit rule violation:

```
2024-07-09T13:48:15.4851704Z [ERROR]   Architecture Violation [Priority: MEDIUM] - Rule 'ITCASE tests should use a MiniCluster resource or extension' was violated (1 times):
2024-07-09T13:48:15.4853792Z org.apache.flink.connector.kafka.sink.KafkaWriterFaultToleranceITCase does not satisfy: only one of the following predicates match:
2024-07-09T13:48:15.4856292Z * reside in a package 'org.apache.flink.runtime.*' and contain any fields that are static, final, and of type InternalMiniClusterExtension and annotated with @RegisterExtension
2024-07-09T13:48:15.4859585Z * reside outside of package 'org.apache.flink.runtime.*' and contain any fields that are static, final, and of type MiniClusterExtension and annotated with @RegisterExtension or are , and of type MiniClusterTestEnvironment and annotated with @TestEnv
2024-07-09T13:48:15.4862798Z * reside in a package 'org.apache.flink.runtime.*' and is annotated with @ExtendWith with class InternalMiniClusterExtension
2024-07-09T13:48:15.4864813Z * reside outside of package 'org.apache.flink.runtime.*' and is annotated with @ExtendWith with class MiniClusterExtension
2024-07-09T13:48:15.4871237Z  or contain any fields that are public, static, and of type MiniClusterWithClientResource and final and annotated with @ClassRule or contain any fields that is of type MiniClusterWithClientResource and public and final and not static and annotated with @Rule
```

JimmyZZZ:
Added KafkaWriterFaultToleranceITCase to the known archunit violations, pls take a look. @AHeise

AHeise:
At this point, we just try to get the build green. You can check locally by running `mvn verify`.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriter.java
AHeise: As a follow-up, I opened and resolved https://issues.apache.org/jira/browse/FLINK-35796.

JimmyZZZ:
> At this point, we just try to get the build green. You can check locally by running `mvn verify`.

push another commit for the violations issue as the latest ci failed again , should be fixed now.

JimmyZZZ:
> At this point, we just try to get the build green. You can check locally by running `mvn verify`.

checked, should be green now.

AHeise:
> > At this point, we just try to get the build green. You can check locally by running `mvn verify`.
> 
> checked, should be green now.

Retriggering CI.

JimmyZZZ:
@AHeise  It's strange that only this test failed:  compile_and_test (1.17.2, 8, 11) / compile_and_test (8)

I digged into it and found two Test modified in this PR are green:
2024-07-11T06:44:35.4459342Z [INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 37.521 s - in org.apache.flink.connector.kafka.sink.KafkaWriterFaultToleranceITCase
2024-07-11T06:43:56.5827523Z [INFO] Tests run: 19, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 14.883 s - in org.apache.flink.connector.kafka.sink.KafkaWriterITCase

and seems something wrong abount the env or network because I found this log in the failed test: 2024-07-11T07:19:03.3807103Z ##[error]The action 'Compile and test' has timed out after 50 minutes.

So could you help to take a look or just trigger again

AHeise:
The hanging test comes from the source `DynamicKafkaSourceITTest.IntegrationTests.testIdleReader`. I don't see any sink usage in the test, so the failure is unrelated.

PR #109, opened 2024-07-23
[FLINK-22748] Allow dynamic target topic selection in SQL Kafka sinks

klam-shop:
Allows writing to different Kafka topics based on the `topic` metadata column value in SQL, and updates the Table API's `KafkaDynamicSink` to accept a `List<String> topics` instead of `String topic` as well as support `topic-pattern`. The list acts as an allow-list of acceptable values for the `topic` metadata column. `topic-pattern` for sinks is a pattern that must match `topic` metadata column values, or else an error is thrown.

- If a single topic is provided, it is used by default for the target topic to produce to
- If a list is provided, only that list of topics can be produced to
- If a topic pattern is provided, it must match `topic` metadata column values

Builds on the work in https://github.com/apache/flink/pull/16142 by @SteNicholas 

## Brief change log
- Adds `topic` as writable metadata
- Adds `topic-pattern` support, mirror source side
- Updates Table API Kafka sink to accept `List<String> topics` instead of `String topic`, mirroring the source side.
- Implements allow-list and match behaviour in `DynamicKafkaRecordSerializationSchema`
- Updates Upsert Kafka Sink to also support reading and writing multiple topics using the topic list

## Verifying this change
- [x] Tests that the Sink Factory and related machinery works as expected
- [x] Tests the various valid and invalid scenarios for `topic` metadata in `DynamicKafkaRecordSerializationSchema`

- The public API, i.e., is any changed class annotated with `@Public(Evolving)`: no
- The serializers: yes

## Documentation

- Does this pull request introduce a new feature? yes
- If yes, how is the feature documented? updated documentation


flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchema.java
```diff
@@ -144,14 +147,34 @@ public void open(
...
     }
 
+    private String getTargetTopic(RowData element) {
+        final String topic = readMetadata(element, KafkaDynamicSink.WritableMetadata.TOPIC);
+        if (topic == null && topics == null) {
+            throw new IllegalArgumentException(
+                    "The topic of the sink record is not valid. Expected a single topic but no topic is set.");
+        } else if (topic == null && topics.size() == 1) {
+            return topics.get(0);
+        } else if (topics != null && topics.size() > 0 && !topics.contains(topic)) {
```
klam-shop, line 157: Open to discuss: I decided to keep topics as a List since, but we can change it to HashSet. I am assuming the list will be short on average.

List can outperform HashSet for small number of elements:
https://stackoverflow.com/questions/150750/hashset-vs-list-performance

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchema.java
```diff
@@ -144,14 +147,34 @@ public void open(
         valueSerialization.open(context);
     }
 
+    private String getTargetTopic(RowData element) {
+        final String topic = readMetadata(element, KafkaDynamicSink.WritableMetadata.TOPIC);
+        if (topic == null && topics == null) {
+            throw new IllegalArgumentException(
+                    "The topic of the sink record is not valid. Expected a single topic but no topic is set.");
+        } else if (topic == null && topics.size() == 1) {
+            return topics.get(0);
```
dmariassy, line 156: Can we start wit the `topics.size == 1` check and immediately return if the connector is used with a static target topic? This could save the processing in `readMetadata`.

docs/content/docs/connectors/table/kafka.md
```diff
@@ -196,11 +196,11 @@ Connector Options
...
-      <td>required for sink</td>
+      <td>optional</td>
       <td>yes</td>
       <td style="word-wrap: break-word;">(none)</td>
       <td>String</td>
-      <td>Topic name(s) to read data from when the table is used as source. It also supports topic list for source by separating topic by semicolon like <code>'topic-1;topic-2'</code>. Note, only one of "topic-pattern" and "topic" can be specified for sources. When the table is used as sink, the topic name is the topic to write data to. Note topic list is not supported for sinks.</td>
+      <td>Topic name(s) to read data from when the table is used as source, or allowed topics for writing when the table is used as sink. It also supports topic list for source by separating topic by semicolon like <code>'topic-1;topic-2'</code>. Note, only one of "topic-pattern" and "topic" can be specified for sources. For sinks, the topic or list of topics provided are they only valid values (whitelist) that can be written to the `topic` metadata column for determining what topic to produce to. If only a single topic is provided, this is the default topic to produce to. If not provided, any value of `topic` metadata is permitted.</td>
```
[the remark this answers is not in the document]
klam-shop: Good call, fixed.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchema.java
klam-shop: Good idea, I thought about it and thought it might make the semantics a bit complicated, because if we do this, then for `topics.size == 1` then I think the `topic` metadata column should not be writable, but I think it's worth saving the processing. Updated

dmariassy: > then for topics.size == 1 the topic metadata column should not be writable

I think that makes a lot of sense!

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaConnectorOptions.java
```diff
@@ -118,8 +118,8 @@ public class KafkaConnectorOptions {
                     .asList()
                     .noDefaultValue()
                     .withDescription(
-                            "Topic names from which the table is read. Either 'topic' or 'topic-pattern' must be set for source. "
-                                    + "Option 'topic' is required for sink.");
+                            "Topic name(s) to read data from when the table is used as source. It also supports topic list for source by separating topic by semicolon like 'topic-1;topic-2'. Note, only one of 'topic-pattern' and 'topic' can be specified for sources. "
+                                    + "When the table is used as sink, the topic name is the topic to write data. It also supports topic list for sinks. The provided topic-list is treated as a allow list of valid values for the `topic` metadata column, if not provided, any value is valid. If the 'topic' option isn't specified, or a list is provided, for sink table, 'topic' metadata column is writable and must be specified.");
```
rich-c-shop, line 122: > if not provided, any value is valid

how about supporting `topic-pattern` in sink as well? This would align with source and `*` wildcard pattern stands for "any value is valid".

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
```diff
@@ -251,6 +251,8 @@ public DataStreamSink<?> consumeDataStream(
     public Map<String, DataType> listWritableMetadata() {
         final Map<String, DataType> metadataMap = new LinkedHashMap<>();
         Stream.of(WritableMetadata.values())
+                // When `topic` is a singleton list, TOPIC metadata is not writable
```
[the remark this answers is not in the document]
klam-shop: There will be an error, mentioning that TOPIC is not in the writable metadata list. Hit it when I tested locally

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchema.java
```diff
@@ -144,14 +147,36 @@ public void open(
         valueSerialization.open(context);
     }
 
+    private String getTargetTopic(RowData element) {
+        if (topics != null && topics.size() == 1) {
+            // If topics is a singleton list, we only return the provided topic.
+            return topics.get(0);
+        }
+        final String topic = readMetadata(element, KafkaDynamicSink.WritableMetadata.TOPIC);
+        if (topic == null && topics == null) {
```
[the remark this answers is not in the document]
klam-shop: For an empty list, all `topic` metadata will not be allowed. Though I'm not sure how you'd specify that since `""` -> `null`, I guess the user might specify something like `";"`

klam-shop:
@fapaul do you think you have time to take a look? Or pull in folks who may be better reviewers?

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaConnectorOptions.java
klam-shop: After discussion, we thought the `topic-pattern` would be a more intuitive semantic than `topic=null` for supporting 'any value' for topic metadata column.

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicTableFactoryTest.java
```diff
@@ -616,6 +617,10 @@ public void testTableSink() {
         final SinkV2Provider sinkProvider = (SinkV2Provider) provider;
         final Sink<RowData> sinkFunction = sinkProvider.createSink();
         assertThat(sinkFunction).isInstanceOf(KafkaSink.class);
+        assertThat(actualKafkaSink.listWritableMetadata())
```
dmariassy, line 620: 👍🏻

dmariassy (approved):
Great work! 🚀

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchemaTest.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.List;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/** Tests for {@link DynamicKafkaRecordSerializationSchema}. */
+public class DynamicKafkaRecordSerializationSchemaTest extends TestLogger {
+    private static final List<String> TOPICS = Arrays.asList("topic1;topic2".split(";"));
```
AHeise, line 26:
```suggestion
    private static final List<String> MULTIPLE_TOPICS = Arrays.asList("topic1", "topic2");
```

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchemaTest.java
```diff
@@ -0,0 +1,147 @@
...
+    private static final String TOPIC = "topic";
```
AHeise, line 27:
```suggestion
    private static final String SINGLE_TOPIC = "topic";
```

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchemaTest.java
```diff
@@ -0,0 +1,147 @@
...
+            List<String> topics, Pattern topicPattern, String rowTopic, String expectedTopic) {
+        GenericRowData rowData = createRowData(rowTopic);
+        DynamicKafkaRecordSerializationSchema schema = createSchema(topics, topicPattern);
+        KafkaRecordSerializationSchema.KafkaSinkContext context = createContext();
+
+        // Call serialize method
+        ProducerRecord<byte[], byte[]> record = schema.serialize(rowData, context, null);
+
+        // Assert the returned ProducerRecord is routed to the correct topic
+        assertEquals(record.topic(), expectedTopic);
```
AHeise, line 42: I'm not sure what the default is now, but you mix assertj and junit5 assertions in this test. How about using just assertj?

`assertThat(record.topic()).isEqualTo(expectedTopic)`

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicTableFactoryTest.java
```diff
@@ -1002,39 +1081,22 @@ public void testExactlyOnceGuaranteeWithoutTransactionalIdPrefix() {
...
-                        getBasicSinkOptions(),
-                        options -> options.put("topic-pattern", TOPIC_REGEX));
-
-        try {
-            createTableSink(SCHEMA, modifiedOptions);
-        } catch (Throwable t) {
-            assertThat(t.getCause().getMessage())
-                    .isEqualTo(String.format(errorMessageTemp, "'topic-pattern'", TOPIC_REGEX));
+            assertThat(t.getCause().getMessage()).isEqualTo(errorMessageTemp);
         }
```
AHeise, line 1100: Can we switch to `assertThatThrownBy`?

AHeise (commented):
Most changes look good. My main concern is the performance when using topic patterns.

I'm also curious if we can add an ITCase that tests everything jointly.
I think we should create a DynamicKafkaSinkITCase that really shows the whole story. `KafkaSinkITCase#testWriteRecordsToKafka` should contain most of the code pointers needed to setup the infrastructure. I'd just would like to see a test on Table API that has a couple of records with different target topics and we then use `drainAllRecordsFromTopic` to verify that all of the records are routed as expected.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
```diff
@@ -393,6 +409,20 @@ private RowData.FieldGetter[] getFieldGetters(
     // --------------------------------------------------------------------------------------------
 
     enum WritableMetadata {
+        TOPIC(
+                "topic",
+                DataTypes.STRING().nullable(),
```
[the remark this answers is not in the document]
klam-shop: Good catch, we cannot set topic to null when it is writable

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
```diff
@@ -251,6 +258,12 @@ public DataStreamSink<?> consumeDataStream(
     public Map<String, DataType> listWritableMetadata() {
         final Map<String, DataType> metadataMap = new LinkedHashMap<>();
         Stream.of(WritableMetadata.values())
+                // When `topic` is a singleton list, TOPIC metadata is not writable
+                .filter(
+                        m ->
+                                topics == null
+                                        || topics.size() > 1
+                                        || !WritableMetadata.TOPIC.key.equals(m.key))
```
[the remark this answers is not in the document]
klam-shop: Changed to for-loop

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaConnectorOptionsUtil.java
```diff
@@ -636,21 +618,25 @@ public static DynamicTableFactory.Context autoCompleteSchemaRegistrySubject(
...
-        final Optional<String> keyFormat = configuration.getOptional(KEY_FORMAT);
-        final Optional<String> format = configuration.getOptional(FORMAT);
-        final String topic = configuration.get(TOPIC).get(0);
-
-        if (format.isPresent() && SCHEMA_REGISTRY_FORMATS.contains(format.get())) {
-            autoCompleteSubject(configuration, format.get(), topic + "-value");
-        } else if (valueFormat.isPresent() && SCHEMA_REGISTRY_FORMATS.contains(valueFormat.get())) {
-            autoCompleteSubject(configuration, "value." + valueFormat.get(), topic + "-value");
-        }
+        // the subject autoComplete should only be used in sink with a single topic, check the topic
```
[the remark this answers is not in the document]
klam-shop: I think to make the subject autoComplete work with the SCHEMA_REGISTRY_FORMATS, we need to have a follow-up issue to have the formats handle multiple subjects dynamically. 

Writing to a static subject for many topics would not match with the Confluent Schema registry default [TopicNameStrategy](https://docs.confluent.io/platform/current/schema-registry/fundamentals/serdes-develop/index.html#subject-name-strategy) which derives its subject from the topic name, eg. `topic-value` and `topic-key` subjects for topic `topic`. I decided to disable the logic here for when there are multiple topics being produced to, and leave the extension for future work.

klam-shop:
Thanks for your review @AHeise! I added integration tests into `KafkaTableITCase.java` since there are other Kafka Table Sink and Source IT test cases there, and it seemed like the appropriate place. I'm still working on running these tests locally, as the container setup does not work out of the box with `podman`. I have `podman` available but not `docker`. 

Otherwise, I have tried to address your other comments as well.

klam-shop:
> I'm still working on running these tests locally, as the container setup does not work out of the box with podman. I have podman available but not docker.

Tested `KafkaTableITCase.java` using GH Codespaces ✅

klam-shop:

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableITCase.java
```diff
@@ -188,6 +188,120 @@ public void testKafkaSourceSink() throws Exception {
...
+        tEnv.fromValues(values).insertInto("kafka").execute().await();
+
+        // ---------- Consume stream from Kafka -------------------
+
+        List<Row> results = collectAllRows(tEnv.sqlQuery("SELECT * from kafka"));
+
+        assertThat(results)
+                .containsExactly(
+                        Row.of(topic1, 1, 1102, "behavior 1"),
+                        Row.of(topic2, 2, 1103, "behavior 2"));
```
AHeise, line 240: This assertion is too unspecific for me. Records could be written into only one of the topics.

I'd add another more fine-grain assertion with `drainAllRecordsFromTopic` where you really check that the records are written to the respective topics correctly.

Btw for this assertion you could also simply `assertThat(results).containsExactlyInAnyOrder(values)`. Also note the `InAnyOrder` because we have no guarantee that the order is maintained.

AHeise (commented):
Sorry I was on vacation.

LAGTM with some nits on the newly added IT.

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableITCase.java
klam-shop: Fair points, added assertions to check the specific topics and also use `containsExactlyInAnyOrder`

AHeise:
LGTM from my side, I'd approve once CI passes. Thank you very much for your contribution!

klam-shop:
Thanks for your review @AHeise! I see CI is failing, I'll look into it.

klam-shop:
@AHeise CI has passed now: https://github.com/Shopify/flink-connector-kafka/actions/runs/10639144193

Please take a look when you have a moment, thanks! 

image

klam-shop:
Friendly ping, @AHeise 🙏

AHeise:
> Friendly ping, @AHeise 🙏

I'm running CI one last time and then approve/merge.

klam-shop:
Thanks @AHeise !

AHeise (approved):
LGTM. Thank you very much for your contribution.

klam-shop:
Thank you for your review. I think we can update https://issues.apache.org/jira/browse/FLINK-22748

PR #110, opened 2024-07-28
[FLINK-35870] Change DEFAULT_KAFKA_TRANSACTION_TIMEOUT in KafkaSinkBuilder

twosom:
change `DEFAULT_KAFKA_TRANSACTION_TIMEOUT` in `KafkaSinkBuilder` class.

The default value is 15 minutes, as shown in the official Kakfa documentation.

So I changed it from 1 hour to 15 minutes and added the test.
스크린샷 2024-07-28 오후 1 56 35

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaSinkBuilder.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaSinkBuilder.java
@@ -61,7 +61,7 @@ import static org.apache.flink.util.Preconditions.checkState;
 public class KafkaSinkBuilder<IN> {
 
     private static final Logger LOG = LoggerFactory.getLogger(KafkaSinkBuilder.class);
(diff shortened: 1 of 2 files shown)


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaSinkBuilder.java
```diff
@@ -61,7 +61,7 @@
 public class KafkaSinkBuilder<IN> {
 
     private static final Logger LOG = LoggerFactory.getLogger(KafkaSinkBuilder.class);
-    private static final Duration DEFAULT_KAFKA_TRANSACTION_TIMEOUT = Duration.ofHours(1);
+    private static final Duration DEFAULT_KAFKA_TRANSACTION_TIMEOUT = Duration.ofMinutes(15);
```
davidradl, line 64: It seems a reasonable change, but I am concerned about changing the default of an option. The cautious approach would be to introduce a new parameter with the new default, so the user can choose when to move to the new default and there are no unexpected changes in behaviour for minor upgrades.

PR #115, opened 2024-08-29
[FLINK-36176] Remove support for Kafka 0.1

AHeise:
Warp up FLINK-19152 by also removing all related producer code.

Move FlinkKafkaProducerBase.getPropertiesFromBrokerList to a new test utility class as this method is used by some tests.


flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducerBase.java
```diff
@@ -1,442 +0,0 @@
...
- * Flink Sink to produce data into a Kafka topic.
- *
- * <p>Please note that this producer provides at-least-once reliability guarantees when checkpoints
- * are enabled and setFlushOnCheckpoint(true) is set. Otherwise, the producer doesn't provide any
- * reliability guarantees.
- *
- * @param <IN> Type of the messages to write into Kafka.
- */
-@Internal
-public abstract class FlinkKafkaProducerBase<IN> extends RichSinkFunction<IN>
```
fapaul, line 72: For my sanity: Nothing in this repo actually extended this class and it was unused.

fapaul (approved):
LGTM although I am slightly confused if users could have used the classes removed by this PR

AHeise:
> LGTM although I am slightly confused if users could have used the classes removed by this PR

It was originally used by the universal producer (FlinkKafkaProducer) and the FlinkKafkaProducer01(1). It was already refactored some years ago and only left in to be able to read the serialized state of that old producer.

The base class is annotated as @Internal. The `FlinkKafkaProducer011` is not marked at all but the java doc clearly states that this is just for migration.

So I have a hard time to imagine anyone using this class and being surprised that it doesn't work anymore.

PR #119, opened 2024-09-16
[FLINK-36278] Reduce log size

AHeise:
Currently, container logs appear under an o.a.f logger and thus are visible on CI. This results in compressed log size >40MB for a run and often leads to download errors.

This PR reroutes container logs to a special container logger. It also uses a custom format to significantly reduce the size of each log line. The logs for containers are disabled by default.

Here is the original log of an ITCase (308 KB)
[old.log](https://github.com/user-attachments/files/17010325/old.log)

```
2794 [docker-java-stream--2048098242] INFO  org.apache.flink.streaming.connectors.kafka.KafkaTestEnvironmentImpl [] - [Kafka-0] STDOUT: [2024-09-16 08:26:53,236] INFO Server environment:java.library.path=/usr/java/packages/lib:/lib:/usr/lib:/usr/lib64:/lib64 (org.apache.zookeeper.server.ZooKeeperServer)
2794 [docker-java-stream--2048098242] INFO  org.apache.flink.streaming.connectors.kafka.KafkaTestEnvironmentImpl [] - [Kafka-0] STDOUT: [2024-09-16 08:26:53,236] INFO Server environment:java.io.tmpdir=/tmp (org.apache.zookeeper.server.ZooKeeperServer)
2794 [docker-java-stream--2048098242] INFO  org.apache.flink.streaming.connectors.kafka.KafkaTestEnvironmentImpl [] - [Kafka-0] STDOUT: [2024-09-16 08:26:53,236] INFO Server environment:java.compiler=<NA> (org.apache.zookeeper.server.ZooKeeperServer)
```

New formatter, same level (199 KB)
[with container.log](https://github.com/user-attachments/files/17010326/with.container.log)

```
[Kafka-0] [2024-09-16 08:18:25,496] INFO Server environment:java.library.path=/usr/java/packages/lib:/lib:/usr/lib:/usr/lib64:/lib64 (org.apache.zookeeper.server.ZooKeeperServer)
[Kafka-0] [2024-09-16 08:18:25,496] INFO Server environment:java.io.tmpdir=/tmp (org.apache.zookeeper.server.ZooKeeperServer)
[Kafka-0] [2024-09-16 08:18:25,496] INFO Server environment:java.compiler=<NA> (org.apache.zookeeper.server.ZooKeeperServer)

```

Disabled container log (113 KB)
[no container.log](https://github.com/user-attachments/files/17010324/no.container.log)

### DIFF
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/testutils/KafkaUtil.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/testutils/KafkaUtil.java
@@ -18,8 +18,6 @@
 
 package org.apache.flink.connector.kafka.testutils;
 
-import org.apache.flink.util.StringUtils;
-
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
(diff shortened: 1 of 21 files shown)


fapaul:
Looking at the current test failures of this PR the log size (https://github.com/apache/flink-connector-kafka/actions/runs/10881066111/job/30189211552?pr=119) is still 50MB+. I wonder if this PR fixes the issue.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/testutils/KafkaUtil.java
```diff
@@ -81,24 +89,19 @@ public static KafkaContainer createKafkaContainer(
...
-                .withEnv(
-                        "KAFKA_TRANSACTION_MAX_TIMEOUT_MS",
-                        String.valueOf(Duration.ofHours(2).toMillis()))
-                .withEnv("KAFKA_LOG4J_TOOLS_ROOT_LOGLEVEL", logLevel)
-                .withLogConsumer(logConsumer);
+    public static Logger getLogger(String containerName) {
+        return LoggerFactory.getLogger("container." + containerName);
+    }
+
+    public static Logger getLogger(String type, Class<?> testClass) {
```
fapaul, line 99: Afaict this method is always called with `"flink"` as a first parameter. WDY about removing the type parameter and always set it to `flink` when called with a testClass.

flink-connector-kafka/src/test/resources/log4j2-test.properties
```diff
@@ -27,15 +27,41 @@ appender.testlogger.target = SYSTEM_ERR
...
+logger.flinkconnector.name = org.apache.flink.connector
+logger.flinkconnector.level = OFF # INFO/DEBUG for starting debugging
+
+# Kafka producer and consumer level
+logger.kafka.name = org.apache.kafka
 logger.kafka.level = OFF
-logger.kafka2.name = state.change
+
+# For old school kafka test that spawn the test server in the same JVM
+logger.kafka2.name = kafka
```
fapaul, line 41: Which tests still spawn kafka in the same JVM?

fapaul (approved):
Thanks for driving the investigation to an end. Only one minor comment regarding the APi of the `KafkaUtil`

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/testutils/KafkaUtil.java
AHeise: Good idea. I'll rename to getFlinkLogger then.

flink-connector-kafka/src/test/resources/log4j2-test.properties
AHeise: Errr it seems none. Good catch.

We still have test dependencies on kafka server and zookeeper that I can probably remove and with it these log confs.

PR #121, opened 2024-09-26
[FLINK-36177] Deprecate KafkaShuffle and more

AHeise:
This commit deprecates all classes that are slated for removal in the kafka-4.0 release compatible with Flink 2.0. I also deprecated internal classes to make later removal easier. Some public classes will cease to be public API but are still internally used.

Also introduce KafkaPartitioner to replace FlinkKafkaPartitioner. Relocate FlinkKafkaPartitioner to KafkaSink package and turn it into a functional interface.

Tentative timeline:
* Release kafka-3.3.0 including this commit for Flink 1.19+1.20: early October
* Release kafka-3.4.0 targeting Flink 1.20: mid October
* Release kafka-4.0.0 targeting Flink 2.0(-preview): end of October


fapaul (commented):
Mostly good but please double check the removal of the `serialVersionUID`

fapaul (approved):
Thanks for fixing 👍

PR #122, opened 2024-09-30
[FLINK-35109] Bump to Flink 1.19 and support Flink 1.20

AHeise:
Also
- adds the migration support tests up to 1.20.
- bumps Kafka-client to 3.6.2
- various test cases that become flaky with the bumps


flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableTestUtils.java
```diff
@@ -77,10 +77,17 @@ public static List<Row> collectAllRows(Table table) throws Exception {
...
+        try {
             while (iterator.hasNext()) {
                 collectedRows.add(iterator.next());
             }
+        } finally {
+            try {
+                iterator.close();
+            } catch (Exception e) {
+                // This partially solves FLINK-19204 for Kafka Tests
+            }
```
fapaul, line 90: I know it's only test code, but this pattern always catches an eye since we also catch exceptions like `InterruptedException` and we reset the interrupted flag.

Is this necessary to fix on this PR? I would much rather invest in fixing the cause than fighting the symptoms.

flink-connector-kafka/src/test/java/org/apache/flink/KafkaAssertjConfiguration.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package org.apache.flink;
+
+import org.assertj.core.configuration.Configuration;
+
+/** Configuration for AssertJ that shows full stack traces for unmatched exceptions. */
+public class KafkaAssertjConfiguration extends Configuration {
+    public KafkaAssertjConfiguration() {
+        setMaxStackTraceElementsDisplayed(10000);
```
fapaul, line 26: Since we recently reduced the log sizes significantly, what's the impact of this change?

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableITCase.java
```diff
@@ -1514,4 +1510,22 @@ private static boolean isCausedByJobFinished(Throwable e) {
...
+            e.run();
+        } catch (Exception ex) {
+            // ignore
+        }
+    }
+
+    private static void cancelJob(TableResult tableResult) {
+        if (tableResult != null) {
+            ignoreExceptions(() -> tableResult.getJobClient().ifPresent(JobClient::cancel));
+        }
```
fapaul, line 1530: This only looks like a temporary solution. Can we at least file/link a ticket to fix it properly?

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/testutils/DockerImageVersions.java
```diff
@@ -24,9 +24,9 @@
  */
 public class DockerImageVersions {
 
-    public static final String KAFKA = "confluentinc/cp-kafka:7.4.4";
+    public static final String KAFKA = "confluentinc/cp-kafka:7.7.1";
```
fapaul, line 27: CP 7.7 https://docs.confluent.io/platform/current/release-notes/index.html used Kafka 3.7 while in the pom file we only bump the dependency to 3.6.2.

We should either use CP 7.6 or also increase the used kafka version to 3.7.x to avoid suprises

.github/workflows/weekly.yml
```diff
@@ -30,34 +30,21 @@ jobs:
     strategy:
       matrix:
         flink_branches: [{
-          flink: 1.18-SNAPSHOT,
-          branch: main
-        }, {
           flink: 1.19-SNAPSHOT,
-          jdk: '8, 11, 17, 21',
```
fapaul, line 37: Why is it okay to remove the JDK matrix?

.github/workflows/weekly.yml
```diff
@@ -30,34 +30,21 @@ jobs:
...
-        }, {
-          flink: 1.18.1,
-          branch: v3.1
-        }, {
-          flink: 1.19.0,
-          branch: v3.1,
-          jdk: '8, 11, 17, 21',
+          flink: 1.20.0,
+          branch: v3.2
```
fapaul, line 43: This change feels strange. For an already released connector version `3.2` we are changing the supported Flink version.

We should document this somewhere, why we can do this, and what kind of guarantees we offer.

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/testutils/TypeSerializerMatchers.java
```diff
@@ -0,0 +1,225 @@
...
+
+import java.util.function.Predicate;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * A Collection of useful {@link Matcher}s for {@link TypeSerializer} and {@link
+ * TypeSerializerSchemaCompatibility}.
+ *
+ * <p>Note copied from Flink 1.18. Remove when we drop 1.18 support.
```
fapaul, line 38: I do not understand this from a CI perspective `workflows/weekly.yaml` we removed all testing for 1.18. Why do we need to maintain the tests, then?

fapaul (commented):
The PR feels a bit overloaded. It's great to see so many quality of life improvements but I think they should be separated from the original intent of the PR (probably also been backported).

Regarding the intent of the PR I am not sure I understand it fully. 

- What is the relation between the supported Flink versions and the connector version?
- Do we only support the latest two Flink releases with one connector version (in this case can we remove the 1.18 compatibility)?

flink-connector-kafka/src/test/java/org/apache/flink/KafkaAssertjConfiguration.java
AHeise: If you get a test failure, the log size is larger now. But you actually see the issue. For green builds, everything is unchanged.

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableTestUtils.java
AHeise: The root cause is that closing the iterator is not graceful. But that's an issue in Flink 1.20 that can't be fixed without a new Flink release.
I'll try to pick a more specific exception though to avoid catching IE.

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableITCase.java
AHeise: Why do you think it's temporary? Cleanup code can be invoked at any point in time during test lifecycle. So job may have already failed or not. In the former case, JobClient::cancel throws. In the latter case, it's needed.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/testutils/DockerImageVersions.java
AHeise: The kafka version here is used for server. The pom specifies the client. I wouldn't want to force synchronization as this is rather unrealistic in real life.

.github/workflows/weekly.yml
AHeise: It's defined a few lines below. We previously had different JDK versions for Flink releases, but that's not needed anymore since 1.19 and 1.20 have the same supported JDK versions.

.github/workflows/weekly.yml
AHeise: v3.2 is not a released connector but the release branch in case we need to do bugfixes later.

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/testutils/TypeSerializerMatchers.java
AHeise: Good catch.
I originally started with also supporting 1.18 but I still need to copy the file because 1.20 dropped it.
I will amend the comment to point to 1.19 instead.

AHeise:
>     * What is the relation between the supported Flink versions and the connector version?

There is no relation and that was the intent of factoring it out. We currently encode the supported version in the name of the artifact, e.g.,  `flink-kafka-connector-3.1.0-1.18`. With Flink 2.0 and hopefully binary stable APIs, we would drop the name and just say that it's compatible with all Flink 2 releases (in some kind of version matrix). Hopefully, that works out.

>     * Do we only support the latest two Flink releases with one connector version (in this case can we remove the 1.18 compatibility)?
There is no strict guideline whatsoever. It's important that we maintain bugfixes for critical bugs for all maintained Flink bugfixes. I have outlined my proposal for the next 3 supported versions on the dev ML.

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableTestUtils.java
AHeise: I now specifically pass a list of allowed exceptions.

```
    private void cleanupTopic(String topic) {
        ignoreExceptions(() -> deleteTestTopic(topic), UnknownTopicOrPartitionException.class);
    }

    private static void ignoreExceptions(
            RunnableWithException e, Class<? extends Exception>... exClasses) {
        try {
            e.run();
        } catch (Exception ex) {
            assertThatChainOfCauses(ex)
                    .anyMatch(
                            cause -> Arrays.stream(exClasses).anyMatch(cl -> cl.isInstance(cause)));
        }
    }

    private static void cancelJob(TableResult tableResult) {
        if (tableResult != null) {
            ignoreExceptions(
                    () -> tableResult.getJobClient().ifPresent(JobClient::cancel),
                    FlinkJobNotFoundException.class,
                    FlinkJobTerminatedWithoutCancellationException.class);
        }
    }
```

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/testutils/DockerImageVersions.java
AHeise: Btw 3.7.x had a breaking interface change on the things that we hack to get transactions working. That's why I left it at 3.6.2.

AHeise:
I forked out FLINK-36441 for the test fixes.

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaChangelogTableITCase.java
```diff
@@ -65,7 +64,6 @@ public void testKafkaDebeziumChangelogSource() throws Exception {
         tableConf.set(
                 ExecutionConfigOptions.TABLE_EXEC_MINIBATCH_ALLOW_LATENCY, Duration.ofSeconds(1));
         tableConf.set(ExecutionConfigOptions.TABLE_EXEC_MINIBATCH_SIZE, 5000L);
-        tableConf.set(OptimizerConfigOptions.TABLE_OPTIMIZER_AGG_PHASE_STRATEGY, "TWO_PHASE");
```
fapaul, line 68: What was the intention behind this change?

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducerMigrationTest.java
```diff
@@ -42,7 +42,7 @@
 public class FlinkKafkaProducerMigrationTest extends KafkaMigrationTestBase {
     @Parameterized.Parameters(name = "Migration Savepoint: {0}")
     public static Collection<FlinkVersion> parameters() {
-        return FlinkVersion.rangeOf(FlinkVersion.v1_8, FlinkVersion.v1_16);
```
fapaul, line 45: Why did start of the range change from `1.8` to `1.12`?

fapaul (approved):
LGTM % please look at the inline questions

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaChangelogTableITCase.java
AHeise: The option went from String Flink 1.19 to Enum in Flink 1.20, so you can't compile the same code in both versions.
Talked to Timo what the option does and he said that it shouldn't apply here as we don't have aggregates. And lo-and-behold, nothing changed when I removed the line. (The other options are probably meaningless as well but I don't want to dig into the table tests).

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducerMigrationTest.java
AHeise: Good catch, reverted that part.

PR #126, opened 2024-10-08
[FLINK-36441] Ensure producers are not leaked

AHeise:
Fixes a production and various test issues around producer leaks. The issues appeared when I tried to upgrade Kafka-clients to 3.6.2 but are bugs in our code.

### DIFF
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaWriterITCase.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaWriterITCase.java
@@ -88,31 +88,40 @@ public class KafkaWriterITCase extends KafkaWriterTestBase {
 
         try (final KafkaWriter<Integer> writer =
                 createWriterWithConfiguration(
-                        getKafkaClientConfiguration(), DeliveryGuarantee.NONE, metricGroup)) {
+                        getKafkaClientConfiguration(),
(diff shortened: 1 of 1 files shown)


flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer.java
```diff
@@ -1246,8 +1230,18 @@ private Set<String> generateNewTransactionalIds() {
...
-            Collection<FlinkKafkaProducer.KafkaTransactionState> handledTransactions) {
+    protected void finishProcessing(@Nullable KafkaTransactionState transaction) {
+        super.finishProcessing(transaction);
+        // TwoPhaseCommitSink sets transaction = null on final checkpoint and thus closing will leak
+        // the producer. For transactional producers, we track the producer in pendingTransactions.
+        if (transaction != null && !transaction.isTransactional()) {
+            transaction.producer.flush();
+            transaction.producer.close(Duration.ZERO);
+        }
+    }
```
fapaul, line 1241: I feel uncomfortable merging this change as part of a "fix-tests-ticket". This is a change in production code to more or less support checkpoints-after-task-finish for the deprecated sink function.

fapaul (commented):
I am fine with the test only fixes great work but I am skeptical about the new feature/fix we added to the `FlinkKafkaProducer`.

fapaul (approved):
Thanks for fixing 👍

PR #130, opened 2024-10-14
[FLINK-34466] Lineage interfaces for kafka connector

pawel-big-lebowski:
[FLINK-31275](https://issues.apache.org/jira/browse/FLINK-31275) is aiming to provide native lineage support in Flink's codebase with custom job listeners that get notified about job state changes as well as lineage graph extracted. As a part of that, lineage interfaces introduced in [FLINK-33210](https://issues.apache.org/jira/browse/FLINK-33210) need to be implemented on the connectors' side for sources and sinks to expose lineage metadata about input and output datasets of job runs. 

https://issues.apache.org/jira/browse/FLINK-34466


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/LineageFacetProvider.java
```diff
@@ -0,0 +1,19 @@
...
+
+import org.apache.flink.streaming.api.lineage.LineageDatasetFacet;
+
+import java.util.List;
+
+/**
+ * Contains method which can be used for lineage schema facet extraction. Useful for classes like
+ * topic selectors or serialization schemas to extract dataset information from.
+ */
+public interface LineageFacetProvider {
```
HuangZhenQiu, line 11: I feel this interface can be moved to flink core repo.

HuangZhenQiu:
@pawel-big-lebowski 
Would you please change the PR title with the jira ticket prefix. Please also enhance the PR summary with the original templace

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/LineageUtil.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /**
+     * Creates dataset from a list of facets. Uses {@link KafkaTopicListFacet} to extract dataset
+     * name from. Dataset per each element of topic list is created
+     *
+     * @param facets
+     * @return
+     */
+    public static List<LineageDataset> datasetsFrom(
```
AHeise, line 66: This whole information flow around the facets looks a bit unclean to me.
Both Source/Sink throw a bunch of information into a list of `LineageDatasetFacet`, then this method is applied to take that list apart and construct the actually intended `LineageDataset`. So we first deliberately lose the information of what the facets are about and then we need to use a lot of (hidden) if-else to extract that information again.

WDYT of replacing the `List<LineageDatasetFacet>` instead with a value class that contains all relevant information:
```java
class KafkaFacet { 
  @Nullable
  String topicPattern;
  @Nullable
  List<String> topicList;
  Properties properties;
  @Nullable
  TypeInformation typeInformation;
}
```

Then you can access all the different pieces of information without the isInstance/cast pattern that you use.
You can then in this method still turn all the pieces of information into separate facets.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/facets/KafkaPropertiesFacet.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.flink.streaming.api.lineage.LineageDatasetFacet;
+
+import java.util.Objects;
+import java.util.Properties;
+
+/** Facet containing Kafka properties. */
+public class KafkaPropertiesFacet implements LineageDatasetFacet {
+
+    public static final String KAFKA_PROPERTIES_FACET_NAME = "kafkaProperties";
+    public Properties properties;
```
AHeise, line 29: What assumptions do we make about the mutability and thread-safety of the facades? Do we nned to make defensive copies of the mutable information such as the Properties?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilder.java
```diff
@@ -369,5 +377,23 @@ public ProducerRecord<byte[], byte[]> serialize(
                     value,
                     headerProvider != null ? headerProvider.getHeaders(element) : null);
         }
+
+        @Override
+        public List<LineageDatasetFacet> getDatasetFacets() {
+            List<LineageDatasetFacet> facets = new ArrayList<>();
+            facets.add(new KafkaTopicListFacet(Arrays.asList(topicSelector.apply(null))));
```
AHeise, line 384: Is `topicSelector.apply(null)` guaranteed to work?
Is this even the right thing to do? TopicSelector could return different topics coming from different inputs.
I think we should instead check if TopicSelector is also a LineageProvider and ask it directly.
Our TopicSelectors should then implement it and we should add to the javadoc that LineageProvider is encouraged.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilder.java
```diff
@@ -369,5 +377,23 @@ public ProducerRecord<byte[], byte[]> serialize(
...
+
+            // gets type information from serialize method signature
+            Arrays.stream(this.valueSerializationSchema.getClass().getMethods())
```
AHeise, line 387: Again we should probably check for the serializer to return the TypeInformation directly (by implementing ResultTypeQueryable).
If not we could fallback to extract that as you do, but I'd use things like `org.apache.flink.shaded.guava31.com.google.common.reflect.TypeToken` to be more robust. Your implementation fails if you have some intermediate interface that forward the type parameter.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaSink.java
```diff
@@ -132,4 +140,16 @@ public SimpleVersionedSerializer<KafkaWriterState> getWriterStateSerializer() {
...
         return kafkaProducerConfig;
     }
+
+    @Override
+    public LineageVertex getLineageVertex() {
+        List<LineageDatasetFacet> facets = new ArrayList<>();
+
+        // add all the facets from deserialization schema and subscriber
+        facets.addAll(LineageUtil.facetsFrom(recordSerializer));
+        facets.add(new KafkaPropertiesFacet(this.kafkaProducerConfig));
```
AHeise, line 150: See feedback on LineageUtil

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/LineageFacetProvider.java
pawel-big-lebowski: I think so too. To me, it makes more sense to add it first into flink-core and remove here later when upgrading flink-core for flink-connector-kafka. This shouldn't block lineage interface implementation.

AHeise (commented):
Overall approach looks good. I challenged a central piece of transferring the information via facets to the sink/source, so I haven't checked the tests yet. PTAL.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/LineageFacetProvider.java
```diff
@@ -0,0 +1,19 @@
...
+
+    /**
+     * List of lineage dataset facets.
+     *
+     * @return
+     */
+    List<LineageDatasetFacet> getDatasetFacets();
```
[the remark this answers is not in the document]
pawel-big-lebowski: flink-core lineage interfaces like `LineageVertex` and `LineageGraph`  also have lists, but `Collection` should be enough.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/facets/KafkaPropertiesFacet.java
pawel-big-lebowski: don't know the answers to those questions. I think it's safer to create new properties object.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilder.java
pawel-big-lebowski: Checking for `ResultTypeQueryable` first is fair. I've switched to guava reflection helpers as suggested, but I am not sure if this helps with `implementation fails if you have some intermediate interface that forward the type parameter.`. Could you provide example of this issue? Not sure if this is now covered or not.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilder.java
pawel-big-lebowski: Thanks for pointing this. It works only for the scenario `KafkaRecordSerializationSchema.builder().setTopic(DEFAULT_TOPIC)`. I've changed the implementation to more clear on this case. 

TopicSelector deserves more abstraction than being just `Function<? super IN, String>`, but I don't think this should be part of the scope of this PR.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/facets/KafkaDatasetFacet.java
```diff
@@ -0,0 +1,97 @@
...
+    public static class KafkaDatasetIdentifier {
+        public final List<String> topics;
+        public final Pattern topicPattern;
+
+        public KafkaDatasetIdentifier(List<String> fixedTopics, Pattern topicPattern) {
+            this.topics = fixedTopics;
+            this.topicPattern = topicPattern;
+        }
+
+        public static KafkaDatasetIdentifier of(Pattern pattern) {
```
AHeise, line 71:
```suggestion
        public static KafkaDatasetIdentifier ofPattern(Pattern pattern) {
```

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilder.java
```diff
@@ -369,5 +416,43 @@ public ProducerRecord<byte[], byte[]> serialize(
...
+                typeInformation =
+                        Arrays.stream(this.valueSerializationSchema.getClass().getMethods())
+                                .map(m -> Invokable.from(m))
+                                .filter(m -> "serialize".equalsIgnoreCase(m.getName()))
+                                .map(m -> m.getParameters().get(0))
+                                .filter(p -> !p.getType().equals(TypeToken.of(Object.class)))
+                                .findFirst()
+                                .map(p -> p.getType())
+                                .map(t -> TypeInformation.of(t.getRawType()))
+                                .orElse(null);
```
AHeise, line 450: This looks way more complicated as it should be. Here is what I had in mind.

TypeToken<? extends SerializationSchema> serializationSchemaType = TypeToken.of(valueSerializationSchema.getClass());
Class<?> parameterType = serializationSchemaType.resolveType(SerializationSchema.class.getTypeParameters()[0]).getRawType();
if (parameterType != Object.class) {
    typeInformation = TypeInformation.of(parameterType);
}

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaSink.java
```diff
@@ -132,4 +142,30 @@ public SimpleVersionedSerializer<KafkaWriterState> getWriterStateSerializer() {
...
+    @Override
+    public LineageVertex getLineageVertex() {
+        // enrich dataset facet with properties
+        if (recordSerializer instanceof KafkaDatasetFacetProvider) {
+            Optional<KafkaDatasetFacet> kafkaDatasetFacet =
+                    ((KafkaDatasetFacetProvider) recordSerializer).getKafkaDatasetFacet();
+
+            if (!kafkaDatasetFacet.isPresent()) {
+                LOG.warn("Provided did not return kafka dataset facet");
+                return null;
```
AHeise, line 155: I don't think we are allowed to return null. The interface doesn't specify @Nullable.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaSink.java
```diff
@@ -132,4 +142,30 @@ public SimpleVersionedSerializer<KafkaWriterState> getWriterStateSerializer() {
...
+            }
+
+            kafkaDatasetFacet.get().addProperties(this.kafkaProducerConfig);
```
AHeise, line 158: Do we ever actually get the properties from the recordSerializer? So are we actually just setting here?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilder.java
```diff
@@ -369,5 +416,43 @@ public ProducerRecord<byte[], byte[]> serialize(
...
+                return Optional.empty();
+            }
+
+            TypeInformation typeInformation;
+            if (this.valueSerializationSchema instanceof ResultTypeQueryable) {
+                typeInformation =
+                        ((ResultTypeQueryable<?>) this.valueSerializationSchema).getProducedType();
+            } else {
+                // gets type information from serialize method signature
+                typeInformation =
```
AHeise, line 441: How do we use this type information later? This is the input type, right?

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilderTest.java
```diff
@@ -256,6 +267,102 @@ public void testSerializeRecordWithTimestamp() {
...
+    }
+
+    @Test
+    public void testGetLineageDatasetFacetsWhenNoTopicsIdentifiersFound() {
+        TopicSelector topicSelector =
+                mock(
+                        TopicSelector.class,
+                        withSettings().extraInterfaces(KafkaDatasetIdentifierProvider.class));
+        when(((KafkaDatasetIdentifierProvider) topicSelector).getDatasetIdentifier())
+                .thenReturn(Optional.empty());
```
AHeise, line 290: I haven't looked too closely at the tests. But a high-level comment: In Flink, we don't use mockito (anymore). The main idea is that we use interfaces (as you did) and then just explicitly create our MockImplementation.

```
class MockTopicSelector implements TopicSelector, KafkaDatasetIdentifierProvider {
  KafkaDatasetIdentifier id; // init with ctor or factory method

  KafkaDatasetIdentifier getDatasetIdentifier() { return id; }
}
```

AHeise (commented):
This is definitively heading into the right direction. Structure of production code is good as-is but the handling of the properties doesn't look fully right.

Tests look complete at the first class but we need to get rid of mockito :).

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/LineageUtil.java
```diff
@@ -0,0 +1,105 @@
...
+
+            @Override
+            public String namespace() {
+                return namespace;
+            }
+
+            @Override
+            public Map<String, LineageDatasetFacet> facets() {
+                return Collections.singletonMap(
+                        KafkaDatasetFacet.KAFKA_FACET_NAME, kafkaDatasetFacet);
```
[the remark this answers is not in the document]
pawel-big-lebowski: Duplicates in metadata aint that bad thing. Kafka facet contains topic pattern or topics' list which are necessary to identify datasets and name them properly. On the other hand, the name field is obligatory, which makes sense for other datasets where naming is more straight-forward. Having only the name field wouldn't allow distinction between those scenarios.

A separate `TypeInformationFacet` makes sense as it clearly describes some aspect of a dataset. On the other hand, I feel Kafka properties should be a part of `KafkaDatasetFacet` bcz it's Kafka specific and bcz the properties allow resolving topics' pattern, so it's worth to have them both in a single facet.

What would make most sense to me now, is having two facets: `TypeInformationFacet` for type information and `KafkaDatasetFacet` to contain: Kafka properties, topics list and topic pattern in case it's present. Does it make sense to you as well?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/LineageUtil.java
```diff
@@ -0,0 +1,105 @@
...
+
+    public static LineageDataset datasetOf(String namespace, KafkaDatasetFacet kafkaDatasetFacet) {
+        return new LineageDataset() {
+            @Override
+            public String name() {
+                if (kafkaDatasetFacet.topicIdentifier.topicPattern != null) {
+                    return kafkaDatasetFacet.topicIdentifier.toString();
+                }
+
+                return String.join(",", kafkaDatasetFacet.topicIdentifier.topics);
```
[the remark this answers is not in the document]
pawel-big-lebowski: It's important to differentiate it on the job listener's code. Within this code, I want to expand topic pattern into its current representation of the topics. So, the code outside the connector should know if identifier is topic pattern or topic list.

A job listener's code is part of OpenLineage repo, currently within the PR: https://github.com/OpenLineage/OpenLineage/pull/3099. It's used within `KafkaTopicPatternDatasetIdentifierVisitor` class.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/KafkaDatasetIdentifierProvider.java
```diff
@@ -0,0 +1,16 @@
+package org.apache.flink.connector.kafka.lineage;
+
+import org.apache.flink.connector.kafka.lineage.facets.KafkaDatasetFacet.KafkaDatasetIdentifier;
+
+import java.util.Optional;
+
+/** Contains method which allows extracting topic identifier. */
+public interface KafkaDatasetIdentifierProvider {
```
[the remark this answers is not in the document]
pawel-big-lebowski: Lineage classes in flink-core do contain separate interfaces like `LineageVertexProvider` in a separate file. Couldn't find any code with `Provider` interface within other interfaces.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/LineageUtil.java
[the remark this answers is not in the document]
pawel-big-lebowski: I am OK with more OOP approach with interfaces containing getters and default classes with the properties.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaSink.java
pawel-big-lebowski: Nope, interface doesn't specify `@nullable`. We can return `LineageVertex` with empty dataset list instead.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaSink.java
pawel-big-lebowski: ok, we can convert it to `setter`.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilder.java
pawel-big-lebowski: This is returned within the facet and then listener (like OpenLineageJobListener) converts it to dataset schema format description. For OpenLineage, it's called `SchemaDatasetFacet`. I think this is not Kafka connector specific and there should be a general schema-alike facet within flink core. However, I don't feel I would be able to achieve this now. Schema information is valuable for both input and output datasets.

I hope typeInformation approach will work well for `Avro` and `Protobuf`. Hopefully, in some time, I create separate tests within OpenLineage job listener to verify this.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/KafkaDatasetIdentifierProvider.java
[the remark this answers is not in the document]
AHeise: I double-checked and it's not that common of a pattern. It's used only in

```
OperatorCoordinator.java
FileEnumerator.java
DynamicFileEnumerator.java
FileSplitAssigner.java
InternalTimeServiceManager.java
```

So you can just stick to the current approach.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/LineageUtil.java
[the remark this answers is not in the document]
AHeise: Yes it would work well. Tbh I'm not sure if 1 or 2 is the better approach. Maybe @HuangZhenQiu can weigh in.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilder.java
AHeise: Yes TypeInformationFacet sounds like a general concept. I'm convinced you want to pull it out of the KafkaFacet now. You probably want to name it "inputType" and "outputType" depending on the type of the connector (source/sink). I'd design it generally and pull it up into flink-core for Flink 2.0 later (so make it work in Kafka first and then propose to port it upwards).

pawel-big-lebowski: `LineageGraph` in the flink-core contains separate lists of sources and sinks. Given that, I am not sure if we want to distinguish "inputType" from "outputType". From the facet perspective, this should be all `type` and the same facet can be used for both scenarios.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/lineage/LineageUtilTest.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Collections;
+import java.util.Properties;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** Tests for {@link LineageUtil}. */
+public class LineageUtilTest {
+    @Test
+    public void testSourceLineageVertexOf() {
+        LineageDataset dataset = Mockito.mock(LineageDataset.class);
```
HuangZhenQiu, line 19: As called out by @AHeise, we need to move out from Mockito with testing classes. I am thinking. I should probably add these helper test classes in flink-core rather than implement in each of connector.

pawel-big-lebowski: Thanks @HuangZhenQiu for noticing that place.

HuangZhenQiu (approved):
Thanks for the contribution. This diff give a great example for connectors to support flink native lineage.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/LineageUtil.java
```diff
@@ -0,0 +1,118 @@
...
+
+        if (bootstrapServers == null) {
+            return KAFKA_DATASET_PREFIX;
+        }
+
+        if (bootstrapServers.contains(COMMA)) {
+            bootstrapServers = bootstrapServers.split(COMMA)[0];
+        } else if (bootstrapServers.contains(SEMICOLON)) {
+            bootstrapServers = bootstrapServers.split(SEMICOLON)[0];
+        }
```
AHeise, line 91: Can you check if there is already some util in kafka that does that? If not, leave as is.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/LineageUtil.java
```diff
@@ -0,0 +1,118 @@
...
+    public static SourceLineageVertex sourceLineageVertexOf(Collection<LineageDataset> datasets) {
+        return new SourceLineageVertex() {
+            @Override
+            public Boundedness boundedness() {
+                return Boundedness.CONTINUOUS_UNBOUNDED;
+            }
+
+            @Override
+            public List<LineageDataset> datasets() {
+                return datasets.stream().collect(Collectors.toList());
```
AHeise, line 105:
```suggestion
                return List.copyOf(datasets);
```

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/TypeDatasetFacetProvider.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Optional;
+
+/** Contains method to extract {@link TypeDatasetFacet}. */
+public interface TypeDatasetFacetProvider {
+
+    /**
+     * Returns a type dataset facet or `Optional.empty` in case an implementing class is not able to
+     * resolve type.
+     *
+     * @return
```
AHeise, line 12: Please remove all empty javadoc tags or let Copilot help you ;)

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/KafkaDatasetIdentifierProvider.java
```diff
@@ -0,0 +1,15 @@
+package org.apache.flink.connector.kafka.lineage;
+
+import java.util.Optional;
+
+/** Contains method which allows extracting topic identifier. */
+public interface KafkaDatasetIdentifierProvider {
```
AHeise, line 6: Make sure to tag all public API with `@PublicEvolving`. It needs to be clearly visible if a user is supposed to touch the class or not (the easiest way is to not use public unless needed).

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/subscriber/PartitionSetSubscriber.java
```diff
@@ -26,13 +29,14 @@
...
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
 import static org.apache.flink.connector.kafka.source.enumerator.subscriber.KafkaSubscriberUtils.getTopicMetadata;
 
 /** A subscriber for a partition set. */
-class PartitionSetSubscriber implements KafkaSubscriber {
+class PartitionSetSubscriber implements KafkaDatasetIdentifierProvider, KafkaSubscriber {
```
AHeise, line 39:
```suggestion
class PartitionSetSubscriber implements KafkaSubscriber, KafkaDatasetIdentifierProvider {
```

keep it consistent

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilderTest.java
```diff
@@ -256,6 +271,132 @@ public void testSerializeRecordWithTimestamp() {
...
+    public void testGetLineageDatasetFacetsWhenTopicSelectorNotKafkaTopicsIdentifierProvider() {
+        SerializationSchema<String> serializationSchema = new SimpleStringSchema();
+        KafkaRecordSerializationSchema<String> schema =
+                KafkaRecordSerializationSchema.builder()
+                        .setTopicSelector((TopicSelector<Object>) o -> DEFAULT_TOPIC)
+                        .setValueSerializationSchema(serializationSchema)
+                        .setKeySerializationSchema(serializationSchema)
+                        .build();
+
+        assertThat(((KafkaDatasetFacetProvider) schema).getKafkaDatasetFacet()).isEmpty();
```
AHeise, line 284: A bit more assertj-ish would be

```
        assertThat(schema)
                .asInstanceOf(InstanceOfAssertFactories.type(KafkaDatasetFacetProvider.class))
                .returns(List.of(), KafkaDatasetFacetProvider::getKafkaDatasetFacet);
```

That would result in an assertion error instead of runtime error if the Schema does not implement the interface.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilder.java
```diff
@@ -79,6 +95,7 @@
  */
 @PublicEvolving
 public class KafkaRecordSerializationSchemaBuilder<IN> {
+    private static final Logger LOG = LoggerFactory.getLogger(KafkaSource.class);
```
AHeise, line 98: Not the correct place, but please update the docs of the KafkaRecordSerializationSchema to point to the FacetProvider interface. Same to all other APIs where you hope that optional interfaces are implemented.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceTest.java
```diff
@@ -0,0 +1,165 @@
...
+                                return null;
+                            }
+                        },
+                        null,
+                        null,
+                        Boundedness.CONTINUOUS_UNBOUNDED,
+                        null,
+                        kafkaProperties,
+                        null);
+        assertThat(source.getLineageVertex().datasets()).isEmpty();
```
AHeise, line 75: Can you use the builder instead? That should also be less verbose.

AHeise (commented):
LAGTM. A few more nits. The most important part is around documentation. Make sure all Public elements are properly annotated and that you link from existing interfaces to the new optional mixins.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/LineageUtil.java
pawel-big-lebowski: Seems like piece of code that has to be available somewhere, but I wasn't able to find it.

AHeise (approved):
LGTM. I triggered the hopefully final CI run.

AHeise:
Thank you very much for your contribution (and patience).

pawel-big-lebowski:
Awesome. @AHeise Thank you for your feedback and cooperation on that.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/LineageUtil.java
```diff
@@ -0,0 +1,118 @@
...
+            }
+        };
+    }
+
+    public static LineageVertex lineageVertexOf(Collection<LineageDataset> datasets) {
+        return new LineageVertex() {
+            @Override
+            public List<LineageDataset> datasets() {
+                return datasets.stream().collect(Collectors.toList());
```
davidradl, line 114: Can we not use the `List.copyOf(datasets);` as per Arvid's suggested change here also.

pawel-big-lebowski: Merged version is not using `List.copyOf`. We agreed on that in offline discussion.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilder.java
```diff
@@ -369,5 +420,46 @@ public ProducerRecord<byte[], byte[]> serialize(
...
+                return Optional.of(
+                        new DefaultTypeDatasetFacet(
+                                ((ResultTypeQueryable<?>) this.valueSerializationSchema)
+                                        .getProducedType()));
+            } else {
+                // gets type information from serialize method signature
+                TypeToken serializationSchemaType =
+                        TypeToken.of(valueSerializationSchema.getClass());
+                Class parameterType =
+                        serializationSchemaType
```
davidradl, line 454: Is there a way to avoid using this reflection - instanceof and Class. Maybe using a config driven approach and java SPI. The connectors / formats bring in serialization implementations in this way that avoid the overhead of reflection.

pawel-big-lebowski: You can avoid this with implementing `ResultTypeQueryable` of the value serialisation schema.

PR #132, opened 2024-10-23
[FLINK-36595][docs] Explicitly set connector compatibility as string to prevent version comparison mismatch

z3d1k:
Fix connector release flink compatibility configuration for documentation build to render released artifact version correctly.

#### Before fix:
image

#### After fix:
image


dannycranmer (approved):
LGTM, thanks

PR #133, opened 2024-11-08
[FLINK-36630][Connectors/Kafka] Wrap consumer.position in retryOnWakeup

trompa:
FLINK-34470 introduced a new consumer.position call that was not wrapped by retryOnWakeUp, differently than the rest of calls ont he class.

It was generating random WakeupExceptions on our production code for Flink19.

Trivial code change. But I extracted this to a method so i could add a simple test for all the calls

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
@@ -279,6 +279,10 @@ public class KafkaPartitionSplitReader
         }
     }
 
+    long getConsumerPosition(TopicPartition tp, String msg) {
+        return retryOnWakeup(() -> consumer.position(tp), msg);
+    }
+
     private void parseStartingOffsets(
             KafkaPartitionSplit split,
             List<TopicPartition> partitionsStartingFromEarliest,
@@ -442,11 +446,6 @@ public class KafkaPartitionSplitReader
 
     private long getStoppingOffset(TopicPartition tp) {
         return stoppingOffsets.getOrDefault(tp, Long.MAX_VALUE);
-    }
-
-    @VisibleForTesting
-    protected long getConsumerPosition(TopicPartition tp, String msg) {
-        return retryOnWakeup(() -> consumer.position(tp), msg);
     }
 
     private void maybeRegisterKafkaConsumerMetrics(
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
@@ -124,7 +124,7 @@ public class KafkaPartitionSplitReader
         List<TopicPartition> finishedPartitions = new ArrayList<>();
         for (TopicPartition tp : consumer.assignment()) {
             long stoppingOffset = getStoppingOffset(tp);
-            long consumerPosition = consumer.position(tp);
+            long consumerPosition = getConsumerPosition(tp, "retrieving consumer position");
             // Stop fetching when the consumer's position reaches the stoppingOffset.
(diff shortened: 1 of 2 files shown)


trompa:
Friendly ping to ask for a review.
It does not introduce any new code, it just uses the retryOnWakeUp on a consumer.position call introduced by FLINK-34470 as it was done on the other calls. minor refactor to reuse code, too.

trompa:
@PatrickRen @AHeise @leonardBang
sorry to bother, but this small change is preventing us to raise from 1.18
PTAL

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -279,6 +279,10 @@ void setConsumerClientRack(Properties consumerProps, String rackId) {
         }
     }
 
+    long getConsumerPosition(TopicPartition tp, String msg) {
```
fapaul, line 282:
```suggestion
    @VisibleForTesting
    long getConsumerPosition(TopicPartition tp, String msg) {
```

fapaul (commented):
The change itself looks good but I am not convinced that the test ensures the behavior. 

Theoretically the existing wakeUp test can already cover that scenario (https://github.com/apache/flink-connector-kafka/blob/7c112abe8bf78e0cd8a310aaa65b57f6a70ad30a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReaderTest.java#L110) although very unreliable.

I'd suggest to follow a similar approach where we call `reader.wakeUp` as long as the thread is running to increase the likelihood of running into the wakeup during position fetch.

trompa:
TBH, I think the code is just about adding a missing wrapper on the position call to make it consistent with the rest of the usages on the file, but didnt liked to create a PR without tests at all, so i created this one. 

I never liked it much, But I was not able to deterministically make it throw within a thread.
Any help is welcome.

fapaul:
> I never liked it much, But I was not able to deterministically make it throw within a thread.
Any help is welcome.

I agree with that the current code structure makes it hard to test. My suggestion is to change your existing test slightly to use a similar pattern to the existing test: https://github.com/apache/flink-connector-kafka/blob/7c112abe8bf78e0cd8a310aaa65b57f6a70ad30a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReaderTest.java#L110.

The existing test retries calling `reader.wakeUp` as long as the thread runs, increasing the likelihood of hitting the case. Your test, on the other hand, only calls it once and then waits for the thread to finish.

trompa:
Added the wakeUp loop on my tests:

```
while (t.isAlive() && System.currentTimeMillis() < deadline) {
            reader.wakeUp();
            Thread.sleep(10);
        }
```

with short periods, it happen to throw, reason is because on first wakeUp it retries the consumer.position() call and a new wakeup is triggered that is not wrpped.

would it make sense to wrap that second call?

    private <V> V retryOnWakeup(Supplier<V> consumerCall, String description) {
        try {
            return consumerCall.get();
        } catch (WakeupException we) {
            LOG.info(
                    "Caught WakeupException while executing Kafka consumer call for {}. Will retry the consumer call.",
                    description);
    -       return consumerCall.get();
    +       return retryOnWakeup(consumerCall, description);

        }
    }

fapaul:
Good catch. This also shows that the existing test with the retry loop is more or less flaky due to your discovered issue. With all that in mind, I am tending to remove all the test changes from this PR and just merge the production code change as is. 

Solving the testing issue needs a bigger refactoring that shouldn't block this fix.

WDYT?

trompa:
fine for me

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
trompa: no need anymore as test is being removed

trompa:
done!

trompa:
Thank you!

PR #136, opened 2024-11-23
[FLINK-36780] Kafka source disables partition discovery unexpectedly

liuml07:
Currently Kafka source enables partition discovery. This is set by partition.discovery.interval.ms, aka KafkaSourceOptions.PARTITION_DISCOVERY_INTERVAL_MS. The default value is 5 minutes, which is equal to the default value of metadata.max.age.ms in Kafka.

However, it's disabled by default unexpectedly in the source builder ([code](https://github.com/apache/flink-connector-kafka/blob/main/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSourceBuilder.java#L476-L480)). The intention I believe was to only disable for bounded source.

We need a fix that is able to keep the default partition discovery. This could cause data loss after Kafka retention if the new partitions are not consumed silently.

### DIFF
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceBuilderTest.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceBuilderTest.java
@@ -217,6 +217,29 @@ public class KafkaSourceBuilderTest {
                 .hasMessageContaining(expectedError);
     }
 
+    @Test
+    public void testDefaultPartitionDiscovery() {
+        final KafkaSource<String> kafkaSource = getBasicBuilder().build();
+        // Commit on checkpoint and auto commit should be disabled because group.id is not specified
+        assertThat(
+                        kafkaSource
+                                .getConfiguration()
+                                .get(KafkaSourceOptions.PARTITION_DISCOVERY_INTERVAL_MS))
+                .isEqualTo(KafkaSourceOptions.PARTITION_DISCOVERY_INTERVAL_MS.defaultValue());
+    }
+
+    @Test
+    public void testPeriodPartitionDiscovery() {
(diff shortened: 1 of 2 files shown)


flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceBuilderTest.java
```diff
@@ -217,6 +217,29 @@ public void testSettingInvalidCustomDeserializers(
                 .hasMessageContaining(expectedError);
     }
 
+    @Test
+    public void testDefaultPartitionDiscovery() {
```
liuml07, line 221: This test fails without the fix, and passes now with the fix.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceBuilderTest.java
```diff
@@ -217,6 +217,29 @@ public void testSettingInvalidCustomDeserializers(
...
+        // Commit on checkpoint and auto commit should be disabled because group.id is not specified
+        assertThat(
+                        kafkaSource
+                                .getConfiguration()
+                                .get(KafkaSourceOptions.PARTITION_DISCOVERY_INTERVAL_MS))
+                .isEqualTo(KafkaSourceOptions.PARTITION_DISCOVERY_INTERVAL_MS.defaultValue());
+    }
+
+    @Test
+    public void testPeriodPartitionDiscovery() {
```
liuml07, line 232: This newly added test is to avoid regressions for bounded source.

pnowojski (commented):
I think the change looks good to me. It would be good to have some feedback from someone more familiar with the Kafka Source. If that doesn't happen within a couple of days, I would propose to just go ahead and merge it.

PR #137, opened 2024-11-28
FLINK-36817: Introduce KafkaConsumerFactory

lkokhreidze:
In certain scenarios users of the `KafkaSource` in the flink-connector-kafka might want to provide their own KafkaConsumer. Right now this is not possible as consumer is created in the [KafkaPartitionSplitReader](https://github.com/apache/flink-connector-kafka/blob/main/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java#L97) which makes customisation impossible.

Proposal is to let users pass `KafkaConsumerFactory` when building the KafkaSource.

```
public interface KafkaConsumerFactory {
  KafkaConsumer<byte[], byte[]> get(Properties properties);
}
```

Builder will have a default implementation which creates the KafkaConsumer the same as it is done now.

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaConsumerFactory.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaConsumerFactory.java
@@ -1,0 +1,9 @@
+package org.apache.flink.connector.kafka.source;
+
+import org.apache.kafka.clients.consumer.KafkaConsumer;
+
+import java.util.Properties;
+
+public interface KafkaConsumerFactory {
+    KafkaConsumer<byte[], byte[]> get(Properties properties);
+}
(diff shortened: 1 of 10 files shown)


lkokhreidze:
Will add the javadocs. But first wanted to get the initial feedback if this makes sense.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaConsumerFactory.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+package org.apache.flink.connector.kafka.source;
+
+import org.apache.kafka.clients.consumer.KafkaConsumer;
+
+import java.util.Properties;
+
+public interface KafkaConsumerFactory {
+    KafkaConsumer<byte[], byte[]> get(Properties properties);
```
lkokhreidze, line 25: I was debating myself if it would be better to return `Consumer` interface instead of the implementation. Happy to change it if community agrees that returning an interface is the better choice.

PR #139, opened 2024-12-03
[FLINK-36832] Remove Deprecated classes and relevant tests.

lvyanquan:
Remove Deprecated classes and relevant tests, and annotated some classes from Deprecated to Internal.


lvyanquan:
Hi, @PatrickRen  @AHeise. Could you please help to review this?

AHeise:
Hi @lvyanquan , thank you very much for tackling the release. Unfortunately, I don't have time to review it this year and per prior agreement, I expected @PatrickRen and @leonardBang to pick this up. I will ping them again on slack.

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaITCase.java
```diff
@@ -1,380 +0,0 @@
...
-
-import javax.annotation.Nullable;
-
-import java.io.ByteArrayInputStream;
-import java.io.IOException;
-import java.util.Optional;
-import java.util.UUID;
-
-/** IT cases for Kafka. */
-public class KafkaITCase extends KafkaConsumerTestBase {
```
leonardBang, line 52: Do we have new test/ITCase which bases on new API that covers these cases ?

leonardBang (commented):
Thanks @lvyanquan for the PR, I left one comment

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaITCase.java
lvyanquan: We have independent test for new Source/SInk API in [KafkaSinkITCase](https://github.com/apache/flink-connector-kafka/blob/f6a077a9dd8d1d5e43fc545cc9baab227d8438a0/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaSinkITCase.java#L108) and [KafkaSourceITCase](https://github.com/apache/flink-connector-kafka/blob/f6a077a9dd8d1d5e43fc545cc9baab227d8438a0/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaSourceITCase.java), but I don't see they are equivalent to the previous testing scenarios, these new tests are designed separately.

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaShortRetentionTestBase.java
```diff
@@ -128,67 +118,6 @@ public static void shutDownServices() throws Exception {
      */
     private static boolean stopProducer = false;
 
-    public void runAutoOffsetResetTest() throws Exception {
```
AHeise, line 131: Remove entire class?

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/internals/KafkaTopicPartitionTest.java
```diff
@@ -26,14 +27,14 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.fail;
 
-/** Tests for the {@link KafkaTopicPartition}. */
+/** Tests for the {@link TopicPartition}. */
 public class KafkaTopicPartitionTest {
 
     @Test
     public void validateUid() {
```
AHeise, line 34: You can remove this whole class. With Flink 2.0, we will not be able to restore state from old consumers anyways.

We should document (and I think this applies to Flink 2.0 more than to Kafka 4.0) that users need to migrate from SourceFunction/SinkFunction to Source/Sink with Flink 1.20 before switching to Flink 2.0.

AHeise: Note that the test at it's current state is absolutely meaningless.

AHeise (commented):
LGTM % 3 smaller things. We also should add state migration test to Source/Sink from 1.20 to 2.0 before releasing kafka-4.0. I'd create a follow-up task. In the worst case, we need to re-add some resources (but I don't foresee it). This PR, however, is big enough on its on, so I don't want to increase scope.

AHeise:
I added https://issues.apache.org/jira/browse/FLINK-37330 as a follow-up task and marked it a blocker.

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/internals/KafkaTopicPartitionTest.java
lvyanquan: Hi, @AHeise.

At the beginning, for the convenience of review, I split the entire work of bump 2.0 into two PRs: [deleting deprecated APIs](https://github.com/apache/flink-connector-kafka/pull/139) and [adapting 2.0 APIs](https://github.com/apache/flink-connector-kafka/pull/140). 

However, this seems to have resulted in duplicate review work. I have already deleted the classes mentioned here in the [adapting 2.0 API](https://github.com/apache/flink-connector-kafka/pull/140). 

I plan to close this PR and let's trace it in a unified PR.

PR #142, opened 2024-12-16
[FLINK-36909] Reduce Stream Creation in StreamPatternSubscriber

xiaochen-zhou:
In StreamPatternSubscriber mode, our current approach is to first fetch metadata for all streams, then create all streams, and subsequently apply filters. We can optimize this process by initially filtering stream identifiers, then fetching metadata and creating streams only for the filtered ones, thereby enhancing performance and reducing resource consumption.


xiaochen-zhou:

PR #143, opened 2024-12-23
fix DynamicKafkaRecordSerializationSchema, add serialVersionUID

lwjhn:


lwjhn:
Caused by: java.io.InvalidClassException: org.apache.flink.streaming.connectors.kafka.table.DynamicKafkaRecordSerializationSchema; local class incompatible: stream classdesc serialVersionUID = 7922944055764611744, local class serialVersionUID = -3355247194660797383

PR #153, opened 2025-02-17
[FLINK-37304] Reduce the log level about metadataUpdateEvent in DynamicKafkaSourceEnumerator.

lvyanquan:
To avoid displaying a large amount of connection information.


leonardBang (approved):
Thanks @lvyanquan for the contribution, LGTM

PR #154, opened 2025-02-19
[FLINK-34554] Introduce transaction strategies

AHeise:
Make naming and abortion of transaction configurable by strategies.

Add list abortion strategy with uses ListTransaction API.

Add pooling name strategy to reduce load onto broker.


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
```diff
@@ -53,7 +56,14 @@
  */
 class ExactlyOnceKafkaWriter<IN> extends KafkaWriter<IN> {
     private static final Logger LOG = LoggerFactory.getLogger(ExactlyOnceKafkaWriter.class);
+    /**
+     * Prefix for the transactional id. Must be unique across all sinks writing to the same broker.
+     */
     private final String transactionalIdPrefix;
+    /** Strategy to abort lingering transactions from previous executions. */
```
fapaul, line 71: Nit: `during writer initialization.`

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
```diff
@@ -136,13 +158,10 @@ public void initialize() {
...
+        namingContext.setNextCheckpointId(checkpointId);
         FlinkKafkaInternalProducer<byte[], byte[]> producer =
-                producerPool.getTransactionalProducer(
-                        TransactionalIdFactory.buildTransactionalId(
-                                transactionalIdPrefix,
-                                kafkaSinkContext.getParallelInstanceId(),
-                                checkpointId),
-                        checkpointId);
+                transactionNamingStrategy.getTransactionalProducer(namingContext);
+        namingContext.setLastCheckpointId(checkpointId);
```
fapaul: This looks strange. Why do we expose `setNextCheckpointId` and `setLastCheckpointId` if both methods are always used together? Is the interleaving of the producer creation necessary?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
```diff
@@ -224,13 +243,32 @@ private void abortLingeringTransactions(
...
+        transactionAbortStrategy.abortTransactions(context);
+    }
+
+    private TransactionAbortStrategyContextImpl getTransactionAbortStrategyContext(
+            long startCheckpointId, List<String> prefixesToAbort) {
+        TransactionAbortStrategyImpl.TransactionAborter aborter =
+                transactionalId -> {
+                    FlinkKafkaInternalProducer<byte[], byte[]> producer =
+                            producerPool.getTransactionalProducer(transactionalId, 0);
+                    LOG.debug("Aborting transaction {}", transactionalId);
```
fapaul, line 283: Shouldn't we do any operation on the transaction here e.g. abort/init? Can you add a comment what is happening?

fapaul (commented):
I reviewed the first commit and I am not sure about splitting the naming strategy and the probing. IMO both feel closely related and I do not see the need to expose them separately.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/TransactionAbortStrategy.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.flink.connector.kafka.sink.internal.TransactionAbortStrategyImpl;
+
+/**
+ * The strategy to abort a transactions left by ungraceful job termination, which is the usual case
+ * for failures in Flink. On start of the sink, the strategy will be used to abort any ongoing
+ * transactions left open while cautiously avoid to abort transactions that are still trying to
+ * commit.
+ */
+@PublicEvolving
+public enum TransactionAbortStrategy {
```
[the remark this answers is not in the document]
AHeise: You can use the list transaction abort strategy with the incrementing naming strategy if you don't experience pressure on the memory of the broker. There are some advantages in having unique transaction ids per checkpoints (such as ease of debugging).

A major point why I separated it is because both the default abortion and the default naming strategies are already battle-tested. Folks will be reluctant to set the new strategies on prod and have the option to migrate in separate steps.

I'm also assuming that we will have more strategies in the future with more possible combinations.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
AHeise: Not sure how to improve it. There could be an arbitrary number of aborted transactions between the two invocations of the call-site. For me, these are two disjunct states that are coupled in the best case but not in the general case. Remember that we need to `setLastCheckpointId` once already during recovery.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
AHeise: I'll add a comment: "getTransactionalProducer already calls initTransactions, which cancels the transaction".

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionNamingStrategyContextImpl.java
```diff
@@ -0,0 +1,78 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.connector.kafka.sink.internal;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/** Implementation of {@link TransactionNamingStrategyImpl.Context}. */
+public class TransactionNamingStrategyContextImpl implements TransactionNamingStrategyImpl.Context {
```
fapaul, line 26: Internal?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionNamingStrategyImpl.java
```diff
@@ -0,0 +1,77 @@
...
+                    expectedCheckpointId,
+                    lastCheckpointId);
+            // in case checkpoints have been aborted, Flink would create non-consecutive transaction
+            // ids
+            // this loop ensures that all gaps are filled with initialized (empty) transactions
+            for (long checkpointId = lastCheckpointId + 1;
+                    checkpointId < expectedCheckpointId;
+                    checkpointId++) {
+                context.recycle(context.getProducer(context.buildTransactionalId(checkpointId)));
+            }
```
fapaul, line 55: Is this newly added logic or was this previously part of the recovery?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionNamingStrategyImpl.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+package org.apache.flink.connector.kafka.sink.internal;
+
+import org.apache.flink.connector.kafka.sink.KafkaSinkOptions.TransactionNamingStrategy;
+
+import static org.apache.flink.util.Preconditions.checkState;
+
+/** Implementation of {@link TransactionNamingStrategy}. */
+public enum TransactionNamingStrategyImpl {
```
fapaul, line 28: Something looks off why is `PROBING` not defined in this class.

fapaul (commented):
More comments

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
```diff
@@ -264,11 +282,51 @@ private TransactionAbortStrategyContextImpl getTransactionAbortStrategyContext(
...
+                getDatasetIdentifier()
+                        .orElseThrow(
+                                () ->
+                                        new IllegalStateException(
+                                                "The record serializer does not expose a static list of target topics."));
+        if (identifier.getTopics() != null) {
+            return identifier.getTopics();
+        }
+        return AdminUtils.getTopicsByPattern(getAdminClient(), identifier.getTopicPattern());
+    }
```
fapaul, line 318: Does the list transaction mechanism only work when using the lineage feature? In general, how do we determine the target topic if the topic is only decided by the user provided serializer?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriterStateSerializer.java
```diff
@@ -47,8 +54,15 @@ public byte[] serialize(KafkaWriterState state) throws IOException {
     public KafkaWriterState deserialize(int version, byte[] serialized) throws IOException {
         try (final ByteArrayInputStream bais = new ByteArrayInputStream(serialized);
                 final DataInputStream in = new DataInputStream(bais)) {
-            final String transactionalIdPrefx = in.readUTF();
-            return new KafkaWriterState(transactionalIdPrefx);
+            final String transactionalIdPrefix = in.readUTF();
+            final Collection<CheckpointTransaction> ongoingTransactions = new ArrayList<>();
+            if (version >= 2) {
```
fapaul, line 59: Nit: Isn't it better to assert on version 2 and fail for other versions?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java
```diff
@@ -107,6 +119,64 @@ private int abortTransactionOfSubtask(String prefix, int subtaskId, Context cont
...
+
+        @Override
+        public void abortTransactions(Context context) {
+
+            Collection<TransactionListing> openTransactionsForTopics =
+                    AdminUtils.getOpenTransactionsForTopics(
+                            context.getAdminClient(), context.getTopicNames());
+
+            // This list could be huge if TransactionNamingStrategy#INCREMENTING is used, so cap it
+            LOG.trace(
```
fapaul, line 139: Let's warn the user how many transactions will be aborted to spot unwanted behavior earlier

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaSinkOptions.java
```diff
@@ -72,7 +72,15 @@ public enum TransactionNamingStrategy {
...
          */
         INCREMENTING(
-                TransactionNamingStrategyImpl.INCREMENTING, TransactionAbortStrategyImpl.PROBING);
+                TransactionNamingStrategyImpl.INCREMENTING, TransactionAbortStrategyImpl.PROBING),
+        /**
+         * This strategy reuses transaction names. It is more resource-friendly than {@link
+         * #INCREMENTING} on the Kafka broker.
+         *
+         * <p>It's a new strategy introduced in flink-connector-kafka 4.X. It requires Kafka 3.0+
+         * and additional permissions.
```
fapaul, line 81: Can we list the additional operations e.g. calls to the admin client.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionNamingStrategyImpl.java
```diff
@@ -53,6 +55,20 @@ public FlinkKafkaInternalProducer<byte[], byte[]> getTransactionalProducer(
...
+        @Override
+        public FlinkKafkaInternalProducer<byte[], byte[]> getTransactionalProducer(
+                Context context) {
+            Collection<String> usedTransactionalIds = context.getOngoingTransactions();
+            for (int offset = 0; ; offset++) {
+                String transactionalIdCandidate = context.buildTransactionalId(offset);
+                if (usedTransactionalIds.contains(transactionalIdCandidate)) {
+                    continue;
+                }
+                return context.getProducer(transactionalIdCandidate);
```
fapaul, line 69: Do we have a log line somewhere that indicates how big the pool is? Since it might be an indicator of an serious issues we should probably log on every new producer creation.

flink-connector-kafka/src/test/resources/log4j2-test.properties
```diff
@@ -18,7 +18,7 @@
 
 # Set root logger level to OFF to not flood build logs
 # set manually to INFO for debugging purposes
-rootLogger.level = OFF
+rootLogger.level = WARN
```
fapaul, line 21: Please revert the logging changes before merging.

fapaul (commented):
Reviewed the PR now e2e. Regarding the new list transaction logic it's a bit unclear to me how we want to discover the topic names.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java
pnowojski: +1 this should be `WARN`

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java
```diff
@@ -0,0 +1,134 @@
...
+
+package org.apache.flink.connector.kafka.sink.internal;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+
+/** Implementations of an abort strategy for transactions left over from previous runs. */
+public enum TransactionAbortStrategyImpl {
```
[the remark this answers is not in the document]
AHeise: Yes. (it's already part of the internal package but it's better to be safe than sorry).

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionNamingStrategyImpl.java
AHeise: This is the previous logic expressed through the context.

```
    private FlinkKafkaInternalProducer<byte[], byte[]> getTransactionalProducer(long checkpointId) {
        checkState(
                checkpointId > lastCheckpointId,
                "Expected %s > %s",
                checkpointId,
                lastCheckpointId);
        FlinkKafkaInternalProducer<byte[], byte[]> producer = null;
        // in case checkpoints have been aborted, Flink would create non-consecutive transaction ids
        // this loop ensures that all gaps are filled with initialized (empty) transactions
        for (long id = lastCheckpointId + 1; id <= checkpointId; id++) {
            String transactionalId =
                    TransactionalIdFactory.buildTransactionalId(
                            transactionalIdPrefix, kafkaSinkContext.getParallelInstanceId(), id);
            producer = getOrCreateTransactionalProducer(transactionalId);
        }
        this.lastCheckpointId = checkpointId;
        assert producer != null;
        LOG.info("Created new transactional producer {}", producer.getTransactionalId());
        return producer;
    }
```

Now that I'm looking at the old code, this probably leaked producers en mass...

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
AHeise: In that case, the user has implement the interfaces. 

```
            if (transactionNamingStrategy.requiresKnownTopics()) {
                checkState(
                        recordSerializer instanceof KafkaDatasetFacetProvider,
                        "For %s naming strategy, the recordSerializer needs to expose the target topics though implementing KafkaDatasetFacetProvider.",
                        transactionNamingStrategy);
            }
```

Note that Artem planned to extend ListTransaction API to look by transaction id prefix (which is what we actually want) which makes this requirement obsolete.

For all internal serializers, especially SQL and going through your builder, the interface is already implemented.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriterStateSerializer.java
AHeise: I can do it. Both patterns exist. Using >= you could potentially downgrade again if your state is forward compatible but that's rare.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionNamingStrategyImpl.java
```diff
@@ -0,0 +1,77 @@
...
+
+package org.apache.flink.connector.kafka.sink.internal;
+
+import org.apache.flink.connector.kafka.sink.KafkaSinkOptions.TransactionNamingStrategy;
+
+import static org.apache.flink.util.Preconditions.checkState;
+
+/** Implementation of {@link TransactionNamingStrategy}. */
+public enum TransactionNamingStrategyImpl {
+    INCREMENTING {
```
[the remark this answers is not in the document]
AHeise: I'd like to stick to INCREMENTING. ChatGPT says:

> The words "incrementing" and "incremental" are related but have different semantics:
> 
>     Incrementing (verb, present participle)
> 
>         Refers to the action of increasing something step by step.
> 
>         Example: "The counter is incrementing by one every second."
> 
>         It describes an ongoing process or operation of increasing something.
> 
>     Incremental (adjective)
> 
>         Describes something that happens in small, gradual steps.
> 
>         Example: "The software received incremental updates over time."
> 
>         It characterizes the nature of the increase rather than the action itself.
> 
> Summary:
> 
>     "Incrementing" → Ongoing process of increasing.
> 
>     "Incremental" → Gradual or step-by-step increase in nature.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java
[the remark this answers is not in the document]
AHeise: I added Internal now to all classes in that packages which results in random changes but I think it's for the better.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java
AHeise: I usually only use WARN for things that are actionable and would prefer INFO instead. But I'll abide by the majority here ;). (I agree that trace is way too low)

fapaul (approved):
Thanks for addressing the comments 👍

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
```diff
@@ -277,6 +286,13 @@ private void abortLingeringTransactions(
         TransactionAbortStrategyContextImpl context =
                 getTransactionAbortStrategyContext(startCheckpointId, prefixesToAbort);
         transactionAbortStrategy.abortTransactions(context);
+        Collection<String> unsafeTransactionalIds = context.getUnsafeTransactionalIds();
+        if (!unsafeTransactionalIds.isEmpty()) {
+            namingContext.setUnsafeTransactionalIds(unsafeTransactionalIds);
+            LOG.info(
+                    "Unsafe transactional ids {} detected. They will be skipped until first checkpoint.",
```
fapaul, line 293: I do not fully understand the comment. 

I guess this relys on the fact that the committer during recovery tries to commit all outstanding transactions so we do not need to wait for `notifyCheckpointComplete`? My concern is that transactions are only committed on `notifyCheckpointComplete` that might not be aligned with the general checkpointing cycle.

I also find `skipped` without context hard to understand. Maybe something like `reused` would be easier.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
```diff
@@ -293,22 +309,38 @@ private TransactionAbortStrategyContextImpl getTransactionAbortStrategyContext(
...
-        Set<String> ongoingTransactionIds =
+        Set<String> precommittedTransactionalIds =
                 recoveredStates.stream()
                         .flatMap(
                                 s ->
-                                        s.getOngoingTransactions().stream()
+                                        s.getPrecommittedTransactionalIds().stream()
                                                 .map(CheckpointTransaction::getTransactionalId))
                         .collect(Collectors.toSet());
+        Set<Integer> oldSubtaskIds =
```
fapaul, line 319: Nit: I wonder whether the `getTransactionAbortStrategyContext` method isn't better placed inside the `TransactionAbortStrategyContextImpl` as a builder method or similar

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
```diff
@@ -293,22 +309,38 @@ private TransactionAbortStrategyContextImpl getTransactionAbortStrategyContext(
...
+        Integer oldParallelism =
+                recoveredStates.stream()
+                        .map(KafkaWriterState::getParallelism)
+                        .findFirst()
+                        .orElse(null);
+        // Special case: recovery without checkpoint -> first subtask owns everything
+        if (recoveredStates.isEmpty() && kafkaSinkContext.getParallelInstanceId() == 0) {
+            oldSubtaskIds = Set.of(0);
+            oldParallelism = 1;
+        }
```
fapaul, line 332: I think the block is easier to understand with a dedicated method that returns immediately on the special condition. It also avoids reassigning the variables

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyContextImpl.java
```diff
@@ -104,4 +127,13 @@ public long getStartCheckpointId() {
     public TransactionAborter getTransactionAborter() {
         return transactionAborter;
     }
+
+    @Override
+    public void markTransactionalIdUnsafe(String transactionalId) {
+        this.unsafeTransactionalIds.add(transactionalId);
+    }
+
+    public Collection<String> getUnsafeTransactionalIds() {
```
fapaul, line 136: Why is this method not part of the interface?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
```diff
@@ -293,22 +309,38 @@ private TransactionAbortStrategyContextImpl getTransactionAbortStrategyContext(
...
+                recoveredStates.stream()
+                        .map(KafkaWriterState::getSubtaskId)
+                        .collect(Collectors.toSet());
+        Integer oldParallelism =
+                recoveredStates.stream()
+                        .map(KafkaWriterState::getParallelism)
+                        .findFirst()
+                        .orElse(null);
+        // Special case: recovery without checkpoint -> first subtask owns everything
+        if (recoveredStates.isEmpty() && kafkaSinkContext.getParallelInstanceId() == 0) {
```
fapaul, line 329: I might have missed something but I understood the current code as follows for this special case

1. oldSubtaskIds = empty, oldParallelism =1 
2. call `abortTransactions(Context context)`
3. Get all open transactions for prefix https://github.com/apache/flink-connector-kafka/blob/85942c398c253471d602116bddd55c33ff17db8f/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java#L132
4. Filter by "ownership" https://github.com/apache/flink-connector-kafka/blob/85942c398c253471d602116bddd55c33ff17db8f/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java#L139
5. For all open transactions for subtask we abort https://github.com/apache/flink-connector-kafka/blob/85942c398c253471d602116bddd55c33ff17db8f/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java#L156

Isn't Step 4 returning always false and we don't abort anything because of https://github.com/apache/flink-connector-kafka/blob/85942c398c253471d602116bddd55c33ff17db8f/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java#L209 being empty due to no state

fapaul (commented):
Minor comments but I do not fully understand the case of recovery with no state and a downscale and how we prevent closing all transactions.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
```diff
@@ -57,6 +58,7 @@
 
 import static org.apache.flink.util.IOUtils.closeAll;
 import static org.apache.flink.util.Preconditions.checkNotNull;
+import static org.apache.flink.util.Preconditions.checkState;
 
 /**
```
fapaul, line 63: Can you create a high-level doc string about how the transaction mechanism works that future readers do not have to fully reverse engineer the code? :)

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriterStateSerializer.java
```diff
@@ -59,14 +61,20 @@ public KafkaWriterState deserialize(int version, byte[] serialized) throws IOExc
         try (final ByteArrayInputStream bais = new ByteArrayInputStream(serialized);
                 final DataInputStream in = new DataInputStream(bais)) {
             final String transactionalIdPrefix = in.readUTF();
+            int subtaskId = 0;
```
fapaul, line 64: Is this safe to do in case of a state migration now we have multiple states from subtaskId 0?

fapaul (commented):
Looks mostly good one last comment.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java
```diff
@@ -0,0 +1,215 @@
...
+            return context.getPrefixesToAbort().stream().anyMatch(name::startsWith);
+        }
+    };
+
+    private static final Logger LOG = LoggerFactory.getLogger(TransactionAbortStrategyImpl.class);
+
+    /** Aborts all transactions that have been created by this subtask in a previous run. */
+    public abstract void abortTransactions(Context context);
+
+    /** Injects the actual abortion of the transactional id generated by one of the strategies. */
```
davidradl, line 168: Please could we use the word `abort` `aborted` `aborting`, which seems to be the words used in reference to transactions. I think we should avoid `abortion` in any text or variable/method name, as it could be triggering.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriterStateSerializer.java
AHeise: It's a good point. I suggest to add a special variable UNKNOWN=-1 and disallow using POOLING then. In that way, we can guide the user to safely convert the checkpoint first to INCREMENTING and then POOLING.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
AHeise: I added a new class TransactionOwnership that is used to document things but also to check the migration path.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriterStateSerializer.java
AHeise: I added a new class TransactionOwnership that is used to document things but also to check the migration path.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionOwnership.java
```diff
@@ -0,0 +1,205 @@
...
+                return currentParallelism;
+            }
+            int totalNumberOfOwnedSubtasks =
+                    recoveredStates.iterator().next().getTotalNumberOfOwnedSubtasks();
+            assertKnown(totalNumberOfOwnedSubtasks);
+
+            if (currentParallelism >= totalNumberOfOwnedSubtasks) {
+                // Assumption of the ownership model: state is distributed consecutively across the
+                // subtasks starting with subtask 0
+                checkState(recoveredStates.size() == 1, "Not uniformly assigned");
```
fapaul, line 182: Is this an upscale scenario? 

Shouldn't the state assignment be 0 or 1?

fapaul (approved):
LGTM left one last clarification question

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionOwnership.java
AHeise: Yes that's correct. But the case 0 cannot happen because of `if (recoveredStates.isEmpty())`

PR #155, opened 2025-02-26
[FLINK-37366] Allow configurable retry for Kafka topic metadata fetch

suez1224:
TBD


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/subscriber/KafkaSubscriberUtils.java
```diff
@@ -57,12 +65,42 @@ static Map<String, TopicDescription> getTopicMetadata(
...
+                            topicNames,
+                            retryDelay);
+                    try {
+                        TimeUnit.MILLISECONDS.sleep(retryDelay);
+                    } catch (InterruptedException ie) {
+                        Thread.currentThread().interrupt(); // Restore interrupted state
+                        LOG.error("Thread was interrupted during metadata fetch retry delay.", ie);
+                    }
+                }
+            }
```
AHeise, line 99: Instead of introducing our own retry logic, can't we reuse what's already implemented in the KafkaAdmin?

```
    public static final String RETRIES_CONFIG = "retries";
    public static final String RETRIES_DOC = "Setting a value greater than zero will cause the client to resend any request that fails with a potentially transient error." +
        " It is recommended to set the value to either zero or `MAX_VALUE` and use corresponding timeout parameters to control how long a client should retry a request.";

    public static final String RETRY_BACKOFF_MS_CONFIG = "retry.backoff.ms";
    public static final String RETRY_BACKOFF_MS_DOC = "The amount of time to wait before attempting to retry a failed request to a given topic partition. This avoids repeatedly sending requests in a tight loop under some failure scenarios.";
```

You should be able to set it with even without your PR like this
```
  properties.retries = '10',
  properties.retry.backoff.ms = '30000',
```
But that also influences the consumer retry behavior.

We could think about supporting `properties.admin.retry = 10`. WDYT?

AHeise (commented):
Before doing a detailed review. I have a fundamental question on the approach that I inlined. PTAL

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/subscriber/KafkaSubscriberUtils.java
suez1224: thanks for the suggestion, @AHeise . However, the default value for the `retries` config is already set to MAX_INT (see [code](https://github.com/apache/kafka/blob/3.4/clients/src/main/java/org/apache/kafka/clients/admin/AdminClientConfig.java#L179-L184) and [confluent doc here](https://docs.confluent.io/platform/current/installation/configuration/admin-configs.html#retries). And I believe Flink does not overwrite the config value. w/o my PR, the flink job will fail as soon as the metadata request fails. So I don't think this config control the behavior for failed metadata requests from AdminClient.

AHeise: Could you please share the failure? If it's something that the AdminClient doesn't deem to be retriable than it's obvious that we need your solution.

It actually may be worth to draft a test that fails without your fix and succeeds with it. We do have some tests that remove the kafka broker (search for `stopBroker`).

PR #156, opened 2025-02-27
[FLINK-37380] Change TransactionalIdPrefix to a required option if Exactly once 

loserwang1024:
As described in https://issues.apache.org/jira/browse/FLINK-37380 , though org.apache.flink.connector.kafka.sink.KafkaSinkBuilder#sanityCheck said `EXACTLY_ONCE delivery guarantee requires a transactionIdPrefix to be set to provide unique transaction names across multiple KafkaSinks writing to the same Kafka cluster` ,  it 's no use because the transactionalIdPrefix with a default value "kafka-sink" will never be null.

This PR just remove the default value, and make the sanityCheck valid.


loserwang1024:
@AHeise @leonardBang @PatrickRen  , CC, WDYT?

leonardBang (approved):
+1 from my side

loserwang1024:
All tests pass exception python tests.( which is no related with this pr, See: https://issues.apache.org/jira/browse/FLINK-37396)

AHeise:
A late +1 from my side :). Did you check if documentation (for SQL) reflects that this is a required property? @loserwang1024

loserwang1024:
> Did you check if documentation (for SQL) reflects that this is a required property?

image

Yes, document has already showed it before this pr.

PR #157, opened 2025-02-27
[FLINK-37396] Fix setup.py: multiple top-level packages discovered in a flat-layout

loserwang1024:
Inspired by [Multiple top-level packages discovered in a flat-layout](https://stackoverflow.com/questions/72294299/multiple-top-level-packages-discovered-in-a-flat-layout) , add `py_modules=[]` to fix python CI test [FLINK-37396](https://issues.apache.org/jira/browse/FLINK-37396).

### DIFF
--- a/flink-python/setup.py
+++ b/flink-python/setup.py
@@ -133,6 +133,7 @@ if in_flink_source:
 
 # Runs the python setup
 setup(
+    py_modules=[],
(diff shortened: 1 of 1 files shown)


loserwang1024:
@leonardBang , CC

leonardBang (approved):
Thank for the fix, LGTM

PR #158, opened 2025-02-28
[FLINK-25920] bump flink version to 1.20.1.

loserwang1024:
Currently, test often error with "Currently it is not supported to update the CommittableSummary for a checkpoint coming from the same subtask. Please check the status of FLINK-25920"

https://issues.apache.org/jira/browse/FLINK-25920 hava fixed it, thus bump flink version to 1.20.1 to reduce test error.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,7 +50,7 @@ under the License.
     </modules>
 
     <properties>
-        <flink.version>1.20.0</flink.version>
+        <flink.version>1.20.1</flink.version>
         <kafka.version>3.4.0</kafka.version>
         <confluent.version>7.4.4</confluent.version>
 


AHeise:
I think we should at least bump on respective release branches 3.4 and 3.3?

loserwang1024:
@prshnt ,
If we upgraded to 2.0.0 , the kafka connector maybw not compabality with former flink version(1.19, 1.18, 1.20).

I think we should upgrade to  2.0.0 in a major version! @AHeise , WDYT?

prshnt:
@davidradl hi I think you got it right and the proposal does make sense. I believe there was a separate PR for 2.0.0 that has already been merged so we can go ahead with this now.

loserwang1024:
@prshnt ,  I agree with you. I also noticed that Kafka 4.0 is released recently. Should Kafka connector v4.0 supported it? or we left it in kafka connector v4.0 until the kafka 4.0 is stable.

prshnt:
@loserwang1024 We would need to support the older and the newer versions, need to see what breaking changes the 4.0 has introduced. I believe we have Kafka tests with Zookeeper which we would need to keep and also have additional tests with KRaft for 4.x.x Kafka supporting versions, we can wait a bit on that one what do you think?

loserwang1024:
@prshnt I will close this pr, let's  go forwards!

PR #159, opened 2025-03-26
[FLINK-37356] Recreate producer on error

AHeise:
Reusing a producer in erroneous state can result in further issues that are outside the planned state machine.


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/ProducerPoolImpl.java
```diff
@@ -131,26 +131,33 @@ public void recycleByTransactionId(String transactionalId) {
...
     public void recycle(FlinkKafkaInternalProducer<byte[], byte[]> producer) {
-        recycleProducer(producer);
+        recycleProducer(producer, true);
         ProducerEntry producerEntry =
                 producerByTransactionalId.remove(producer.getTransactionalId());
         transactionalIdsByCheckpoint.remove(producerEntry.getCheckpointedTransaction());
     }
 
-    private void recycleProducer(@Nullable FlinkKafkaInternalProducer<byte[], byte[]> producer) {
+    private void recycleProducer(
```
fapaul, line 148: Nit: Maybe it makes sense to handle the success criteria `recycleByTransactionId` and do not call `recyleProducer` at all if the transaction wasn't finished successfully.

It removes the additional changes that are only caused by the signature change.

fapaul (commented):
Can you add some more decription to the commit message in which scenarios we can get a failed transaction?

I would have expected many of the cases anyway restart the job due to risk of data loss.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/ProducerPoolImpl.java
AHeise: It's a good idea but it makes the call-site messier. There are a total of 4 cases now:
* Producer is recycled in writer (abort loop). That would go well with your proposal.
* Committer regularly commits and recycles. That would also go well with your proposal.
* Special case 1: on recovery, committer commits and recycles the id, in that case the producer entry doesn't contain an internal producer and is null. So we would need to handle the null on call-site.
* Special case 2: on non-chained committer, we may end up with having more transactions in the state than there are running (see the comment in `recycleByTransactionId`). Here, we also may have null values. 

So in cases where the producer is null or reuse is false, it gets messier.

AHeise:
> Can you add some more decription to the commit message in which scenarios we can get a failed transaction?
> 
> I would have expected many of the cases anyway restart the job due to risk of data loss.

The short answer is in all cases with a known error where we don't fall over. See the committer. At this point, I'm not sure if all of these cases actually make sense or if a fall-over is better but I'd leave that for the future. At this point, we just know that there are cases that cause producers to be in invalid state after an error.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/ProducerPoolImpl.java
fapaul: My idea would be to add a simple check to `recycleByTransactionId` that `recycleProducer` is only invoked if `sucess` is true if false we close the producer. 

Afaict `recyleProducer` with success false is only called from `recycleByTransactionId` so the code path is only hit in a single scenario. 

This way we should be able to remove the parameter from `recyleProducer`.

AHeise: Okay I'll try to solve it in that way.

AHeise: Looks better this way. Thanks!

fapaul (approved):
Thanks for addressing the comments.

We discussed the changes of [667ff6d](https://github.com/apache/flink-connector-kafka/pull/159/commits/667ff6d8eabea6404507ae1895c7c3a8cd43f764) offline and agreed that it now becomes possible after recreating the producer in failure cases 👍

PR #161, opened 2025-03-28
[FLINK-37583] Upgrade to Kafka 4.0.0 client.

tomncooper:
Kafka Client version 4.0.0 has been released. This has many performance improvements and security fixes. This version is backwards compatible with older Kafka versions, however as noted in the [release notes](https://kafka.apache.org/40/documentation.html#upgrade_servers_400_notable):

>Old protocol API versions have been removed. Users should ensure brokers are version 2.1 or higher before upgrading Java clients (including Connect and Kafka Streams which use the clients internally) to 4.0.

Therefore, moving to this version will mean that the connector cannot be used with Kafka clusters running Kafka version 2.0 or older. Hopefully, users have upgraded beyond this point by now but it will need to be stated clearly in the release notes for the version containing this commit.


MartijnVisser:
@tomncooper I suspect that this will fail, given that the Flink Kafka connector relies on reflection for the two-phase commit part.

MartijnVisser:
> Are there other tests I should run (perhaps in the main Flink repo) to verify?

Yeah, `mvn clean install -Dflink.version=2.0.0 -Dscala-2.12 -Prun-end-to-end-tests -DdistDir=/Users/mvisser/Developer/flink-2.0.0 -Dflink.convergence.phase=install -Dlog4j.configurationFile=tools/ci/log4j.properties`  is what CI also runs (and has failed with unfortunately)

tomncooper:
So I have run: 
```shell
mvn clean install -Prun-end-to-end-tests -Dflink.version=2.0.0 -Dscala-2.12 -DdistDir=$HOME/tools/flink/flink-2.0.0 -Dflink.convergence.phase=install -Dlog4j.configurationFile=tools/ci/log4j.properties
```
With both Java 17 and 21 locally and all tests are passing 🤔

tomncooper:
Ok, that run passed the e2e tests but failed on the licence check.

tomncooper:
Hummm, so I assume [this is the test](https://github.com/apache/flink-connector-shared-utils/blob/fbe1df6345293db31e592699d64b028673ab139b/.github/workflows/ci.yml#L183) that is failing in CI. But when I run the same locally:
```shell
mvn exec:java@check-license -N -Dexec.args="/tmp/flink-maven-build-output $(pwd) /tmp/flink-validation-deployment" -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120 -Dlog4j.configurationFile=file://$(pwd)/tools/ci/log4j.properties
```
It passes?

davidradl:
@tomncooper @MartijnVisser @AHeise I suggest we move to Kafka 4.0..0 client when we do Kafka connector v4 for Flink 2. 
WDYT? ( assuming we can sort out the tests)    

I think there is a case to say we do not backport Kafka client v4.0.0 support to Kafka connector v3.3 or 3.4, in case there are old Kafka clusters that we would not want to break on the v3 stream.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaWriterFaultToleranceITCase.java
```diff
@@ -40,7 +40,7 @@ public class KafkaWriterFaultToleranceITCase extends KafkaWriterTestBase {
     private static final String INIT_KAFKA_RETRIES = "0";
     private static final String INIT_KAFKA_REQUEST_TIMEOUT_MS = "1000";
     private static final String INIT_KAFKA_MAX_BLOCK_MS = "1000";
-    private static final String INIT_KAFKA_DELIVERY_TIMEOUT_MS = "1000";
+    private static final String INIT_KAFKA_DELIVERY_TIMEOUT_MS = "1500";
```
davidradl, line 43: I may have missed it, but could you give some context around this change in timeout value.

tomncooper: Sorry, I should have put a comment in the PR description about this. The new kafka client now enforces the requirement for the delivery timeout to be longer than the request timeout (which is part of the overall delivery timeout) at compile time.

tomncooper:
> @tomncooper @MartijnVisser @AHeise I suggest we move to Kafka 4.0..0 client when we do Kafka connector v4 for Flink 2. WDYT? ( assuming we can sort out the tests)
> 
> I think there is a case to say we do not backport Kafka client v4.0.0 support to Kafka connector v3.3 or 3.4, in case there are old Kafka clusters that we would not want to break on the v3 stream.

@davidradl As per the [discussion](https://lists.apache.org/thread/l0kzx2wb2qz7ntcv6wkbbfwx688y0o69) on the dev mailing list, we are going to move ahead with a Connector 4.0 release with Flink 2.0 and Kafka 3.9.0. We can then do point release updates.

I still think, given that this PR would drop support for older Kafka versions it should be part of a further major version bump (ie 5.0) but we can have that discussion when the time comes to merge this.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaWriterFaultToleranceITCase.java
davidradl: thanks for the explanation - I suggest a comment.

fapaul:
Thanks for the thorough explanation. In summary, in addition to dropping support for 0.x versions (has been done in the past already), we also now drop support for everything, including 2.0. 

I am okay with the major version bump then. Can you look at the last failing tests to get this PR merged?

Also I noted the docs are slightly out of date https://nightlies.apache.org/flink/flink-docs-master/docs/connectors/datastream/kafka/#dependency still states 
> Modern Kafka clients are backwards compatible with broker versions 0.10.0 or later. For details on Kafka compatibility, please refer to the official [Kafka documentation](https://kafka.apache.org/protocol.html#protocol_compatibility).

Can you also update these docs to reflect that with this change, we expect the Kafka broker to be at least version 2.1?

tomncooper:
@fapaul This PR already updated the docs to say Kafka 2.1.0 or later. Do you think it needs more in depth description? I am happy to add that but will need to create a JIRA for the Chinese language changes.

tomncooper:
Ok I am stumped? Locally I run (with maven 3.8.6 and java 17):
```shell
mvn clean deploy -U -B --no-transfer-progress -Dflink.version=2.0.0 \
    -DaltDeploymentRepository=validation_repository::default::file:/tmp/flink-validation-deployment \
    -Dscala-2.12 \
    -Prun-end-to-end-tests -DdistDir="$HOME/tools/flink/flink-2.0.0" \
    -Dflink.convergence.phase=install -Pcheck-convergence \
    -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120 \
    -Dlog4j.configurationFile="file://$(pwd)/tools/ci/log4j.properties" \
   | tee /tmp/mvn_build_output.out
```
Then the licence checker:
```shell
mvn -U -B --no-transfer-progress -Dflink.version=2.0.0 \
    exec:java@check-license -N \
    -Dexec.args="/tmp/mvn_build_output.out $(pwd) /tmp/flink-validation-deployment" \
    -Dhttp.keepAlive=false \
    -Dmaven.wagon.http.pool=false \
    -Dmaven.wagon.httpconnectionManager.ttlSeconds=120 \
    -Dlog4j.configurationFile=file://$(pwd)/tools/ci/log4j.properties
 ```
 Which as far as I can tell is what the CI is running and it passes?

tomncooper:
Ok, I have no idea why but rebasing on main has lead to [green tests](https://github.com/tomncooper/flink-connector-kafka/actions/runs/16470498973) on my fork. 

I added an override to the CI log4j config to make sure the licence checker is able to output its logs.

@fapaul Are you able to kick the CI on this PR?

fapaul (approved):
Can you squash the commits on this PR and rebase on the latest master?

PR #165, opened 2025-04-04
[FLINK-37613] Fix resource leak during abortion

AHeise:
Transaction abortion can take a while. Apparently, task cleanup may not trigger correctly if this abortion is interrupted (for some unrelated reason). This leaks producers. The fix is to close() manually during initialization of the EOS writer until we fixed the Flink bug FLINK-37612.

### DIFF
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducerITCase.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducerITCase.java
@@ -34,11 +34,13 @@ import org.apache.kafka.common.errors.InvalidTxnStateException;
 import org.apache.kafka.common.errors.ProducerFencedException;
 import org.apache.kafka.common.serialization.StringDeserializer;
 import org.apache.kafka.common.serialization.StringSerializer;
+import org.assertj.core.api.AbstractThrowableAssert;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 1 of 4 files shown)


flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducerITCase.java
```diff
@@ -174,6 +177,51 @@ void testResetInnerTransactionIfFinalizingTransactionFailed(
...
+                            });
+            if (firstCommit == secondCommit) {
+                secondOp.doesNotThrowAnyException();
+            } else {
+                secondOp.isInstanceOf(InvalidTxnStateException.class);
+            }
+        }
+
+        assertNumTransactions(1, transactionIdPrefix);
+        assertThat(readRecords(topic).count()).isEqualTo(firstCommit ? 1 : 0);
```
rkhachatryan, line 222: This is testing Producer, which wasn't actually changed in this PR, wasn't it?

rkhachatryan (approved):
Thanks for the fix, production code changes LGTM.
As for the test, it looks like it doesn't cover the change (but I wouldn't block the fix on it)

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducerITCase.java
AHeise: Yes, I was using the test to verify the expected behavior around kafka transactions (e.g. double commits are fine). It's not covering the changes but it's a test that I thought we already have. So I'd like to keep it.

PR #167, opened 2025-04-09
[FLINK-37644] Include guava dependency and shade it.

lvyanquan:
Include guava dependency to avoid ClassNotFoundException, and shade it to avoid conflict.


lvyanquan:
At present, there is a problem of LICENSE check failure in CI, and I am trying to solve it

AHeise:
I was wondering if we should already shade in the datastream connector. My main reason is that if a datastream application uses multiple connectors, we will already have dependency hell there.

AHeise:
I'm currently trying to see if we can replace guava entirely and it's surprisingly easy. I'm currently running tests locally and will open a PR soonish.

lvyanquan:
> I'm currently trying to see if we can replace guava entirely and it's surprisingly easy. I'm currently running tests locally and will open a PR soonish.
> 
> Most of the stuff is not needed anymore in java 11. TypeToken can be replaced by TypeExtractor which is public API.

Thank you for fixing it. I haven't run CI for the issue with LICENSE yet. If you create a new PR, I will close the current one.

lvyanquan:
Hi @AHeise, The current code should be able to make CI green.
you can help trigger it, then make a choice between my changes and yours.

tomncooper (approved):
LGTM. Obviously, if possible, removing guava all together would be better.

lvyanquan:
Closed as https://github.com/apache/flink-connector-kafka/pull/170 provided a more thorough solution

PR #170, opened 2025-04-10
[FLINK-37644] Straighten dependencies

AHeise:
We have a couple of dependencies that are not properly shaded for SQL. Guava is also bad as a direct dependency for a connector because chances are high that we clash with other connectors or user code dependencies. This PR removes rarely used dependencies by using modern Java features or built-in Flink APIs.

Additionally, to ensure that future PRs are not resetting the state enforcer and dependency plugin + checkstyle will ensure that guava is not finding its way back.


lvyanquan (approved):
LGTM.

PR #171, opened 2025-04-16
[FLINK-34466][LINEAGE] Support dataset type facet for Avro

pawel-big-lebowski:
With an implementation of lineage interfaces, Kafka source and sink provide `TypeDatasetFacet` which contains type information of data processed. This was working with extracting type information from `SerializationSchema`:
```
TypeExtractor.getParameterType(SerializationSchema.class, valueSerialization.getClass(), 0)
```

However, this ain't working for `ConfluentRegistryAvroSerializationSchema` due to type erasure. 

As a workaround, `TypeDatasetFacet` can be extended to contain `SerializationSchema` when direct extraction of `TypeInformation` is not possible. This allows handling this case on the lineage listener side, while keeping flink-connector-kafka unaware of Avro type extraction logic.

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/TypeDatasetFacet.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/TypeDatasetFacet.java
@@ -1,11 +1,26 @@
 package org.apache.flink.connector.kafka.lineage;
 
 import org.apache.flink.annotation.PublicEvolving;
+import org.apache.flink.api.common.serialization.SerializationSchema;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.streaming.api.lineage.LineageDatasetFacet;
+
+import java.util.Optional;
 
 /** Facet definition to contain type information of source and sink. */
 @PublicEvolving
 public interface TypeDatasetFacet extends LineageDatasetFacet {
     TypeInformation getTypeInformation();
+
+    /**
+     * Sometimes a sink implementing {@link TypeDatasetFacetProvider} is not able to extract type.
+     * This is happening for AvroSerializationSchema due to type erasure problem. In this case, it
+     * makes sense to expose SerializationSchema to the lineage consumer so that it can use it to
+     * extract type information.
+     *
+     * @return
+     */
+    default Optional<SerializationSchema> getSerializationSchema() {
+        return Optional.empty();
+    }
 }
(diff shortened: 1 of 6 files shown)


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/TypeDatasetFacet.java
```diff
@@ -1,11 +1,26 @@
...
+
+    /**
+     * Sometimes a sink implementing {@link TypeDatasetFacetProvider} is not able to extract type.
+     * This is happening for AvroSerializationSchema due to type erasure problem. In this case, it
+     * makes sense to expose SerializationSchema to the lineage consumer so that it can use it to
+     * extract type information.
+     *
+     * @return
+     */
+    default Optional<SerializationSchema> getSerializationSchema() {
```
davidradl, line 23: Shouldn't this interface live in Flink? As I assume a File connector using the Avro format could hit this issue as well.

ferenc-csaky: I strongly agree. The whole `TypeDatasetFacet` should be removed from here. As far as i can tell, this PR should be revived and merged beforehand: https://github.com/apache/flink/pull/25712

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaRecordSerializationSchemaBuilder.java
```diff
@@ -432,13 +433,17 @@ public Optional<TypeDatasetFacet> getTypeDatasetFacet() {
...
                 try {
+                    Type type =
+                            TypeExtractor.getParameterType(
+                                    SerializationSchema.class,
+                                    valueSerializationSchema.getClass(),
+                                    0);
+
                     return Optional.of(
                             new DefaultTypeDatasetFacet(TypeExtractor.createTypeInfo(type)));
+                } catch (InvalidTypesException e) {
```
Efrat19, line 445: Wouldn't a type check beforehand be less expensive resource-wise? (Depending on how often do you expect to hit this fallback)

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchema.java
```diff
@@ -211,12 +212,14 @@ public Optional<TypeDatasetFacet> getTypeDatasetFacet() {
...
-            Type type =
-                    TypeExtractor.getParameterType(
-                            SerializationSchema.class, valueSerialization.getClass(), 0);
             try {
+                Type type =
+                        TypeExtractor.getParameterType(
+                                SerializationSchema.class, valueSerialization.getClass(), 0);
+
                 return Optional.of(new DefaultTypeDatasetFacet(TypeExtractor.createTypeInfo(type)));
+            } catch (InvalidTypesException e) {
```
Efrat19, line 221: Same as https://github.com/apache/flink-connector-kafka/pull/171/changes#r2659457226

PR #174, opened 2025-05-09
[FLINK-32609] Support Projection Pushdown

fqshopify:
Implements `SupportsProjectionPushDown` for `KafkaDynamicSource`, allowing queries to skip deserializing columns they don't need.

### How it works

The source always reports to the planner that it supports projection pushdown (including nested projection pushdown) so the planner pushes projections down to the connector. Whether those projections are *also* pushed all the way down into the underlying key/value formats - so the formats only deserialize the required fields - is opt-in per key/value. When enabled, the format pushdown level is determined automatically based on whether the format implements `ProjectableDecodingFormat` and `ProjectableDecodingFormat.supportsNestedProjection()`.

For each of key and value, the `Decoder` picks one of three strategies:

| Condition | Strategy |
|-----------|----------|
| Pushdown disabled or<br>Pushdown enabled + format does not support projection pushdown | Deserialize all fields, then project afterward |
| Pushdown enabled + format supports nested projection pushdown | Push the full (nested) projection into the format |
| Pushdown enabled + format supports only top-level projection pushdown | Push top-level fields into the format, extract nested sub-fields afterward |

### Configuration

| Option | Default | Type |
|--------|---------|------|
| `key.format.projection-pushdown.enabled` | `false` | boolean |
| `value.format.projection-pushdown.enabled` | `false` | boolean |

### Benefits

1. **Better Schema evolution tolerance** - Queries only fail on breaking changes to fields they actually use. For example, if column `a` changes from `INT` to `STRING`, the query `SELECT b FROM kafka` still succeeds because `a` is never deserialized. (Only applies to formats that decode fields independently, e.g. `json`, `avro-confluent`, `debezium-avro-confluent`.)

2. **Better Performance** - Unneeded columns are filtered at the `TableSourceScan` node rather than downstream. 

### Why opt-in?

If a format implements projection-pushdown incorrectly, it can cause runtime errors or maybe even incorrect results. Enabling is therefore opt-in. Notably, the `avro` format does not support projection pushdown correctly (see [FLINK-35324](https://issues.apache.org/jira/browse/FLINK-35324)).

Enabling it should be safe for other formats e.g. `json`, `debezium-json-confluent`, `csv`, `avro-confluent`, `debezium-avro-confluent`.

### Why not push projections all the way down into Kafka?

Kafka does not support projection pushdown so the furthest we can push down the projections is into the formats (if applicable).

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/Decoder.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/Decoder.java
@@ -1,0 +1,471 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.streaming.connectors.kafka.table;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.api.common.serialization.DeserializationSchema;
+import org.apache.flink.table.connector.Projection;
+import org.apache.flink.table.connector.format.DecodingFormat;
+import org.apache.flink.table.connector.format.ProjectableDecodingFormat;
+import org.apache.flink.table.connector.source.DynamicTableSource.Context;
+import org.apache.flink.table.data.GenericRowData;
+import org.apache.flink.table.data.RowData;
+import org.apache.flink.table.types.DataType;
+import org.apache.flink.table.types.logical.RowType;
+
+import javax.annotation.Nullable;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/**
+ * Decoding messages consists of two potential steps:
+ *
+ * <ol>
+ *   <li>Deserialization i.e deserializing the {@code byte[]} into a {@link RowData}. This process
+ *       is handled by a {@link DeserializationSchema}.
+ *   <li>Projection i.e. projecting any required fields from the deserialized {@link RowData}
+ *       (returned by the {@link DeserializationSchema} in the first step) to their positions in the
+ *       final produced {@link RowData}. This process is handled by a {@link Projector}.
+ * </ol>
+ *
+ * <p>In order to decode messages correctly, the {@link DeserializationSchema} and the {@link
+ * Projector} need to work together. For example, the {@link Projector} needs to know the positions
+ * of the required fields in the {@link RowData} returned by the {@link DeserializationSchema} in
+ * order to be able to correctly set fields in the final produced {@link RowData}.
+ *
+ * <p>That's why we have this {@link Decoder} class. This class ensures that the returned {@link
+ * DeserializationSchema} and {@link Projector} will work together to decode messages correctly.
+ */
+@Internal
+public class Decoder {
+
+    /**
+     * Can be null. Null is used inside {@link DynamicKafkaDeserializationSchema} to avoid
+     * deserializing keys if not required.
+     */
+    private final @Nullable DeserializationSchema<RowData> deserializationSchema;
+
+    /** Mapping of the physical position in the key to the target position in the RowData. */
+    private final Projector projector;
+
+    private Decoder(
+            final DeserializationSchema<RowData> deserializationSchema, final Projector projector) {
+        this.deserializationSchema = deserializationSchema;
+        this.projector = projector;
+    }
+
+    /**
+     * @param decodingFormat Optional format for decoding bytes.
+     * @param physicalTableDataType The data type representing the table schema.
+     * @param physicalDataTypeProjection Indices indicate the position of the field in the dataType
+     *     (key/value). Values indicate the position of the field in the tableSchema.
+     * @param prefix Optional field prefix
+     * @param projectedPhysicalFields Indices indicate the position of the field in the produced
+     *     Row. Values indicate the position of the field in the table schema.
+     * @param pushProjectionsIntoDecodingFormat if this is true and the format is a {@link
+     *     ProjectableDecodingFormat}, any {@param projectedPhysicalFields} will be pushed down into
+     *     the {@link ProjectableDecodingFormat}. Otherwise, projections will be applied after
+     *     deserialization.
+     * @return a {@link Decoder} instance.
+     */
+    public static Decoder create(
+            final Context context,
+            final @Nullable DecodingFormat<DeserializationSchema<RowData>> decodingFormat,
+            final DataType physicalTableDataType,
+            final int[] physicalDataTypeProjection,
+            final @Nullable String prefix,
+            final int[][] projectedPhysicalFields,
+            final List<String> metadataKeys,
+            final boolean pushProjectionsIntoDecodingFormat) {
+        if (decodingFormat == null) {
+            return Decoder.noDeserializationOrProjection();
+        } else if (!pushProjectionsIntoDecodingFormat
+                || !(decodingFormat instanceof ProjectableDecodingFormat)) {
+            return Decoder.projectAfterDeserializing(
+                    context,
+                    decodingFormat,
+                    physicalTableDataType,
+                    physicalDataTypeProjection,
+                    prefix,
+                    projectedPhysicalFields,
+                    metadataKeys);
+        } else {
+            final ProjectableDecodingFormat<DeserializationSchema<RowData>>
+                    projectableDecodingFormat =
+                            (ProjectableDecodingFormat<DeserializationSchema<RowData>>)
+                                    decodingFormat;
+            if (projectableDecodingFormat.supportsNestedProjection()) {
+                return Decoder.projectInsideDeserializer(
+                        context,
+                        projectableDecodingFormat,
+                        physicalTableDataType,
+                        physicalDataTypeProjection,
+                        prefix,
+                        projectedPhysicalFields,
+                        metadataKeys);
+            } else {
+                return Decoder.projectTopLevelInsideDeserializerThenNestedAfter(
+                        context,
+                        projectableDecodingFormat,
+                        physicalTableDataType,
+                        physicalDataTypeProjection,
+                        prefix,
+                        projectedPhysicalFields,
+                        metadataKeys);
+            }
+        }
+    }
+
+    /**
+     * @return a {@link DeserializationSchema} or null.
+     */
+    @Nullable
+    public DeserializationSchema<RowData> getDeserializationSchema() {
+        return deserializationSchema;
+    }
+
+    /**
+     * @return a {@link Projector}.
+     */
+    public Projector getProjector() {
+        return projector;
+    }
+
+    /**
+     * Creates an identity projection array where each field in the row type maps to itself.
+     *
+     * @param rowType the row type representing the table schema
+     * @return an int[][] with one entry per field in the row type, where entry i is {i}
+     */
+    public static int[][] identityProjection(final RowType rowType) {
+        final int tableSchemaSize = rowType.getFieldCount();
+        final int[][] projectedFields = new int[tableSchemaSize][];
+        for (int i = 0; i < tableSchemaSize; i++) {
+            projectedFields[i] = new int[] {i};
+        }
+        return projectedFields;
+    }
+
+    private static Decoder noDeserializationOrProjection() {
+        return new Decoder(null, new ProjectorImpl(Collections.emptyMap(), 0, 0));
+    }
+
+    private static DataType toPhysicalDataType(
+            final DataType physicalTableDataType,
+            final int[] physicalDataTypeProjection,
+            final @Nullable String prefix) {
+        final DataType temp =
+                Projection.of(physicalDataTypeProjection).project(physicalTableDataType);
+        return Optional.ofNullable(prefix)
+                .map(s -> TableDataTypeUtils.stripRowPrefix(temp, s))
+                .orElse(temp);
+    }
+
+    private static Map<Integer, Integer> tableToDeserializedTopLevelPos(
+            final int[] dataTypeProjection) {
+        final HashMap<Integer, Integer> tableToDeserializedPos = new HashMap<>();
+        for (int i = 0; i < dataTypeProjection.length; i++) {
+            tableToDeserializedPos.put(dataTypeProjection[i], i);
+        }
+        return tableToDeserializedPos;
+    }
+
+    private static int[] copyArray(final int[] arr) {
+        return Arrays.copyOf(arr, arr.length);
+    }
+
+    private static void addMetadataProjections(
+            final DecodingFormat<?> decodingFormat,
+            final int deserializedSize,
+            final int physicalSize,
+            final List<String> requestedMetadataKeys,
+            final Map<List<Integer>, Integer> deserializedToProducedPos) {
+
+        if (!requestedMetadataKeys.isEmpty()) {
+            decodingFormat.applyReadableMetadata(requestedMetadataKeys);
+
+            // project only requested metadata keys
+            for (int i = 0; i < requestedMetadataKeys.size(); i++) {
+                // metadata is always added to the end of the deserialized row by the DecodingFormat
+                final int deserializedPos = deserializedSize + i;
+                // we need to always add metadata to the end of the produced row
+                final int producePos = physicalSize + i;
+                deserializedToProducedPos.put(
+                        Collections.singletonList(deserializedPos), producePos);
+            }
+        }
+    }
+
+    /**
+     * This method generates a {@link Decoder} which pushes projections down directly into the
+     * {@link ProjectableDecodingFormat} which takes care of projecting the fields during the
+     * deserialization process itself.
+     */
+    private static Decoder projectInsideDeserializer(
+            final Context context,
+            final ProjectableDecodingFormat<DeserializationSchema<RowData>>
+                    projectableDecodingFormat,
+            final DataType physicalTableDataType,
+            final int[] physicalDataTypeProjection,
+            final @Nullable String prefix,
+            final int[][] projectedPhysicalFields,
+            final List<String> metadataKeys) {
+        final Map<Integer, Integer> tableToDeserializedTopLevelPos =
+                tableToDeserializedTopLevelPos(physicalDataTypeProjection);
+
+        final List<int[]> deserializerProjectedFields = new ArrayList<>();
+        final Map<List<Integer>, Integer> deserializedToProducedPos = new HashMap<>();
+        for (int producedPos = 0; producedPos < projectedPhysicalFields.length; producedPos++) {
+            final int[] tablePos = projectedPhysicalFields[producedPos];
+            final int tableTopLevelPos = tablePos[0];
+
+            final Integer dataTypeTopLevelPos =
+                    tableToDeserializedTopLevelPos.get(tableTopLevelPos);
+            if (dataTypeTopLevelPos != null) {
+                final int[] dataTypePos = copyArray(tablePos);
+                dataTypePos[0] = dataTypeTopLevelPos;
+
+                deserializerProjectedFields.add(dataTypePos);
+
+                final int deserializedPos = deserializerProjectedFields.size() - 1;
+                deserializedToProducedPos.put(
+                        Collections.singletonList(deserializedPos), producedPos);
+            }
+        }
+
+        addMetadataProjections(
+                projectableDecodingFormat,
+                deserializerProjectedFields.size(),
+                projectedPhysicalFields.length,
+                metadataKeys,
+                deserializedToProducedPos);
+
+        return new Decoder(
+                projectableDecodingFormat.createRuntimeDecoder(
+                        context,
+                        toPhysicalDataType(
+                                physicalTableDataType, physicalDataTypeProjection, prefix),
+                        deserializerProjectedFields.toArray(
+                                new int[deserializerProjectedFields.size()][])),
+                new ProjectorImpl(
+                        deserializedToProducedPos,
+                        deserializerProjectedFields.size(),
+                        metadataKeys.size()));
+    }
+
+    /**
+     * This method generates a {@link Decoder} for a {@link ProjectableDecodingFormat} that does not
+     * support <em>nested</em> projection. Only the required <em>top-level</em> fields are pushed
+     * down into the format, and any nested sub-fields are extracted from the deserialized row
+     * afterward by the {@link Projector}.
+     */
+    private static Decoder projectTopLevelInsideDeserializerThenNestedAfter(
+            final Context context,
+            final ProjectableDecodingFormat<DeserializationSchema<RowData>>
+                    projectableDecodingFormat,
+            final DataType physicalTableDataType,
+            final int[] physicalDataTypeProjection,
+            final @Nullable String prefix,
+            final int[][] projectedPhysicalFields,
+            final List<String> metadataKeys) {
+        final Map<Integer, Integer> tableToDeserializedTopLevelPos =
+                tableToDeserializedTopLevelPos(physicalDataTypeProjection);
+
+        // Top-level fields (in data type space) to push into the format, deduplicated so each
+        // required top-level field is only deserialized once.
+        final List<int[]> topLevelProjectedFields = new ArrayList<>();
+        final Map<Integer, Integer> dataTypeTopLevelToDeserializedPos = new HashMap<>();
+        final Map<List<Integer>, Integer> deserializedToProducedPos = new HashMap<>();
+        for (int producedPos = 0; producedPos < projectedPhysicalFields.length; producedPos++) {
+            final int[] tablePos = projectedPhysicalFields[producedPos];
+            final int tableTopLevelPos = tablePos[0];
+
+            final Integer dataTypeTopLevelPos =
+                    tableToDeserializedTopLevelPos.get(tableTopLevelPos);
+            if (dataTypeTopLevelPos != null) {
+                final int deserializedTopLevelPos =
+                        dataTypeTopLevelToDeserializedPos.computeIfAbsent(
+                                dataTypeTopLevelPos,
+                                k -> {
+                                    topLevelProjectedFields.add(new int[] {k});
+                                    return topLevelProjectedFields.size() - 1;
+                                });
+
+                // The remaining (nested) path is extracted from the deserialized row afterward.
+                final int[] deserializedPos = copyArray(tablePos);
+                deserializedPos[0] = deserializedTopLevelPos;
+                deserializedToProducedPos.put(
+                        Collections.unmodifiableList(
+                                Arrays.stream(deserializedPos)
+                                        .boxed()
+                                        .collect(Collectors.toList())),
+                        producedPos);
+            }
+        }
+
+        addMetadataProjections(
+                projectableDecodingFormat,
+                topLevelProjectedFields.size(),
+                projectedPhysicalFields.length,
+                metadataKeys,
+                deserializedToProducedPos);
+
+        return new Decoder(
+                projectableDecodingFormat.createRuntimeDecoder(
+                        context,
+                        toPhysicalDataType(
+                                physicalTableDataType, physicalDataTypeProjection, prefix),
+                        topLevelProjectedFields.toArray(new int[topLevelProjectedFields.size()][])),
+                new ProjectorImpl(
+                        deserializedToProducedPos,
+                        topLevelProjectedFields.size(),
+                        metadataKeys.size()));
+    }
+
+    /**
+     * This method generates a {@link Decoder} which deserializes the data fully using the {@link
+     * DecodingFormat} and then applies any projections afterward.
+     */
+    private static Decoder projectAfterDeserializing(
+            final Context context,
+            final DecodingFormat<DeserializationSchema<RowData>> decodingFormat,
+            final DataType physicalTableDataType,
+            final int[] physicalDataTypeProjection,
+            final @Nullable String prefix,
+            final int[][] projectedPhysicalFields,
+            final List<String> metadataKeys) {
+        final DataType physicalDataType =
+                toPhysicalDataType(physicalTableDataType, physicalDataTypeProjection, prefix);
+        final Map<Integer, Integer> tableToDeserializedTopLevelPos =
+                tableToDeserializedTopLevelPos(physicalDataTypeProjection);
+
+        final Map<List<Integer>, Integer> deserializedToProducedPos = new HashMap<>();
+        for (int producedPos = 0; producedPos < projectedPhysicalFields.length; producedPos++) {
+            final int[] tablePos = projectedPhysicalFields[producedPos];
+            int tableTopLevelPos = tablePos[0];
+
+            final Integer deserializedTopLevelPos =
+                    tableToDeserializedTopLevelPos.get(tableTopLevelPos);
+            if (deserializedTopLevelPos != null) {
+                final int[] deserializedPos = copyArray(tablePos);
+                deserializedPos[0] = deserializedTopLevelPos;
+
+                deserializedToProducedPos.put(
+                        Collections.unmodifiableList(
+                                Arrays.stream(deserializedPos)
+                                        .boxed()
+                                        .collect(Collectors.toList())),
+                        producedPos);
+            }
+        }
+
+        addMetadataProjections(
+                decodingFormat,
+                physicalDataTypeProjection.length,
+                projectedPhysicalFields.length,
+                metadataKeys,
+                deserializedToProducedPos);
+
+        return new Decoder(
+                decodingFormat.createRuntimeDecoder(context, physicalDataType),
+                new ProjectorImpl(
+                        deserializedToProducedPos,
+                        physicalDataTypeProjection.length,
+                        metadataKeys.size()));
+    }
+
+    /** Projects fields from the deserialized row to their positions in the final produced row. */
+    @Internal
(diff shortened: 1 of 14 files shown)


flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/config/FormatProjectionPushdownLevel.java
```diff
@@ -0,0 +1,38 @@
...
+public enum FormatProjectionPushdownLevel {
+
+    /** The format does not support any kind of projection pushdown. */
+    NONE,
+
+    /** The format supports projection pushdown for top-level fields only. */
+    TOP_LEVEL,
+
+    /** The format supports projection pushdown for top-level and nested fields. */
+    ALL
```
Savonitar: Do we validate format compatibility with projection pushdown levels? To prevent unsupported combination.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java
```diff
@@ -300,38 +337,42 @@ public Map<String, DataType> listReadableMetadata() {
...
+    @Override
+    public boolean supportsNestedProjection() {
+        return (keyDecodingFormat == null
+                        || keyFormatProjectionPushdownLevel == FormatProjectionPushdownLevel.ALL)
+                && valueFormatProjectionPushdownLevel == FormatProjectionPushdownLevel.ALL;
+    }
+
+    @Override
+    public void applyProjection(final int[][] projectedFields, final DataType producedDataType) {
+        this.projectedPhysicalFields = projectedFields;
```
Savonitar, line 366: Should we also update `this.producedDataType` here?

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java
```diff
@@ -300,38 +337,42 @@ public Map<String, DataType> listReadableMetadata() {
...
         }
-
-        this.metadataKeys = connectorMetadataKeys;
         this.producedDataType = producedDataType;
     }
 
     @Override
     public boolean supportsMetadataProjection() {
-        return false;
+        throw new IllegalStateException(
```
Savonitar: Should this throw an exception instead of returning a boolean? Could you please elaborate on throwing an exception?

E.g. what if this method is called for logging purpose?

I've checked other `flink-connectors` and they either return false or use the default (return true), none throw an exception.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/config/FormatProjectionPushdownLevel.java
```diff
@@ -0,0 +1,38 @@
...
+package org.apache.flink.streaming.connectors.kafka.config;
+
+import org.apache.flink.annotation.Internal;
+
+/**
+ * Projection pushdown mode for {@link
+ * org.apache.flink.streaming.connectors.kafka.table.KafkaDynamicSource}.
+ */
+@Internal
+public enum FormatProjectionPushdownLevel {
```
Savonitar: This enum (`NONE, TOP_LEVEL, ALL)` requires users to understand:
1. Whether the format supports projection pushdown
2. Whether the format supports nested projections
3. Known bugs in specific formats (e.g. as was mentioned in the PR description "Avro FLINK-35324")

Maybe we can replace per-format level configuration with a single boolean flag:
```
'projection-pushdown.enabled' = 'true'
```
and internally format will decide which projection it could use. 

Would this approach address the use cases you had in mind? Or this is added intentionally as a potential workaround for problems with formats?

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaTableSource.java
Savonitar, line 80: Could you please clarify why `DynamicKafkaTableSource` does not implement `SupportsProjectionPushDown`?

Savonitar (commented):
Hi.
Thanks for working on this feature, it is a useful optimization and the implementation looks solid. I've started reviewing the PR and posted some questions/comments.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/config/FormatProjectionPushdownLevel.java
fqshopify: No, we don't validate format compatibility at configuration time. This was intentional. Although I've done some of the work to figure out what level of projection is supported by some formats (`avro`, `csv`, `json`, `avro-confluent`, `debezium-avro-confluent`), the fact is there are more formats out there beyond those maintained by Flink or even in the open source ecosystem (we have some custom formats at my company for example). Even for the formats I've figured out, it's possible the support could potentially vary by version e.g. if the avro format projection pushdown [bugs](https://issues.apache.org/jira/browse/FLINK-35324) are fixed. 

Something we _could_ consider long-term is adding a method to `ProjectableDecodingFormat` to indicate its supported projection level. However, even that wouldn't fully solve the problem e.g. the avro format implements `ProjectableDecodingFormat` but [doesn't actually support all projection pushdowns](https://issues.apache.org/jira/browse/FLINK-35324).

In the short-term, I've chosen to simply document the appropriate combinations in the flink-connector-kafka config docs [here](https://github.com/apache/flink-connector-kafka/blob/1f283ee629c03b8c1a35a7a64217c40fe314a4d1/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaConnectorOptions.java#L116).

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java
fqshopify: Good question! We don't need to update `producedDataType` in `applyProjection` because the javadoc for `applyReadableMetadata` explicitly addresses this scenario:

```java
    /**
     * ...
     *
     * <p>Note: Use the passed data type instead of {@link ResolvedSchema#toPhysicalRowDataType()}
     * for describing the final output data type when creating {@link TypeInformation}. If the
     * source implements {@link SupportsProjectionPushDown}, the projection is already considered in
     * the given output data type, use the {@code producedDataType} provided by this method instead
     * of the {@code producedDataType} provided by {@link
     * SupportsProjectionPushDown#applyProjection(int[][], DataType)}.
     * 
     * ...
     */
    void applyReadableMetadata(List<String> metadataKeys, DataType producedDataType);
```

So we're following the recommended approach i.e. using the `producedDataType` from `applyReadableMetadata` rather than the one from `applyProjection`

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/config/FormatProjectionPushdownLevel.java
fqshopify: > Or this is added intentionally as a potential workaround for problems with formats?

Yep, that's one of the main reasons but also because there are more formats out there that I can't account for. See my response to your first question here for more [details](https://github.com/apache/flink-connector-kafka/pull/174#discussion_r2755948694).

I'm open to discussing this more. This is the only part of this PR I don't like but I felt was a necessary compromise in the short-term.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaTableSource.java
fqshopify: `DynamicKafkaTableSource` is a new source that was just recently added a couple of weeks ago. I'm trying to keep this PR as small as possible so I haven't added `SupportsProjectionPushDown` support here yet but I'd be happy to implement it in the future. 

Let me know if you'd prefer I include it in this PR or follow up separately.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java
fqshopify: From the javadoc for `supportsMetadataProjection`:

```java
    /**
     * ...
     *
     * <p>This method is only called if the source does <em>not</em> implement {@link
     * SupportsProjectionPushDown}.
     *
     * ...
     */
    default boolean supportsMetadataProjection() {
```

Since `KafkaDynamicSource` now implements `SupportsProjectionPushDown`, this method should never be called by the planner. Originally, my reasoning for throwing an exception here was to catch any unexpected invocations that would indicate a bug in the planner.

That said, I take your point about consistency with other connectors and potential issues if it's called for logging/debugging purposes. I've now removed the method entirely (so it will by default return `true`) 👍

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaTableSource.java
Savonitar: I agree with keeping PRs small and focused. A follow-up PR is fine. I think it makes sense to create a follow-up Jira ticket after the approvals/merge so we don't lose it

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/UpsertKafkaDynamicTableFactory.java
Savonitar, line 108: 1. Shouldn't we add `KEY_PROJECTION_PUSHDOWN_LEVEL` and `VALUE_PROJECTION_PUSHDOWN_LEVEL` to `optionalOptions()` method? 
The factory reads these options at line 190-191 (in `createDynamicTableSource`)  and passes them to `KafkaDynamicSource`, but they aren't registered as optional options. 
As a result, when a user sets `key.format-projection-pushdown-level = TOP_LEVEL` on an `upsert-kafka` table, FactoryHelper will throw 
```
ValidationException: Unsupported options found for 'upsert-kafka' 
```
because the key isn't in the consumed set. 

Or am I missing something?

2. Could we add a test that creates an upsert-kafka table with a non-default pushdown level to catch this?

Savonitar (commented):
Thanks for the updates/replies!

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/UpsertKafkaDynamicTableFactory.java
fqshopify: Sorry, this was a miss, great catch! 
I haven't had to use projection-pushdown + upsert yet. 

Added the options and test in the latest commit

Savonitar: Thanks for adding the pushdown options to `optionalOptions()`. 

However, it looks like format-level projection pushdown doesn't work for **upsert-kafka**. Regardless of the configured pushdown level, all value fields are always fully deserialized, and projection happens only afterwards via the Projector. 

The root cause is `DecodingFormatWrapper` -> it only implements DecodingFormat, not ProjectableDecodingFormat. When Decoder.create() checks `decodingFormat instanceof ProjectableDecodingFormat`, the wrapper always fails the check, so the projectInsideDeserializer path is never taken for **upsert-kafka**. 

I implemented a **failing** test to verify this hypothesis and to provide a basis for the fix: https://github.com/Savonitar/flink-connector-kafka/commit/8e2c1aed 

The test writes data with a breaking schema change on a non-projected value field (name changes from INT to STRING), then reads selecting only user_id and payload. With working format-level pushdown, name would be skipped during deserialization and the query would succeed. Instead, it fails with JsonParseException because all fields are deserialized. 

I took into account the style of existing tests in UpsertKafkaTableITCase and your KafkaTableITCase#testProjectionPushdownWithJsonFormatAndBreakingSchemaChange to align. I kept the commit in my fork rather than pushing to your branch -> feel free to cherry-pick it as a regression test, or reimplement it entirely if you prefer. The test should pass when the fix is implemented.

A possible fix: make DecodingFormatWrapper also implement ProjectableDecodingFormat and delegate createRuntimeDecoder(context, producedDataType, projections) to the inner format when it's a ProjectableDecodingFormat.

Could you please take a look and implement the fix if you agree with the analysis?

fqshopify: Thanks for sharing this! And apologies for the delayed response, I've been massively backlogged at work. 

I thought about this and decided to remove projection pushdown support for the upsert-kafka-connector for now in order to keep this PR slim/focused. It requires a few changes (see below) that would be better to address in a separate PR with proper testing. 

> A possible fix: make DecodingFormatWrapper also implement ProjectableDecodingFormat and delegate createRuntimeDecoder(context, producedDataType, projections) to the inner format when it's a ProjectableDecodingFormat.

This will create other issues. Rather than making `DecodingFormatWrapper` implement `ProjectableDecodingFormat`, I'd rather eliminate the wrapper entirely if possible. It exists only to override `getChangelogMode()`, but `KafkaDynamicSource` already has a `upsertMode` boolean so we can override the changelog mode directly in `KafkaDynamicSource.getChangelogMode()` when `upsertMode` is true, and pass the unwrapped format through. That way `Decoder.create()` sees the real format's capabilities (e.g. JSON's `ProjectableDecodingFormat`) without any delegation tricks. We can do all of that in a separate PR once this one goes in.

Savonitar:
Hi. Thanks for keeping at this PR, appreciate your commitment.
I was also backlogged at work as you mentioned in your message :-) and I'm sick right now. To give this a proper review-run, need to feel better first, I expect to continue the review next week. Let me know if you'd like me to ping someone else to review (and feel free to ping someone yourself if you'd prefer not to wait).

In general, the current version after your fixes looks solid. I think the main remaining question is the user-facing configuration design - specifically the implicit compatibility between formats and pushdown levels, and whether the configuration options could be less ambiguous. I'd like to revisit this in the next review round. Apart from that, I don't have concerns.

By the way, do you run this in production already? If you have any benchmarks or observations on the deserialization savings, I'd be interested to see them.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/config/FormatProjectionPushdownLevel.java
Savonitar: I thought about this more and I agree with the three-level approach. Since the feature is disabled by default (`NONE`), the target audience is users who already understand projection pushdown and their format's capabilities. For those users, the distinction between `TOP_LEVEL` and `ALL` provides useful granularity, it allows them to enable top-level pushdown while opting out of nested projections, in case a format has bugs with nested projection despite declaring support via `supportsNestedProjection()`. Missing an optimization is better than wrong results (as the Avro bug demonstrates).

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java
```diff
@@ -300,36 +337,34 @@ public Map<String, DataType> listReadableMetadata() {
...
-        return false;
+    public void applyWatermark(WatermarkStrategy<RowData> watermarkStrategy) {
+        this.watermarkStrategy = watermarkStrategy;
     }
 
     @Override
-    public void applyWatermark(WatermarkStrategy<RowData> watermarkStrategy) {
-        this.watermarkStrategy = watermarkStrategy;
+    public boolean supportsNestedProjection() {
+        return (keyDecodingFormat == null
```
Savonitar, line 360: I have a question. 
This method checks only the user's config (e.g. `FormatProjectionPushdownLevel.ALL`), but never checks the format's own `ProjectableDecodingFormat.supportsNestedProjection()`. 

What if we have the following scenario: 
1. A format that implements ProjectableDecodingFormat with `supportsNestedProjection()` returning **false** (it supports top-level projection but **not nested** , e.g. avro-confluent, or a custom format). 
2. The user sets 'value.format-projection-pushdown-level' = 'ALL' in DDL. 
3. The schema **has nested fields**.

Wouldn't this cause a runtime crash/incorrect results/other issues? 
My understanding of the chain: 
1. Schema has nested ROW fields and query selects a nested subfield
2. Planner calls `KafkaDynamicSource.supportsNestedProjection()`: returns true (based on user config only) 
3. Planner generates nested field paths (e.g., {1, 0}) and passes them to `applyProjection() `
4. At runtime, `Decoder.create()` sees the format instance of `ProjectableDecodingFormat` and pushdown is enabled -> takes the projectInsideDeserializer path ->calls format.createRuntimeDecoder(ctx, type, [[1, 0]]) 
5. The format receives a nested projection (int[] with length > 1) however, format declared it cannot handle  it.

Should `supportsNestedProjection()` also check the format's declared capability?

Savonitar (changes requested):
Hi. I've completed the next round of review. I appreciate your great work, especially the thorough research into which formats support projection pushdown. 
However, I have one question about KafkaDynamicSource.
Could you please take a look?

fqshopify:
> By the way, do you run this in production already? 

Yes! We do run this in production and have been for the last year with multiple pipelines.

> If you have any benchmarks or observations on the deserialization savings, I'd be interested to see them.

I think all of our pipelines that are making use of this feature use the `debezium-avro-confluent` format. The way that format works [internally](https://github.com/apache/flink/blob/d8010df8795e1a8e4895d66b7b4861a85c8e0e24/flink-formats/flink-avro-confluent-registry/src/main/java/org/apache/flink/formats/avro/registry/confluent/debezium/DebeziumAvroDeserializationSchema.java#L102) is it first deserializes the message fully (using the schema-id encoded in the first 4 bytes of the message) before projection pushdown kicks in to select specific columns. So in this case the deserialization savings would be miniscule, if any. The big unlock for us was really the schema-evolution-tolerance.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java
fqshopify: Haven't forgotten about this, still thinking it through!

Debating between two options:

- **Option 1**: Keep the per-format level config, but have `supportsNestedProjection()` also check the format's own `ProjectableDecodingFormat.supportsNestedProjection()` to make sure they don't conflict

- **Option 2** (which you floated [here](https://github.com/apache/flink-connector-kafka/pull/174#discussion_r2746744072) previously): replace the per-format enum with a `'{key/value}.projection-pushdown.enabled' = 'true/false'` flag, and let the connector decide internally based on the's format's `ProjectableDecodingFormat.supportsNestedProjection` which projection level (top-level or nested) it can use. Simpler API and still gives users the ability to turn off projection pushdown if the format has projection pushdown bugs. Trying to remember if there was any specific reason why I didnt do with this in the past or if I simply overlooked the existence of the `ProjectableDecodingFormat.supportsNestedProjection` API.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java
```diff
@@ -72,13 +74,15 @@
...
 import java.util.stream.Stream;
 
 /** A version-agnostic Kafka {@link ScanTableSource}. */
 @Internal
 public class KafkaDynamicSource
-        implements ScanTableSource, SupportsReadingMetadata, SupportsWatermarkPushDown {
+        implements ScanTableSource,
+                SupportsReadingMetadata,
+                SupportsWatermarkPushDown,
+                SupportsProjectionPushDown {
```
Efrat19, line 86: I'm concerned about the default `KafkaDynamicSource` declaring itself `SupportsProjectionPushDown` while not fully supporting it (i.e upsertMode, incompatible format, or simply turned off)

From  [SupportsProjectionPushDown.java](https://github.com/apache/flink/blob/master/flink-table/flink-table-common/src/main/java/org/apache/flink/table/connector/source/abilities/SupportsProjectionPushDown.java#L42C7-L43C31):

> By default, if this interface is not implemented, a projection is applied in a subsequent operation after the source.

Given the feature is opt in, wouldn't it mean that by default workloads are deprived from downstream optimizations?

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaDeserializationSchema.java
```diff
@@ -139,7 +142,7 @@ public void deserialize(ConsumerRecord<byte[], byte[]> record, Collector<RowData
             throws IOException {
         // shortcut in case no output projection is required,
         // also not for a cartesian product with the keys
-        if (keyDeserialization == null && !hasMetadata) {
+        if (keyDeserialization == null && !hasMetadata && !hasValueProjection) {
```
Efrat19, line 145: value projection might be needed even if `valueFormatProjectionPushdownLevel` is set to `NONE` (i.e columns reordering) am I right?
So by default we are presenting another limitation on the shortcut path
Is there any way to avoid it given `deserialize` is the record level hotpath?

Efrat19 (commented):
Hi,
I think this is a most needed contribution, thank you!

Efrat19:
~Also possibly qualifies for a FLIP according to https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=65145551#FlinkImprovementProposals-Whatisconsidereda"majorchange"thatneedsaFLIP?~
(I was concerned about possible downgrade in default behavior but was proven wrong)

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java
fqshopify: The interface is always honored. `applyProjection` updates `producedDataType` and the projection is applied at runtime by the post-deserialization `Projector` regardless of `*_PROJECTION_PUSHDOWN_LEVEL`. 

The option only controls whether projection is *additionally* pushed into the format (when it implements `ProjectableDecodingFormat`):
- `level=FULL` + `ProjectableDecodingFormat` => projection happens inside the deserializer (skip unneeded bytes).
- Otherwise (level=`NONE`, non-projectable format, upsert mode) => format deserializes the full physical row, then `Projector` trims it down before emit.

Either way the source emits the projected row, so downstream operators see exactly what they would with a post-source Calc - no optimizations are lost.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaDeserializationSchema.java
fqshopify: > value projection might be needed even if valueFormatProjectionPushdownLevel is set to NONE (i.e columns reordering)

Some clarifications: 
- "projection" in Flink covers both narrowing and/or reordering of fields.
- This is unrelated to `valueFormatProjectionPushdownLevel` (as covered in my earlier reply): that option only controls whether projection is pushed *all the way down into the format*. The narrowing/reordering case above can occur regardless of the level (even with `NONE`). 

> So by default we are presenting another limitation on the shortcut path

The `&& !hasValueProjection` guard was effectively always there, just implicit. Before this PR, the value deserializer was always built from a `physicalDataType` that matched the produced row's value portion exactly. So when there was no key and no metadata, the deserialized row *was* the produced row, and emitting it directly was always safe.

With projection pushdown that's no longer guaranteed. Flink can now call `applyProjection` and ask for a subset/reordering of the value fields. So the deserialized row can legitimately differ from the produced row even when no key and no metadata are involved — hence the new explicit guard.

> Is there any way to avoid it given deserialize is the record level hotpath?

The shortcut will still be taken in the cases it was before. The shortcut can now _also_ potentially be taken in cases where projections are pushed down into the format and it returns exactly what downstream operators expect. There's a bunch of unit tests around this in `DecoderTest`.

Efrat19: Understood, thanks for clarifying.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java
Savonitar: Hi @fqshopify . 
Thanks for laying out both options clearly, and sorry for the late reply. I think we're pretty close to finalizing this work and it would be a shame to abandon it (I hope you still have capacity). 

My vote is Option 1:  keep the three-level enum and have `supportsNestedProjection()` also check the format's own `ProjectableDecodingFormat.supportsNestedProjection()`. The deciding factor for me is the "lying format" case, which is the same reason we kept the three-level enum in the first place. 
As a result, with Option 1 the user's level becomes a "ceiling" rather than a command: the planner is told **true** only when the user opted into **ALL** and the format actually declares nested support. 

I prototyped this locally with regression tests and it works cleanly (cases that fail on the current code pass with the fix, and the full suite stays green). 
I'm not against Option 2, but it would likely require more changes to the PR, whereas the current approach has already been reviewed/discussed/agreed.

WDYT?

fqshopify: > I think we're pretty close to finalizing this work 

I agree, and thank for you attention to this PR, it is much appreciated!

> it would be a shame to abandon it (I hope you still have capacity).

Ha, definitely can't abandon this, we're still using it internally XD I should have some capacity next week to continue with this next week, let me try to resolve this comment then. 

> I'm not against Option 2, but it would likely require more changes to the PR, whereas the current approach has already been reviewed/discussed/agreed.

TBH my current thinking is that, unless I can come up with a concrete objection, I'd actually prefer Option 2. I've prototyped this locally but I want to do some more testing internally.

fqshopify: @Savonitar Alright, I went with Option 2 (see updated PR description + last two commits). Summary of how it resolves your original concern:
- The config is now a simple boolean per key/value: `{key,value}.format.projection-pushdown.enabled` (default `false`). The old `NONE/TOP_LEVEL/ALL` enum is gone, the user no longer determines this. 
- Instead, when pushdown is enabled, the connector decides the actual level internally based whether the format is instance of `ProjectableDecodingFormat` and if `ProjectableDecodingFormat.supportsNestedProjection()`.
- For the "top-level only but the schema has nested fields" case you raised, there's a new path (`projectTopLevelInsideDeserializerThenNestedAfter`): it pushes only the top-level fields into the format and extracts the nested sub-fields from the deserialized row afterward. So no nested path is ever handed to a format that can't take one.

Overall, I think this is a win-win; it's both simpler for the user and also more correct.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaConnectorOptions.java
```diff
@@ -110,6 +110,25 @@ public class KafkaConnectorOptions {
     public static final ConfigOption<Integer> SCAN_PARALLELISM = FactoryUtil.SOURCE_PARALLELISM;
     public static final ConfigOption<Integer> SINK_PARALLELISM = FactoryUtil.SINK_PARALLELISM;
 
+    private static final String PROJECTION_PUSHDOWN_DESCRIPTION =
```
Savonitar, line 113: I think we need to document this option in docs ([Connector Options](https://github.com/apache/flink-connector-kafka/pull/174#discussion_r2755948694)).

Savonitar (approved):
Thanks for seeing this through. The previous issues and concerns have been addressed. I left one comment about the docs update. Let me know if you don’t have the capacity to update them.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaConnectorOptions.java
fqshopify: Done the english side! (not sure how the chinese side is handled)

Also rebased on top of latest main

Savonitar (commented):
Frequently both versions of the docs (English and Chinese) are committed together, it's in Flink's docs contribution guide (https://flink.apache.org/how-to-contribute/contribute-documentation/#chinese-documentation-translation). 

In theory you have at least the following options:
a) use AI to translate (ideally verified by a native speaker before committing)
b) copy the English text into `content.zh/.../kafka.md` as a placeholder and file a JIRA under the `chinese-translation` component, linked to FLINK-32609 (see the existing ones here https://issues.apache.org/jira/browse/FLINK-39510?jql=project%20%3D%20FLINK%20AND%20component%20%3D%20chinese-translation)
c) if you'd rather not do a) or b), I can commit an AI translation verified with a Chinese speaker to help.

Could you also squash the commits before we merge? The branch currently has 5 commits and only the first carries the [FLINK-32609] tag, the rest are review-iteration commits. 
By convention every commit on main is tagged `[FLINK-XXXXX][component]`, so collapsing to a single tagged commit keeps history clean regardless of how it's merged. Maybe you can use `[FLINK-32609][Connectors/Kafka] Support projection pushdown`

If you're pushing again for the docs anyway, you could fold both into the same force-push.

fqshopify:
> use AI to translate (ideally verified by a native speaker before committing) 

Done and verified with colleagues!

> Could you also squash the commits before we merge? 

Done!

Savonitar (approved):
LGTM. 
I'm not a committer. I will ask someone to run CI.

Efrat19 (approved):
I also went through the change and have no further concerns.

fqshopify:
@Savonitar Looks like all checks passed 🥳 

Small request: Can you make sure to preserve the `Co-authored-by: fqaiser94 ...` when you merge the commit please? 🥺 Trying to make sure it gets associated with my main GitHub account (@fqaiser94).

Savonitar:
> @Savonitar Looks like all checks passed 🥳

Yes, I asked folks to run CI previous week. 

> 
> Small request: Can you make sure to preserve the `Co-authored-by: fqaiser94 ...` when you merge the commit please? 🥺 Trying to make sure it gets associated with my main GitHub account (@fqaiser94).

I'm not a committer 😞  But I had asked PMC to check this PR. They will check it soon. I will keep my eyes on it.

Savonitar:
Hi, JFYI, I haven’t forgotten about this PR and have gently reminded the relevant committers and PMC members about it.

fqaiser94:
Thanks for staying on top of this @Savonitar! 

I'll rebase on top of latest main and address the conflicts as well later today. 

---

Ah just FYI; I retired my @fqshopify GitHub acocunt and am now using @fqaiser94

Savonitar:
cc: @twalthr

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaDeserializationSchema.java
```diff
@@ -264,7 +267,7 @@ private static final class OutputProjectionCollector
         @Override
         public void collect(RowData physicalValueRow) {
             // no key defined
-            if (keyProjection.length == 0) {
+            if (keyProjector.isEmptyProjection()) {
```
MartijnVisser, line 270: This used to mean the table has no key, it now also means the query selects no key column. A record with a null Kafka key is then emitted rather than skipped, so `SELECT c` returns a row that `SELECT a, c` does not. Can it stay on `keyDeserialization == null`?

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/Decoder.java
```diff
@@ -0,0 +1,471 @@
...
+            return deserializedToProducedPos.isEmpty();
+        }
+
+        @Override
+        public boolean isProjectionNeeded() {
+            return isProjectionNeeded;
+        }
+
+        @Override
+        public void project(final RowData deserialized, final GenericRowData producedRow) {
```
MartijnVisser, line 457: This runs per record for every table with a key or metadata, and by default for any narrowing query. The map walk allocates 24 to 160 bytes a record where the `int[]` loop it replaces allocated none, a flat `int[]` path for the top level entries avoids that.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java
```diff
@@ -74,13 +75,15 @@
 import java.util.Properties;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
```
MartijnVisser, line 77: `mvn spotless:check` fails on this import and on the same one in DynamicKafkaTableSource, so CI stops before any test runs. Both went unused when the rebase took main's deletion of `getResetStrategy`.

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableITCase.java
```diff
@@ -1721,6 +1750,504 @@ private void testStartFromGroupOffsetsWithNoneResetStrategy(final String format)
...
+                        + "Calc(select=[a, b, topic, c, partition, d])\n"
+                        + "+- TableSourceScan(table=[[default_catalog, default_database, kafka, metadata=[topic, partition]]], fields=[a, b, c, d, topic, partition])\n",
+                Collections.singletonList(String.format("+I(a,b,%s,c,%d,d)", topic, 0)));
+
+        cleanupTopic(topic);
+    }
+
+    @ParameterizedTest(name = "format: {0}")
+    @MethodSource("formats")
+    public void testProjectionPushdownSelectSpecificPhysicalFields(final String format)
```
MartijnVisser, line 1809: None of these selects only value columns, so the branch is never reached with a key format present, the case where it differs from before. I think a case for that, plus one for a null key, would cover it.

MartijnVisser (changes requested):
Thanks for sticking with this one for so long.

One behaviour change blocks it for me, with both new options off: a record with a null Kafka key is no longer skipped when the query selects no key column, so the row count depends on the projection. Comment inline. The build is also red on spotless.

Worth stating in the description and in a release note: every kafka and upsert-kafka table now advertises projection pushdown, so any query reading a subset of the columns gets a different plan. Where that drops the Calc above the source, every operator below it shifts one id, so a job with state downstream will not restore from its savepoint. The source keeps its id. Declared metadata that is not selected is no longer read either.

CI has not run on this head, the green run was on the pre-rebase commit.

PR #175, opened 2025-05-20
[FLINK-37818] Fix Kafka connector requires unqiue TransactionalIdPrefix even for non-EOS settings.

AHeise:


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/NoopCommitter.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * The committer to be used for non exactly-once delivery guarantees.
+ *
+ * <p>This committer does not commit any records. It is needed because the current {@link
+ * org.apache.flink.api.connector.sink2.Sink} design supports only either transactional or
+ * non-transactional operation and the {@link org.apache.flink.connector.kafka.sink.KafkaSink} is
+ * doing both through {@link org.apache.flink.connector.base.DeliveryGuarantee}s.
+ */
+@Internal
+public class NoopCommitter implements Committer<KafkaCommittable> {
```
davidradl, line 36: nit: I wonder if NoopCommitter naming could be changed - as the the implementation is not a no-op it throws errors in the 2 methods. Maybe AtLeastOnceCommitter and the KafkaCommitter could be renamed ExactlyOnceCommitter.

davidradl (approved):

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/NoopCommitter.java
AHeise: The methods had the exceptions only in the signature. I removed them to avoid confusion. So now it's a true no-op committer.

AHeise:
Verified to work by the original reporter of the issue.

PR #181, opened 2025-07-08
[FLINK-38063] Update dependencies

tomncooper:
While working on updating the kafka client libraries, I noticed that many of the dependencies for the connector are out of date and/or do not match what the main Flink project is using. 

This PR updates dependencies to their latest (compatible) version. If any update required code changes I left it out and will deal with it in a separate future PR.

The PR also moves all dependency and plugin versions in to the properties section so it is easier to manage them.

### DIFF
--- a/flink-connector-kafka/pom.xml
+++ b/flink-connector-kafka/pom.xml
@@ -239,7 +239,6 @@ under the License.
 		<dependency>
 			<groupId>org.yaml</groupId>
 			<artifactId>snakeyaml</artifactId>
-			<version>2.2</version>
 			<scope>test</scope>
 		</dependency>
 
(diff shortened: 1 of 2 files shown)


flink-connector-kafka/pom.xml
```diff
@@ -239,7 +239,6 @@ under the License.
 		<dependency>
 			<groupId>org.yaml</groupId>
 			<artifactId>snakeyaml</artifactId>
```
fapaul, line 241: Nit: Can we delete this dependency block completely? To me, it looks like snakeyaml is already on the classpath (by some transitive dependency) and we are adding it here unnecessarily.

PR #184, opened 2025-07-30
[hotfix] Set version to 4.0-SNAPSHOT

fapaul:
During the release, we accidentally overwrote the snapshot version on the branch.  The fixed versions are only meant to be used on release tags but not on the active branches.


AHeise (approved):
LGTM.

PR #189, opened 2025-09-01
Kafka 4.x Queue Semantics support

Shekharrajak:
Ref https://issues.apache.org/jira/browse/FLINK-38287

This implementation adds Kafka 4.x share group semantics to Flink's Kafka connector while maintaining full backward compatibility with existing code. The code changes are following KIP-932 and FLIP-27 main source architecture and implicit mode acknowledgement.

This directly addresses use cases where:

1. Multiple consumers need to process items efficiently in parallel from a single/multiple topic(s).
2. Messages  need explicit acknowledgment/release (to avoid reprocessing or allow retries).
Use cases where scaling Flink ML/LLM workload is critical - Shifting Kafka coordination and assignment logic to the broker side would simplify today’s complex Flink source management, making consumption more efficient, scalable, and far less error-prone.
  Operational Benefits

  - Higher Throughput: ShareGroupHeartbeat helps in Queue-like workloads, maximum throughput scenarios. Share groups distribute messages at the record level, not partition level, so multiple readers can consume from the same topic with Kafka coordinating message distribution.
  - Better Availability and Flexible Scaling: consumers assignment logic is simpler in server side and rebalancing frequency is minimised.

**Note:**

**Semantic Guarantee**
  **AT-MOST-ONCE**: Records acknowledged immediately after polling. Job failure after acknowledgment but
  before checkpoint = permanent data loss.

Ref https://issues.apache.org/jira/browse/KAFKA-19883 implemented for no data loss in commit : https://github.com/apache/flink-connector-kafka/pull/189/commits/ef83f90ca8ea1095db0883b2cc269731de0558f0

  **Good for**: Logs, monitoring, non-critical analyticsNot for: Financial transactions, critical data,
  audit trails


flink-connector-kafka/pom.xml
```diff
@@ -51,7 +53,7 @@ under the License.
 		<dependency>
 			<groupId>org.apache.flink</groupId>
 			<artifactId>flink-streaming-java</artifactId>
-			<scope>provided</scope>
+			<scope>compile</scope>
```
Shekharrajak: These changes can be reverted.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaShareGroupSourceBuilderTest.java
```diff
@@ -0,0 +1,487 @@
...
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+/**
+ * Comprehensive test suite for {@link KafkaShareGroupSourceBuilder}.
+ * 
+ * <p>This test validates builder functionality, error handling, and property management
+ * for Kafka share group source construction.
+ */
+@DisplayName("KafkaShareGroupSourceBuilder Tests")
```
Shekharrajak, line 46: Testcases improvements is required, I will check and update them accordingly.

pom.xml
```diff
@@ -54,7 +54,7 @@ under the License.
 		<!-- Main Dependencies -->
 		<confluent.version>7.9.2</confluent.version>
 		<flink.version>2.0.0</flink.version>
-		<kafka.version>3.9.1</kafka.version>
+		<kafka.version>4.1.0</kafka.version>
```
Shekharrajak, line 57: This is not yet released and expected to be available to download. Meantime the testing is done by adding this into class path.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/ShareGroupBatchManager.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Manages batches of records from Kafka share consumer for checkpoint persistence.
+ * Controls when new batches can be fetched to work within share consumer's auto-commit constraints.
+ */
+public class ShareGroupBatchManager<K, V>
```
Shekharrajak: ListCheckpointed will help to store the records which is polled but not yet processed in Flink persistent checkpoint state - this will make sure in case of any failure / crash we process the records that we read & ack.

Shekharrajak:
Using Flink SQL, we can have some validation : 

```
CREATE TABLE kafka_share_source (
      message STRING
  ) WITH (
      'connector' = 'kafka-sharegroup',
      'bootstrap.servers' = 'localhost:9092',
      'share-group-id' = 'flink-sql-test-group',
      'topic' = 'test-topic',
      'format' = 'raw',
      'source.parallelism' = '4'  -- 4 subtasks regardless of partition count
  );
 
select * from kafka_share_source;

```

dafd0c9c-ec40-4af2-a929-ca27007076d8

pom.xml
```diff
@@ -53,8 +53,8 @@ under the License.
 
 		<!-- Main Dependencies -->
 		<confluent.version>7.9.2</confluent.version>
-		<flink.version>2.1.0</flink.version>
-		<kafka.version>4.0.0</kafka.version>
+		<flink.version>2.0.0</flink.version>
+		<kafka.version>4.1.0</kafka.version>
```
davidradl, line 57: [PR 190](https://github.com/apache/flink-connector-kafka/pull/190) is also upping the Kafka client level to 4.1. If we do it here then we should amend the NOTOCE as per pr 190. fyi @tomncooper

jnh5y:
As a high-level note, since share groups do not use transactions, there will be some possibility for reprocessing messages.  Is that ok for your use cases?

Generally, do you have any performance numbers to show that this consumer is faster?  (Of course, since transactions are not available, I could imagine it being a little bit faster anyhow...)

pom.xml
huangmingyuan: when will this be released?

Shekharrajak:
This PR served as the initial prototype that validated the core ideas behind share group integration with Flink. The code changes here proved the feasibility of checkpoint-aligned acknowledgment for at-least-once delivery with Kafka's queue semantics. I presented this work at Community Over Code and PyData Global.

Since then, the design has matured significantly. A comprehensive proposal with exhaustive failure analysis, a simplified pause/resume checkpoint model, and zero Flink side record state tracking is now documented here: https://cwiki.apache.org/confluence/x/1IEmGQ

PR #191, opened 2025-09-29
[FLINK-38451] Stop logging interrupts as ERROR

AHeise:
When cancelling a job, we may interrupt the KafkaCommitter. This currently leads to ERROR log "Transaction ... encountered error and data has been potentially lost."

However, that exception is expected and not leading to any data loss beyond the normal inconsistencies because of cancellation. In many cases, the commit already succeeded. Further, a job restart will lead to data being eventually committed.

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitter.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitter.java
@@ -1,12 +1,13 @@
 /*
- * Licensed to the Apache Software Foundation (ASF) under one or more
- * contributor license agreements.  See the NOTICE file distributed with
- * this work for additional information regarding copyright ownership.
- * The ASF licenses this file to You under the Apache License, Version 2.0
- * (the "License"); you may not use this file except in compliance with
- * the License.  You may obtain a copy of the License at
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
  *
- *    http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -26,6 +27,7 @@ import org.apache.flink.util.ExceptionUtils;
 import org.apache.flink.util.IOUtils;
 
 import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.common.errors.InterruptException;
 import org.apache.kafka.common.errors.InvalidTxnStateException;
 import org.apache.kafka.common.errors.ProducerFencedException;
 import org.apache.kafka.common.errors.RetriableException;
@@ -125,6 +127,17 @@ public class KafkaCommitter implements Committer<KafkaCommittable>, Closeable {
                         e);
                 handleFailedTransaction(producer);
                 request.signalFailedWithKnownReason(e);
+            } catch (InterruptException e) {
+                // note that we do not attempt to recover from this exception; producer is likely
+                // left in an inconsistent state
+                LOG.info(
+                        "Committing transaction ({}) was interrupted. This most likely happens because the task is being cancelled.",
+                        request,
+                        e);
+                // reset the interrupt flag that is set when InterruptException is created
+                Thread.interrupted();
+                // propagate interruption through java.lang.InterruptedException instead
+                throw new InterruptedException(e.getMessage());
             } catch (Exception e) {
                 LOG.error(
                         "Transaction ({}) encountered error and data has been potentially lost.",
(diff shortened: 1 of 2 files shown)


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitter.java
```diff
@@ -125,6 +127,17 @@ public void commit(Collection<CommitRequest<KafkaCommittable>> requests)
...
                 request.signalFailedWithKnownReason(e);
+            } catch (InterruptException e) {
+                // note that we do not attempt to recover from this exception; producer is likely
+                // left in an inconsistent state
+                LOG.info(
+                        "Committing transaction ({}) was interrupted. This most likely happens because the task is being cancelled.",
+                        request,
+                        e);
+                // reset interrupt flag that is when the Kafka exception is created
+                Thread.interrupted();
```
fapaul: Nit: Since we rethrow the interrupted exception, it's not necessary to mark the thread interrupted again, isn't it?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitter.java
```diff
@@ -179,11 +192,16 @@ private void handleFailedTransaction(FlinkKafkaInternalProducer<?, ?> producer)
         closeCommitterProducer(producer);
     }
 
-    private void closeCommitterProducer(FlinkKafkaInternalProducer<?, ?> producer) {
-        if (producer == this.committingProducer) {
+    private boolean closeCommitterProducer(FlinkKafkaInternalProducer<?, ?> producer) {
```
fapaul: This change looks unrelated to the fix on this PR. Can you explain why it's necessary afaict the return type is also ignored always?

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitterTest.java
```diff
@@ -158,6 +169,59 @@ public void testCommitterProducerClosedOnError() throws IOException, Interrupted
...
     }
 
+    @Test
+    public void testInterrupt() throws IOException {
+        ServerSocket serverSocket = new ServerSocket(0);
+        Properties properties = getProperties();
+        properties.put(
+                CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG,
+                "http://localhost:" + serverSocket.getLocalPort());
+        properties.put(ProducerConfig.REQUEST_TIMEOUT_MS_CONFIG, "1000");
```
fapaul, line 179: Nit: Let's make the timeout higher to `10000` to reduce the likelihood of flakiness

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitterTest.java
```diff
@@ -158,6 +169,59 @@ public void testCommitterProducerClosedOnError() throws IOException, Interrupted
...
+            assertThat(backchannel).doesNotHave(transactionFinished(true));
+        }
+    }
+
+    private AtomicBoolean interruptOnMessage(Thread mainThread, ServerSocket serverSocket) {
+        final AtomicBoolean interrupting = new AtomicBoolean();
+        new Thread(
+                        () -> {
+                            try {
+                                serverSocket.accept().getInputStream().read();
```
fapaul, line 213: Can't this also trigger by some kind of metadata request that the internal kafkaProducer does? I want to avoid interrupting before we actually call commit.

fapaul (commented):
Looks mostly good left some minor comments.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitter.java
AHeise: Yes, you should not even do it. Hence I reset the flag here. (This is not Thread.interrupt()).

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitter.java
AHeise: I'm reverting. Was part of an earlier attempt that tries to maintain a resumable state.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitterTest.java
AHeise: By default, we are retrying MAX_VALUE times. A longer timeout just unnecessarily prolongs test execution (the request needs to time out).

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitterTest.java
AHeise: We are only listening right before commitTransaction. So any other producer initialization would be done already. However, since we use resumeTransaction with reflection hacks, we actually don't communicate with the broker at all (it's not available for this test by design).

Note that commitTransaction actually issues two requests (one is a metadata, one the actual commit). I haven't found a way or a reason to only wait on the second without relying on internals. WDYT?

fapaul (approved):
LGTM

PR #192, opened 2025-09-30
[FLINK-38453] Add full splits to KafkaSourceEnumState

AHeise:
KafkaEnumerator's state contains the TopicPartitions only but not the offsets, so it doesn't contain the full split state contrary to the design intent.

There are a couple of issues with that approach. It implicitly assumes that splits are fully assigned to readers before the first checkpoint. Else the enumerator will invoke the offset initializer again on recovery from such a checkpoint leading to inconsistencies (LATEST may be initialized during the first attempt for some partitions and initialized during second attempt for others).

Through addSplitBack callback, you may also get these scenarios later for BATCH which actually leads to duplicate rows (in case of EARLIEST or SPECIFIC-OFFSETS) or data loss (in case of LATEST). Finally, it's not possible to safely use KafkaSource as part of a HybridSource because the offset initializer cannot even be recreated on recovery.

All cases are solved by also retaining the offset in the enumerator state. To that end, this commit merges the async discovery phases to immediately initialize the splits from the partitions. Any subsequent checkpoint will contain the proper start offset.


flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumStateSerializerTest.java
```diff
@@ -56,73 +58,108 @@ public void testEnumStateSerde() throws IOException {
...
+        Set<KafkaPartitionSplit> topicPartitions = new HashSet<>();
         for (int readerId = 0; readerId < NUM_READERS; readerId++) {
             for (int partition = startPartition;
                     partition < startPartition + NUM_PARTITIONS_PER_TOPIC;
                     partition++) {
-                topicPartitions.add(new TopicPartition(TOPIC_PREFIX + readerId, partition));
+                topicPartitions.add(
+                        new KafkaPartitionSplit(
+                                new TopicPartition(TOPIC_PREFIX + readerId, partition),
+                                STARTING_OFFSET));
```
Savonitar: Thanks for the PR. This is a very good improvement for the connector.
I noticed that the current test creates splits using the constant `KafkaPartitionSplit.EARLIEST_OFFSET`, would it make sense to add a test case that uses a real-world offset (e.g., `123`)?

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumeratorTest.java
```diff
@@ -316,14 +329,20 @@ public void testDiscoverPartitionsPeriodically() throws Throwable {
         }
     }
 
-    @Test
-    public void testAddSplitsBack() throws Throwable {
+    @ParameterizedTest
+    @EnumSource(StandardOffsetsInitializer.class)
+    public void testAddSplitsBack(StandardOffsetsInitializer offsetsInitializer) throws Throwable {
```
Savonitar, line 403: Is my understanding correct that the test verifies that the offset is correctly recalculated on recovery, but doesn't verify that the original offset(before the failure) was preserved and restored?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumStateSerializer.java
```diff
@@ -132,22 +197,43 @@ private static KafkaSourceEnumState deserializeAssignedTopicPartitions(
...
                 final String topic = in.readUTF();
                 final int partition = in.readInt();
                 final int statusCode = in.readInt();
                 partitions.add(
-                        new TopicPartitionAndAssignmentStatus(
-                                new TopicPartition(topic, partition),
+                        new SplitAndAssignmentStatus(
+                                new KafkaPartitionSplit(
+                                        new TopicPartition(topic, partition),
+                                        DEFAULT_STARTING_OFFSET),
```
fapaul: Isn't this a behavioral change? Previously the unassigned split would get the starting offset configured by the user on reassignment.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
```diff
@@ -96,7 +96,7 @@ public class KafkaSourceEnumerator
     // initializing partition discovery has finished.
     private boolean noMoreNewPartitionSplits = false;
     // this flag will be marked as true if initial partitions are discovered after enumerator starts
-    private boolean initialDiscoveryFinished;
+    private volatile boolean initialDiscoveryFinished;
```
fapaul, line 102: Can you explain why volatile is needed here? Afaik we don't access it from a different thread than before.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
```diff
@@ -229,38 +239,16 @@ public void close() {
...
-     * Check if there's any partition changes within subscribed topic partitions fetched by worker
-     * thread, and invoke {@link KafkaSourceEnumerator#initializePartitionSplits(PartitionChange)}
-     * in worker thread to initialize splits for new partitions.
-     *
-     * <p>NOTE: This method should only be invoked in the coordinator executor thread.
-     *
-     * @param fetchedPartitions Map from topic name to its description
-     * @param t Exception in worker thread
-     */
-    private void checkPartitionChanges(Set<TopicPartition> fetchedPartitions, Throwable t) {
```
fapaul, line 246: Nit: It would have been good to have a small refactoring commit to simplify the method calls it makes the reviewer easier :)

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumeratorTest.java
```diff
@@ -662,63 +688,84 @@ private void verifyLastReadersAssignments(
...
-                (reader, splits) -> {
-                    Set<TopicPartition> expectedAssignmentsForReader =
-                            expectedAssignments.get(reader);
-                    assertThat(expectedAssignmentsForReader).isNotNull();
-                    assertThat(splits.size()).isEqualTo(expectedAssignmentsForReader.size());
-                    for (KafkaPartitionSplit split : splits) {
-                        assertThat(expectedAssignmentsForReader)
-                                .contains(split.getTopicPartition());
+        assertThat(actualAssignments).containsOnlyKeys(expectedAssignments.keySet());
+        SoftAssertions.assertSoftly(
```
fapaul, line 817: I am not a big fan of soft assertions but this pattern with the lambda I think is "okay". It's just a dangerous precedent when folks start using the the version that requires explicitly calling `assertAll()` and we silently  disable all exceptions.

fapaul (commented):
Looks mostly good left some inline comments

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumStateSerializer.java
AHeise: Yes, added a new MIGRATED offset to indicate that this needs to be initialized on recovery.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumStateSerializerTest.java
AHeise: I had to change the logic a bit and introduced a new special value MIGRATED against which all unit tests now go. However, I also added a test with specific offsets to `KafkaSourceEnumeratorTest`.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumeratorTest.java
AHeise: Good catch. I expanded the test to cover snapshotting.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
AHeise: I added a comment explaining the access pattern. The short answer is that I haven't found a way to inject the current state of variable into the worker thread. I could schedule the first discovery differently with `context.callAsync(() -> discover(true));` and then schedule periodic discovery differently but that's a bigger change.
Since it's not on the hot path whatsoever, I'd leave it on a better documented volatile.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
AHeise: in hindsight, yes.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumeratorTest.java
AHeise: Here I use soft assertions for assertion on the nested structure of the same data structure. I haven't found a good way to have all assertions on this one data structure in one line.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
```diff
@@ -131,19 +134,84 @@ public KafkaSourceEnumerator(
...
+        if (migratedPartitions.isEmpty()) {
+            return splitByAssignmentStatus(splits.stream());
+        }
+
+        final Map<TopicPartition, Long> startOffsets =
+                startingOffsetInitializer.getPartitionOffsets(
+                        migratedPartitions, getOffsetsRetriever());
+        return splitByAssignmentStatus(
+                splits.stream()
+                        .map(splitStatus -> resolveMigratedSplit(splitStatus, startOffsets)));
```
fapaul, line 179: Nit: The flow of extracting the migratedPartitions is overly complex because we extract the migrated partitions twice in line 179 and line 161.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumeratorTest.java
```diff
@@ -316,39 +341,67 @@ public void testDiscoverPartitionsPeriodically() throws Throwable {
         }
     }
 
-    @Test
-    public void testAddSplitsBack() throws Throwable {
+    @ParameterizedTest
+    @EnumSource(StandardOffsetsInitializer.class)
+    public void testAddSplitsBack(StandardOffsetsInitializer offsetsInitializer) throws Throwable {
```
fapaul, line 403: Can you also add a test to cover the newly added migration story?

fapaul (commented):
Thanks for addressing the comments 👍  I am only missing a higher level test for the newly added offset migration in the enumerator

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
AHeise: It's unfortunately necessary by design:
* 161 extracts the partitions which are used to jointly look up the partition offsets 
* This is expensive as it uses admin client to contact Kafka cluster
* The design of offset initializer is to jointly look up all partitions to have 1 request to Kafka brokers only
* Now that we received all offsets, 179 is applying them to the split. It could be a simple map lookup but I decided to add some assertion, so it went into a different method.

Savonitar:
LGTM

PR #245, opened 2026-04-21
[FLINK-39514] Release dynamic Kafka split outputs on metadata shrink

lnbest0707:
## Summary

- Release Dynamic Kafka Source split-local outputs when metadata refresh removes assigned splits.
- Defer releases until the runtime `ReaderOutput` exists for restored/pending splits filtered during startup.
- Add reader tests for both assigned split removal and restored pending split cleanup.

## Root Cause

DynamicKafkaSourceReader filtered removed cluster/topic splits out of reader state and closed sub-readers, but did not call `ReaderOutput.releaseOutputForSplit(...)`. Flink keeps split-local outputs registered for watermark aggregation until they are explicitly released, so removed Kafka splits could continue holding back the source watermark indefinitely. ([Flink code link](https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/api/connector/source/ReaderOutput.java#L103))


bowenli86:
lgtm, merging

PR #246, opened 2026-05-05
[FLINK-39608] Change KafkaDatasetFacet to extend DatasetConfigFacet to fix classloader issues

swapna267:
Currently OpenLineage event emitter relies on availability of custom Kafka facet in the classpath.
https://github.com/OpenLineage/OpenLineage/blob/main/integration/flink/flink2/src/main/java/io/openlineage/flink/util/KafkaDatasetFacetUtil.java#L25

This fails when Kafka connector is being loaded as part of user jar and openlineage library loaded as part of Flink distribution. 
Openlineage emitters gets loaded in ApplicationClassLoader , whereas the Kafka connector 
libs gets loaded into FlinkUserClassLoader.

In order to fix this, we make Kafka connector emit facet that is an instance of `org.apache.flink.streaming.api.lineage.DatasetConfigFacet`. Openlineage libs can then rely on the existence of facet by name and pull configs by casting to DatasetConfigFacet. This will avoid depending on Kafka connector classes to be in same class loader as lineage emitter libs.

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/DefaultKafkaDatasetFacet.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/lineage/DefaultKafkaDatasetFacet.java
@@ -22,6 +22,9 @@ package org.apache.flink.connector.kafka.lineage;
 import org.apache.flink.annotation.PublicEvolving;
 import org.apache.flink.connector.kafka.source.KafkaPropertiesUtil;
 
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


spuru9 (approved):
LGTM

mbalassi (approved):

swapna267:
I see Title Validator Failed because [FLINK-39608] was missing before. Is there a way to retrigger that ?

mbalassi:
@swapna267 the title validator was failing on the title of the commit, I fixed it during merging: https://github.com/apache/flink-connector-kafka/commit/fa1021ee09c2594a3087be7983f864574539015c.

swapna267:
Thanks @mbalassi

PR #247, opened 2026-05-08
[FLINK-38790] Update CI to Flink 2.2, adjust code

ferenc-csaky:
With Flink 2.2 `ExactlyOnceKafkaWriter.close()` aborted the transaction before the writer was marked closed, so `TransactionAbortedExceptions` from in-flight sends were recorded as real async producer failures and rethrown during cleanup. This could kill the MiniCluster TaskManager and leave the job restarting without slots.

The applied fix marks the writer closed before `closeAll` gets called. Shutdown related callbacks get ignored, but any prev async exceptions still surface properly. Added a regression test that validates the changed logic only suppresses shutdown-induced abort errors, not real async producer failures recorded before close, so the logic flow stays the same.

---
Also updated the `main` branch version to `SNAPSHOT` in `weekly.yml`, cause `main` refers to the unreleased latest state of the connector, so IMO it makes more sense to test against the latest unreleased core version state we plan to support with the upcoming connector relese.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -33,7 +33,7 @@ jobs:
           flink: 2.2-SNAPSHOT,
           branch: main
         }, {
-          flink: 2.1.1,
+          flink: 2.1-SNAPSHOT,
           branch: main
         }, {
           flink: 2.0.1,
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
@@ -268,6 +268,7 @@ class ExactlyOnceKafkaWriter<IN> extends KafkaWriter<IN> {
 
     @Override
     public void close() throws Exception {
+        markClosed();
         closeAll(
                 this::abortCurrentProducer,
                 () -> closeAll(producerPool),
--- a/.github/workflows/push_pr.yml
+++ b/.github/workflows/push_pr.yml
@@ -28,7 +28,7 @@ jobs:
   compile_and_test:
     strategy:
       matrix:
-        flink: &flink_versions [ 2.0.1, 2.1.1 ]
+        flink: &flink_versions [ 2.2.0, 2.1.1 ]
         jdk: &jdk_versions [ '11, 17, 21' ]
     uses: apache/flink-connector-shared-utils/.github/workflows/ci.yml@ci_utils
     with:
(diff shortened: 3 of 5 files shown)


mateczagany (approved):

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
```diff
@@ -268,6 +268,7 @@ private List<KafkaWriterState> createSnapshots(
 
     @Override
     public void close() throws Exception {
+        markClosed();
```
Savonitar, line 271: Shouldn't we add a regression test?

.github/workflows/weekly.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           flink: 2.2-SNAPSHOT,
           branch: main
         }, {
-          flink: 2.1.1,
```
Savonitar, line 36: Could you please add a note to the PR description about the version choices?

Savonitar (commented):
Hi. Thank you for the fix and CI changes.
Could you please check my comments?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
gaborgsomogyi: Was this the issue when last time we've tried to bump?

ferenc-csaky: > Was this the issue when last time we've tried to bump?

Yep

gaborgsomogyi (approved):
Thanks for the fix 🚢

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriter.java
Savonitar: just curious, Are you guys referring to https://github.com/apache/flink-connector-kafka/pull/204 ?

ferenc-csaky: > just curious, Are you guys referring to #204 ?

correct

Savonitar:
Sorry for the late reply.
Thanks for clarifications and the fix.

PR #257, opened 2026-05-21
[FLINK-39728] Fix pauseOrResumeSplits race with unassigned partitions

jnh5y:
When `pauseOrResumeSplits` is add to the mailbox thread, the set of splits it operates on reflects the mailbox thread's view *at start time*. 

By the time the SplitFetcher executes the action, a split may have already finished and been unassigned.

This can happen when the split-finished event has not yet propagated far enough to abort an in-flight alignment check on the `SourceOperator`. Calling `pause()` or `resume()` on a partition that is no longer in `consumer.assignment()` then throws `IllegalStateException`.

Generated-by: Claude Opus 4.6 (Anthropic)

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
@@ -240,14 +240,29 @@ public class KafkaPartitionSplitReader
     public void pauseOrResumeSplits(
             Collection<KafkaPartitionSplit> splitsToPause,
             Collection<KafkaPartitionSplit> splitsToResume) {
-        consumer.resume(
-                splitsToResume.stream()
-                        .map(KafkaPartitionSplit::getTopicPartition)
-                        .collect(Collectors.toList()));
-        consumer.pause(
-                splitsToPause.stream()
-                        .map(KafkaPartitionSplit::getTopicPartition)
-                        .collect(Collectors.toList()));
+        // Filter against current assignment to avoid IllegalStateException when a partition
+        // was concurrently unassigned by fetch() or removeEmptySplits().
+        Set<TopicPartition> assigned = consumer.assignment();
+        List<TopicPartition> toResume = new ArrayList<>();
+        for (KafkaPartitionSplit split : splitsToResume) {
+            TopicPartition tp = split.getTopicPartition();
+            if (assigned.contains(tp)) {
+                toResume.add(tp);
+            } else {
+                LOG.warn("Skipping resume for unassigned partition {}", tp);
+            }
+        }
+        List<TopicPartition> toPause = new ArrayList<>();
+        for (KafkaPartitionSplit split : splitsToPause) {
+            TopicPartition tp = split.getTopicPartition();
+            if (assigned.contains(tp)) {
+                toPause.add(tp);
+            } else {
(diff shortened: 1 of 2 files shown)


Savonitar (approved):
Hi, thank you for the fix. The filter against consumer.assignment() before each pause/resume call is the right defensive guard.

I find “concurrently” in the comments/description a bit misleading: the unassign and the pause/resume call are sequential on the SplitFetcher thread, not parallel. The race is between the mailbox thread’s view at enqueue time and the SplitFetcher’s state at execution time.

But it does not affect the proposed fix.

LGTM

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -240,13 +240,18 @@ public void close() throws Exception {
     public void pauseOrResumeSplits(
             Collection<KafkaPartitionSplit> splitsToPause,
             Collection<KafkaPartitionSplit> splitsToResume) {
+        // Filter against current assignment to avoid IllegalStateException when a partition
+        // was concurrently unassigned by fetch() or removeEmptySplits().
+        Set<TopicPartition> assigned = consumer.assignment();
```
Efrat19, line 245: Nit: Wdyt about leaving a warn here to unmask any edge case where pauseOrResumeSplits is called for unassigned partition from an unfinished split?

Efrat19 (commented):
Confirmed this `IllegalStateException` can occur when the split finished event doesn't propagate in time to abort alignment check on the sourceOperator, causing pauseOrResumeSplits to be called on a finished split
LGTM % nit

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
jnh5y: Added!

Efrat19: Thanks
I'd give up on the filter to avoid the 2X iterations
`.forEach(tp -> assigned.contains(tp) ? .. : ..)` might be simpler wdyt

jnh5y: Ok, how does this look? https://github.com/apache/flink-connector-kafka/compare/5a53da774d83ce6ddccdf04ebbdbf015c58e9d40..12701fa48b54693fda5d5bdb7034aadcfba321a1

I think there are a lot of ways this could look; happy to go with something else if you prefer.

Efrat19 (approved):
LGTM, thanks

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
Efrat19: Nope it looks good, do you require anything else to merge this?

jnh5y: I am not a committer; I don't see a button to merge this.

PR #258, opened 2026-05-28
[FLINK-39792] Add multi-headers metadata key to Kafka connector

AHeise:
The existing headers metadata key (MAP<STRING, BYTES>) silently drops duplicate header keys and loses insertion order, since iteration into a HashMap lets the last value win. The new multi-headers key exposes headers as ARRAY<ROW<name STRING, value BYTES>>, preserving wire order and duplicate keys faithfully on both source and sink sides.

The two keys are mutually exclusive on the write path: declaring both in the same table schema is rejected at planning time with a ValidationException. Null header values are preserved; null header keys throw IllegalArgumentException since the Kafka protocol does not permit them.

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchema.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchema.java
@@ -37,6 +37,7 @@ import org.apache.flink.types.RowKind;
 import org.apache.flink.util.Preconditions;
 
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.common.header.Header;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -171,7 +172,9 @@ class DynamicKafkaRecordSerializationSchema
                 readMetadata(consumedRow, KafkaDynamicSink.WritableMetadata.TIMESTAMP),
                 keySerialized,
                 valueSerialized,
-                readMetadata(consumedRow, KafkaDynamicSink.WritableMetadata.HEADERS));
+                resolveHeaders(
+                        readMetadata(consumedRow, KafkaDynamicSink.WritableMetadata.HEADERS),
+                        readMetadata(consumedRow, KafkaDynamicSink.WritableMetadata.HEADER_LIST)));
     }
 
     @Override
@@ -284,5 +287,15 @@ class DynamicKafkaRecordSerializationSchema
             return null;
         }
         return (T) metadata.converter.read(consumedRow, pos);
+    }
+
+    private static @Nullable List<Header> resolveHeaders(
+            @Nullable List<Header> mapHeaders, @Nullable List<Header> arrayHeaders) {
+        if (mapHeaders != null && arrayHeaders != null) {
+            throw new IllegalStateException(
+                    "Both 'headers' and 'header-list' metadata keys are active simultaneously. "
+                            + "This should have been caught during planning.");
+        }
+        return mapHeaders != null ? mapHeaders : arrayHeaders;
     }
 }
--- a/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
@@ -455,11 +455,12 @@ public class KafkaDynamicSink implements DynamicTableSink, SupportsWritingMetada
                         final ArrayData valueArray = map.valueArray();
                         final List<Header> headers = new ArrayList<>();
                         for (int i = 0; i < keyArray.size(); i++) {
-                            if (!keyArray.isNullAt(i) && !valueArray.isNullAt(i)) {
+                            if (!keyArray.isNullAt(i)) {
                                 // StringData.toString() decodes UTF-8; invalid bytes produce
                                 // replacement characters (see FLIP-568).
                                 final String key = keyArray.getString(i).toString();
-                                final byte[] value = valueArray.getBinary(i);
+                                final byte[] value =
+                                        valueArray.isNullAt(i) ? null : valueArray.getBinary(i);
                                 headers.add(new KafkaHeader(key, value));
                             }
                         }
--- a/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
@@ -31,6 +31,7 @@ import org.apache.flink.connector.kafka.sink.TransactionNamingStrategy;
 import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.datastream.DataStreamSink;
 import org.apache.flink.table.api.DataTypes;
+import org.apache.flink.table.api.ValidationException;
 import org.apache.flink.table.connector.ChangelogMode;
 import org.apache.flink.table.connector.Projection;
 import org.apache.flink.table.connector.ProviderContext;
@@ -274,6 +275,13 @@ public class KafkaDynamicSink implements DynamicTableSink, SupportsWritingMetada
 
     @Override
     public void applyWritableMetadata(List<String> metadataKeys, DataType consumedDataType) {
+        if (metadataKeys.contains(WritableMetadata.HEADERS.key)
+                && metadataKeys.contains(WritableMetadata.HEADER_LIST.key)) {
+            throw new ValidationException(
+                    String.format(
+                            "The writable metadata is ambiguous. Please use either metadata key '%s' or '%s'.",
+                            WritableMetadata.HEADERS.key, WritableMetadata.HEADER_LIST.key));
+        }
         this.metadataKeys = metadataKeys;
         this.consumedDataType = consumedDataType;
     }
@@ -448,9 +456,53 @@ public class KafkaDynamicSink implements DynamicTableSink, SupportsWritingMetada
                         final List<Header> headers = new ArrayList<>();
                         for (int i = 0; i < keyArray.size(); i++) {
                             if (!keyArray.isNullAt(i) && !valueArray.isNullAt(i)) {
+                                // StringData.toString() decodes UTF-8; invalid bytes produce
+                                // replacement characters (see FLIP-568).
                                 final String key = keyArray.getString(i).toString();
                                 final byte[] value = valueArray.getBinary(i);
                                 headers.add(new KafkaHeader(key, value));
+                            }
+                        }
+                        return headers;
+                    }
+                }),
+
+        /**
+         * Wire-faithful alternative to {@code headers}: preserves duplicate keys and insertion
+         * order using {@code ARRAY<ROW<key STRING, value BYTES>>} instead of a lossy MAP.
+         */
+        HEADER_LIST(
+                "header-list",
+                DataTypes.ARRAY(
+                                DataTypes.ROW(
+                                                DataTypes.FIELD(
+                                                        "key", DataTypes.STRING().nullable()),
+                                                DataTypes.FIELD(
+                                                        "value", DataTypes.BYTES().nullable()))
+                                        .nullable())
+                        .nullable(),
+                new MetadataConverter() {
+                    private static final long serialVersionUID = 1L;
+
+                    @Override
+                    public Object read(RowData row, int pos) {
+                        if (row.isNullAt(pos)) {
+                            return null;
+                        }
+                        final ArrayData arrayData = row.getArray(pos);
+                        final List<Header> headers = new ArrayList<>();
+                        for (int i = 0; i < arrayData.size(); i++) {
+                            if (!arrayData.isNullAt(i)) {
+                                final RowData headerRow = arrayData.getRow(i, 2);
+                                if (headerRow.isNullAt(0)) {
+                                    continue;
+                                }
+                                // StringData.toString() decodes UTF-8; invalid bytes produce
+                                // replacement characters (see FLIP-568).
+                                final String name = headerRow.getString(0).toString();
+                                final byte[] value =
+                                        headerRow.isNullAt(1) ? null : headerRow.getBinary(1);
+                                headers.add(new KafkaHeader(name, value));
                             }
                         }
                         return headers;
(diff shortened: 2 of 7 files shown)


flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchema.java
```diff
@@ -285,4 +289,14 @@ private <T> T readMetadata(RowData consumedRow, KafkaDynamicSink.WritableMetadat
         }
         return (T) metadata.converter.read(consumedRow, pos);
     }
+
+    private static @Nullable List<Header> resolveHeaders(
+            @Nullable List<Header> mapHeaders, @Nullable List<Header> arrayHeaders) {
+        if (mapHeaders != null && arrayHeaders != null) {
+            throw new IllegalStateException(
```
twalthr, line 295: Is this really a problem? Can't we simplify the logic and allow this? One problem less to care about.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
```diff
@@ -457,6 +465,47 @@ public Object read(RowData row, int pos) {
...
+                        final ArrayData arrayData = row.getArray(pos);
+                        final List<Header> headers = new ArrayList<>();
+                        for (int i = 0; i < arrayData.size(); i++) {
+                            if (!arrayData.isNullAt(i)) {
+                                final RowData headerRow = arrayData.getRow(i, 2);
+                                if (headerRow.isNullAt(0)) {
+                                    throw new IllegalArgumentException(
+                                            "Null header keys are not permitted.");
+                                }
+                                final String name = headerRow.getString(0).toString();
```
twalthr, line 503: We should mention here that we assume UTF-8 bytes and accept problems around:
https://cwiki.apache.org/confluence/display/FLINK/FLIP-568%3A+Strict+BYTES-to-STRING+CAST+with+UTF-8+Validation+Utilities
We don't intend to switch to StringData.fromUtf8Bytes as this would potentially block processing.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchema.java
AHeise: What do you suggest the semantics are? The user both wrote header and multi-header entries. Shall we simply concat?

I'd rather fail here: the user would need to choose between those two METADATA column for write, the other may remain available as a VIRTUAL METADATA column for read.

If we merge the entries, stuff becomes asymmetric. If you have a header [k1->v1], you'd end up with 1 entry in both METADATA on read side. If you forward the record, you end up with two entries [k1->v1, k1->v1].

twalthr: If the user asks for both we can simply provide it, in which case we read twice. Not our problem.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
```diff
@@ -457,6 +465,47 @@ public Object read(RowData row, int pos) {
...
+                        if (row.isNullAt(pos)) {
+                            return null;
+                        }
+                        final ArrayData arrayData = row.getArray(pos);
+                        final List<Header> headers = new ArrayList<>();
+                        for (int i = 0; i < arrayData.size(); i++) {
+                            if (!arrayData.isNullAt(i)) {
+                                final RowData headerRow = arrayData.getRow(i, 2);
+                                if (headerRow.isNullAt(0)) {
+                                    throw new IllegalArgumentException(
```
Savonitar: Could u please clarify nullability on the key field? 
Schema declares 
```
DataTypes.ARRAY(
                                DataTypes.ROW(
                                                DataTypes.FIELD(
                                                        "key", DataTypes.STRING().nullable()),
                                                DataTypes.FIELD(
                                                        "value", DataTypes.BYTES().nullable()))
                                        .notNull())
                        .nullable(),
```

however, at this line 
```
if (headerRow.isNullAt(0)) {
                                    throw new IllegalArgumentException(
                                    "Null header keys are not permitted.");
```

we throw. 
So we throw for a value the schema told the planner is valid. Shouldn't nullable() be notNull(), so the planner rejects null-key expressions at compile time instead of the job crashing at runtime?

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
```diff
@@ -457,6 +465,47 @@ public Object read(RowData row, int pos) {
                     }
                 }),
 
+        /**
+         * Wire-faithful alternative to {@code headers}: preserves duplicate keys and insertion
+         * order using {@code ARRAY<ROW<key STRING, value BYTES>>} instead of a lossy MAP.
+         */
+        MULTI_HEADERS(
```
Savonitar: shouldn't we add this to doc docs/content/docs/connectors/table/kafka.md  (which includes  headers for example).

Savonitar (commented):
Hi. Thanks for the PR! Sounds interesting, and I have a few questions.
Could you please take a look?

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java
```diff
@@ -623,6 +625,42 @@ public Object read(ConsumerRecord<?, ?> record) {
...
+                    private static final long serialVersionUID = 1L;
+
+                    @Override
+                    public Object read(ConsumerRecord<?, ?> record) {
+                        final List<GenericRowData> rows = new ArrayList<>();
+                        for (Header header : record.headers()) {
+                            final GenericRowData row = new GenericRowData(2);
+                            row.setField(
+                                    0,
+                                    header.key() == null
```
loserwang1024: Image
header key is never be null

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java
```diff
@@ -623,6 +625,42 @@ public Object read(ConsumerRecord<?, ?> record) {
                     }
                 }),
 
+        /**
+         * Wire-faithful alternative to {@code headers}: preserves duplicate keys and insertion
+         * order using {@code ARRAY<ROW<key STRING, value BYTES>>} instead of a lossy MAP.
+         */
+        MULTI_HEADERS(
+                "multi-headers",
```
loserwang1024: multl-headers is slightly ambiguous: does "multi-headers" mean "multiple headers" (just more than one)? Or "headers with multi-valued keys"? 

What about `headers-list` or `ordered-headers` @leonardBang , WDYT?

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
AHeise: Yes, good catch. I'll use the same behavior as regular header to skip instead. The main reason for nullable values is the ease of using it in queries. That's in line how we defined the existing `headers` metadata column.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java
AHeise: https://github.com/apache/flink-connector-kafka/pull/258#discussion_r3323329665

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java
AHeise: My favorite version was `header-list` but we also had `headers-list` (your proposal) and `headers-raw` in discussion. As always: names are hard so I appreciate feedback. Here is an AI summary:

- header-list — singular header matches Kafka's own Header Java type; -list signals an ordered, duplicate-allowing sequence. Pro: unambiguous item-type semantics, follows hyphen convention. Con: visually disconnected from the existing headers
  key.  
- headers-list — pairs directly with headers, making the relationship obvious. Pro: clear pairing, hyphen convention. Con: encodes structure ("list") in the key name; plural root still suggests MAP-like semantics.
- headers-raw — "raw" signals the lossless, wire-faithful representation vs. the lossy MAP projection (duplicate keys collapsed, order lost). Pro: semantically precise. Con: "raw" isn't established in the Flink connector vocabulary.
- multi-headers (chosen) — the prefix multi- directly conveys what is new: multiple headers per key are preserved. Reads naturally alongside headers without encoding implementation detail.

leonardBang: Naming nit on `multi-headers`: the prefix multi- is a bit unusual since headers is already plural, and the real selling point is "wire-faithful (preserves order + duplicates)" rather than "multi" from my poor English. Two alternatives worth considering:

- `header-list` — pairs nicely with the MAP-typed headers; the name itself hints at the data structure.
- `raw-headers` — emphasizes that this version is loss-free w.r.t. the Kafka wire format.


No strong opinion either way, but I'd lean toward `header-list` for symmetry with headers. Curious what others think.

twalthr: The reason for `multi-` came from the multimap concept. Basically, the same content as `headers` just as a multimap. But I'm also fine with `header-list` if the majority agrees.  But then the question arises `headers-list` or `header-list`?

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
twalthr: The reason for this is implicit casting. In SQL all columns are nullable by default. If we make headers not null, the implicit casting might have issues because casting from nullable to not nullable is not lossless. People would need to add IFNULL function calls.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java
AHeise: @loserwang1024 WDYT about the ideas tossed around? Any preference?

loserwang1024: @AHeise header-list looks good to me

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
```diff
@@ -274,6 +275,13 @@ public Map<String, DataType> listWritableMetadata() {
 
     @Override
     public void applyWritableMetadata(List<String> metadataKeys, DataType consumedDataType) {
+        if (metadataKeys.contains(WritableMetadata.HEADERS.key)
+                && metadataKeys.contains(WritableMetadata.HEADER_LIST.key)) {
+            throw new ValidationException(
+                    String.format(
+                            "Use either metadata keys '%s' or '%s'.",
```
twalthr:
```suggestion
                            "The writable metadata is ambiguous. Please use either metadata key '%s' or '%s'.",
```

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
```diff
@@ -447,16 +455,61 @@ public Object read(RowData row, int pos) {
...
+         */
+        HEADER_LIST(
+                "header-list",
+                DataTypes.ARRAY(
+                                DataTypes.ROW(
+                                                DataTypes.FIELD(
+                                                        "key", DataTypes.STRING().nullable()),
+                                                DataTypes.FIELD(
+                                                        "value", DataTypes.BYTES().nullable()))
+                                        .notNull())
```
twalthr: nit: I would also make this nullable. I don't know how good the SQL planner can handle it otherwise. Take e.g. `header[x].key`. If `header` is nullable, the call chain needs to be nullable anyways.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaRecordSerializationSchema.java
AHeise: The main issue here is that as soon as we allow both readable and writeable, we duplicate all entries unless we do an elaborate merging strategy. Let me show an example:

* Input has header [k->v]
* HEADERS has [k->v], HEADER-LIST has [k->v]
* Now if we write them both out by naive merging, we get [k->v, k->v], the entry is duplicated.

Since duplicates are exactly allowed with the new model we don't have an easy way out. We could always choose HEADER-LIST over HEADERS but I'm assuming users will be surprised if updates to HEADERS are lost. We could try to merge them more intelligently but everything is a heuristic and someone will be surprised.

This solution is the easiest way: both can be used for reading, only one can be used for writing. Might not be the most elegant solution but its the easiest to understand.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
AHeise: Applied.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSink.java
AHeise: Took this further: sink ROW is now nullable (matching the null-skip behavior already in the code). On the source side key is NOT NULL and null-key headers are silently skipped, so VIRTUAL users see a strict type without null checks on the key. Added testHeaderListMetadataTypes asserting both shapes, and extended testKafkaSourceSinkWithMultiHeaders with a null-ROW round-trip.

twalthr (approved):
LGTM, thanks for the update @AHeise

Savonitar (approved):
LGTM, thanks for addressing the feedback!

PR #261, opened 2026-06-01
[FLINK-39723] Wait for partition assignment in KafkaTableTestBase.createTestTopic

Savonitar:
This is the same readiness race that was previously addressed by FLINK-39234 for other topic creation paths.


MartijnVisser (approved):
LGTM. This matches the root cause from the JIRA, and I independently arrived at and verified the same approach (routing `createTestTopic` through `createNewTopicAndWaitForPartitionAssignment`): it compiles, passes spotless/checkstyle, and `KafkaTableITCase` runs green locally.
  
Worth setting expectations: the failure is a load-dependent metadata-propagation race that doesn't reproduce on an idle local box (30 local runs all passed), so the real confirmation will be CI stability over the next few weekly runs. But the change closes the race cleanly and weakens no assertions, so I'm happy with it.
  
Nit (non-blocking): consider adding the component tag to the title, e.g.
`[FLINK-39723][tests] …`, to match the commit-message convention. Approving regardless, assuming CI will pass. I'll merge it afterwards

MartijnVisser:
One thing I'd recommend is to also backport this to the older release that are still supported

PR #262, opened 2026-06-02
[FLINK-39822][tests] Wait for partition assignment in test createTopics calls

Savonitar:
This is the same readiness race that was previously addressed by [FLINK-39234](https://issues.apache.org/jira/browse/FLINK-39234) for other topic creation paths.

### DIFF
--- a/flink-connector-kafka-e2e-tests/flink-end-to-end-tests-common-kafka/src/main/java/org/apache/flink/tests/util/kafka/KafkaContainerClient.java
+++ b/flink-connector-kafka-e2e-tests/flink-end-to-end-tests-common-kafka/src/main/java/org/apache/flink/tests/util/kafka/KafkaContainerClient.java
@@ -19,12 +19,11 @@
 package org.apache.flink.tests.util.kafka;
 
 import org.apache.flink.api.common.time.Deadline;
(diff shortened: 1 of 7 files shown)


MartijnVisser (approved):
Good improvement! Thanks for driving

Will you also backport this one to the other branches?

Savonitar:
Hi @MartijnVisser . Thank you for the review and merge.

> Will you also backport this one to the other branches?

Happy to. However, one thing worth flagging from looking at the release-branch weekly CI: this particular readiness race hasn't actually surfaced on v4.0/v3.4, probably because they still run the older ZooKeeper-based testcontainer and show none of the `UnknownTopicOrPartition` failures we see on main's KRaft setup.
Backporting the current fix will require more review-efforts (because some methods/classes do not exist in v4.0/3.4) meanwhile currently there is no 100% proof on CI that same issues appear there.

However, I analyzed recent CI failures on v4.0 and going to raise: 1 backport of another improvement and fix another potential issue on main branch + backport it to 4.0(unrelated to KRaft and this ticket).

Let me know if you think it makes sense anyway to backport KRaft-based fixes with relevant code.

Savonitar:
https://github.com/apache/flink-connector-kafka/pull/263

PR #266, opened 2026-06-05
[FLINK-39940] DynamicKafkaSource should refresh restored cluster properties from KafkaMetadataService

bowenli86:
## Why / Motivation

`DynamicKafkaSource` restores `KafkaStream` metadata from the checkpointed enumerator state when a job is resumed from a checkpoint or savepoint. Before this change, restored sub-enumerators were rebuilt from the checkpointed cluster properties without first refreshing them from `KafkaMetadataService`. That means non-bootstrap Kafka client properties such as authentication or security settings can be stale at restore time, and the job can fail before the next periodic metadata refresh has a chance to reconcile the latest metadata.

The restore path should use the latest metadata-service properties immediately, while still preserving the checkpointed `bootstrap.servers` for restored clusters so the enumerator reconnects to the same restored cluster identity. Restore-time metadata lookup failures should also fail fast instead of being deferred until a later discovery cycle.

This PR also adds DEBUG-level startup-offset logs for checkpoint snapshot and restore to make restore issues easier to diagnose.

## How

- On restore, fetch the latest subscribed streams from `KafkaMetadataService` before rebuilding restored cluster enumerators.
- Merge refreshed cluster properties into the checkpointed `KafkaStream` metadata by cluster id.
- Preserve checkpointed `bootstrap.servers` when present, and fall back to the checkpointed properties if a restored cluster is not present in the fetched metadata.
- Keep the checkpointed restored cluster/topic set for initial sub-enumerator reconstruction; the existing metadata refresh flow still reconciles removed and newly added clusters afterward.
- Surface restore-time metadata refresh failures immediately.
- Add DEBUG logs that summarize assigned and unassigned startup offsets during checkpoint snapshot and restore.

## Tests

- Added coverage for refreshing non-bootstrap restored cluster properties while preserving checkpointed `bootstrap.servers`.
- Added coverage for reconciling removed and added clusters after restore.
- Added coverage for failing fast when restore-time metadata refresh fails.
- Added coverage for checkpoint/restore startup-offset DEBUG logs.
- Validated with the CI-equivalent Maven build used by the failing Python test job: `mvn clean install -U -B --no-transfer-progress -Dflink.version=2.2.1 -DskipTests -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120 -Dlog4j.configurationFile=file://$(pwd)/tools/ci/log4j.properties`.

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
@@ -41,6 +41,7 @@ import org.apache.flink.connector.kafka.source.enumerator.subscriber.KafkaSubscr
 import org.apache.flink.connector.kafka.source.split.KafkaPartitionSplit;
 import org.apache.flink.util.Preconditions;
 
+import org.apache.kafka.clients.CommonClientConfigs;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.common.KafkaException;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


lnbest0707 (approved):

lnbest0707:
LGTM

bowenli86:
thanks, merging

PR #267, opened 2026-06-08
[FLINK-39888][Kafka] Allow configuring offset reset strategy independently

jigar-bhati:
### What

This PR makes an explicitly configured Kafka `auto.offset.reset` property take precedence over the reset strategy inferred from the source's starting-offset initializer. If the property is not configured, the existing initializer-derived behavior is preserved.

The behavior is applied consistently across:

- the static `KafkaSourceBuilder`;
- `DynamicKafkaSourceBuilder` and dynamic-source cluster metadata handling in the enumerator and reader;
- Kafka and dynamic-Kafka SQL/table sources for every `scan.startup.mode`.

The table sources no longer need to special-case `group-offsets` to propagate the reset strategy. They validate and normalize explicit reset values before forwarding them to Kafka, including case-insensitive values accepted by existing configurations. The DataStream and table documentation is updated to describe the effective `auto.offset.reset` behavior.

### Why

Selecting the initial Kafka position and deciding what to do when an established position becomes unavailable are separate concerns. For example, a source may correctly start from earliest or latest, consume normally, and later encounter an out-of-range concrete offset after a Kafka leader change or log truncation. Silently resetting that position to earliest or latest can replay or skip data.

Previously, the builders and dynamic-source metadata paths overwrote an explicit `auto.offset.reset=none` with the strategy inferred from the startup initializer. This made it impossible to request fail-fast behavior for sources using earliest, latest, specific-offset, or timestamp startup modes. With this change, users can explicitly select `none` (or another Kafka-supported reset strategy), while jobs that do not configure the property retain their existing behavior.

Fresh earliest/latest startup still uses its selected starting position; the reset property controls the fallback when Kafka cannot use a position.

The following table summarizes the behavior for a fresh source and for a restored or already-running reader whose concrete position becomes unavailable:

| Explicit `auto.offset.reset` | Fresh-start position before/after | Out-of-range behavior before this PR | Out-of-range behavior after this PR |
|---|---|---|---|
| Not set | Selected by the startup initializer | Reset using the initializer-derived strategy | Unchanged: reset using the initializer-derived strategy |
| `none` | Selected by the startup initializer | Reset using the initializer-derived strategy | Fail with an out-of-range error |
| `earliest` | Selected by the startup initializer | Reset using the initializer-derived strategy | Reset to earliest |
| `latest` | Selected by the startup initializer | Reset using the initializer-derived strategy | Reset to latest |

The initializer-derived strategies are `earliest` for earliest and specific-offset startup, `latest` for latest and timestamp startup, and `none` by default for group-offset startup. Group-offset startup is the one case where a missing committed position also immediately uses the effective reset strategy.

For dynamic sources, the builder intentionally leaves `auto.offset.reset` absent when the user did not configure it. This preserves the distinction between an explicit property and a builder-generated default until the effective cluster initializer is known. The reader and enumerator then resolve the same precedence:

```text
explicit cluster reset property > explicit global reset property > effective initializer strategy
```

| Global initializer | Cluster initializer | Global reset property | Cluster reset property | Effective dynamic reset |
|---|---|---|---|---|
| earliest | Not set | Not set | Not set | earliest |
| earliest | latest | Not set | Not set | latest |
| latest | earliest | Not set | Not set | earliest |
| earliest | latest | `none` | Not set | none |
| earliest | latest | Not set | `none` | none |
| earliest | latest | `none` | `earliest` | earliest |

This keeps explicit cluster-property precedence consistent with other dynamic Kafka properties. When no cluster reset property is configured, an explicit global reset property still wins over an initializer-derived cluster default. The table sources also retain early validation of invalid reset values across all startup modes.

## Jira

https://issues.apache.org/jira/browse/FLINK-39888

Related existing report describing leader-change/log-truncation reset behavior:
https://issues.apache.org/jira/browse/FLINK-35630

## Testing

The affected suites cover the initializer-derived default, preservation and precedence of an explicit `auto.offset.reset`, static and dynamic source builders, dynamic reader/enumerator paths, and table reset validation and normalization for all SQL startup modes:

```bash
./mvnw -pl flink-connector-kafka -DskipITs -Dfast test \
  -Dtest=KafkaPropertiesUtilTest,KafkaSourceBuilderTest,DynamicKafkaSourceBuilderTest,DynamicKafkaSourceReaderTest,DynamicKafkaSourceEnumeratorTest,KafkaDynamicTableFactoryTest,DynamicKafkaTableFactoryTest
```

Result: `Tests run: 114, Failures: 0, Errors: 0, Skipped: 0`.

```bash
./mvnw -pl flink-connector-kafka spotless:check
```

Result: `BUILD SUCCESS`.


bowenli86:
kicked off CI
dynamic source path lgtm, will be great to get some folks to look at the static source path

jigar-bhati:
Thanks @bowenli86.

Based on recent Kafka connector activity, @Efrat19 and @Savonitar seem like good folks to ask for the static Kafka source/table-source side. Would either of you be willing to take a look at the static source changes in this PR?

The dynamic source path has an initial LGTM from Bowen and CI is green already

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaTableSource.java
```diff
@@ -462,52 +462,57 @@ protected DynamicKafkaSource<RowData> createDynamicKafkaSource(
                 .setDeserializer(kafkaDeserializer)
                 .setProperties(properties);
 
-        switch (startupMode) {
-            case EARLIEST:
-                dynamicKafkaSourceBuilder.setStartingOffsets(OffsetsInitializer.earliest());
+        dynamicKafkaSourceBuilder.setStartingOffsets(getStartingOffsetsInitializer());
```
Efrat19, line 466: There's a change to default behavior we would probably want to document in release notes - because the so-far ignored user set `ConsumerConfig.AUTO_OFFSET_RESET_CONFIG` will now take over the one inferred by the `StartingOffsetsInitializer`

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSourceBuilder.java
```diff
@@ -382,7 +382,7 @@ public KafkaSourceBuilder<OUT> setRackIdSupplier(SerializableSupplier<String> ra
      * created.
      *
      * <ul>
-     *   <li><code>auto.offset.reset.strategy</code> is overridden by {@link
+     *   <li><code>auto.offset.reset</code> is overridden by {@link
```
Efrat19: Nit: Separate commit?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/initializer/OffsetsInitializer.java
```diff
@@ -122,6 +124,31 @@ static OffsetsInitializer committedOffsets(OffsetResetStrategy offsetResetStrate
...
+     * <p>The offset reset strategy is only used when the returned initializer is used to initialize
+     * starting offsets. The initialized offsets themselves are unchanged, so initializers such as
+     * {@link #earliest()} and {@link #latest()} keep their normal startup behavior.
+     *
+     * @param offsetsInitializer the initializer which resolves the starting offsets.
+     * @param offsetResetStrategy the offset reset strategy to use when the initialized starting
+     *     offsets are out of range.
+     * @return an {@link OffsetsInitializer} with the given offset reset strategy.
+     */
+    static OffsetsInitializer withOffsetResetStrategy(
```
Efrat19: Am I right to understand that:
1. In dynamic/ SQL mode, `properties.auto.offset.reset` is the effective configuration 
2. In DataStream mode, the same configuration (`ConsumerConfig.AUTO_OFFSET_RESET_CONFIG`) is still ignored, and a manual `OffsetResetStrategy` can only be specified by the mutually exclusive `withOffsetResetStrategy`?

Can we use `ConsumerConfig.AUTO_OFFSET_RESET_CONFIG` in DataStream as well instead?

docs/content.zh/docs/connectors/datastream/kafka.md
```diff
@@ -222,7 +226,7 @@ Kafka Source 支持流式和批式两种运行模式。默认情况下，KafkaSo
 Kafka consumer 的配置可以参考 [Apache Kafka 文档](http://kafka.apache.org/documentation/#consumerconfigs)。
 
 请注意，即使指定了以下配置项，构建器也会将其覆盖：
-- ```auto.offset.reset.strategy``` 被 OffsetsInitializer#getAutoOffsetResetStrategy() 覆盖
```
Efrat19, line 225: ditto

docs/content/docs/connectors/datastream/kafka.md
```diff
@@ -235,7 +239,7 @@ for more details.
 
 Please note that the following keys will be overridden by the builder even if
 it is configured:
-- ```auto.offset.reset.strategy``` is overridden by ```OffsetsInitializer#getAutoOffsetResetStrategy()```
```
Efrat19, line 238: ditto

Efrat19 (commented):
Hello, thank you for the contribution!

+1 the reported issue is a genuine gap
When recovering from checkpoint, the effective offset is the one pulled from `KafkaSourceEnumState`, and the user-set `StartingOffsetInitializer` is (correctly) ignored.
Therefore during recovery `KafkaSourceEnumerator` should fallback to `OffsetResetStrategy.NONE`, or use the user-set `ConsumerConfig.AUTO_OFFSET_RESET_CONFIG`,
Instead of incorrectly reading the OffsetResetStrategy of the stale StartingOffsetInitializer

However to me it still makes sense for offset reset strategy to be coupled with offset initializer, as it can always be auto inferred except during recovery:
I would propose a slightly different fix - to rely on user-set `ConsumerConfig.AUTO_OFFSET_RESET_CONFIG` during recovery.

This solves the described issue without adding a possibly confusing user knob that can lead to Irrational combinations
like `withOffsetResetStrategy(OffsetsInitializer.latest(), OffsetResetStrategy.EARLIEST)` or vice versa.
wdyt?

jigar-bhati:
Thanks, this makes sense. I revised the PR in this direction.

The new version removes `OffsetsInitializer.withOffsetResetStrategy(...)` and the wrapper initializer entirely. Instead, `ConsumerConfig.AUTO_OFFSET_RESET_CONFIG` is now respected consistently. If the user explicitly configures it, that value is preserved; otherwise the builder keeps the existing behavior and derives `auto.offset.reset` from `OffsetsInitializer#getAutoOffsetResetStrategy()`.

So this should address the recovery gap for users who opt into `auto.offset.reset=none`, while avoiding a default behavior change for existing jobs.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaTableSource.java
jigar-bhati: Good point. I revised this so the default behavior is preserved: `auto.offset.reset` is only taken from user properties when explicitly configured. Otherwise it remains initializer-derived as before.

This still means explicitly configured `auto.offset.reset` is now respected instead of ignored, but there is no behavior change for jobs that do not set it.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/initializer/OffsetsInitializer.java
jigar-bhati: Yes, that was the behavior in the earlier version, and I agree it was not ideal.

I removed `OffsetsInitializer.withOffsetResetStrategy(...)` and the wrapper initializer. The PR now uses the existing `ConsumerConfig.AUTO_OFFSET_RESET_CONFIG` path instead, including DataStream mode.

docs/content/docs/connectors/datastream/kafka.md
jigar-bhati: Updated. The docs no longer mention the removed wrapper API and now say `auto.offset.reset` is derived from the initializer only when it is not explicitly configured.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSourceBuilder.java
jigar-bhati: After the revision, I kept this in the same commit because respecting explicit `ConsumerConfig.AUTO_OFFSET_RESET_CONFIG` in DataStream is now the core of the fix rather than a side change. Happy to split if maintainers still prefer.

docs/content.zh/docs/connectors/datastream/kafka.md
jigar-bhati: Updated to match the English docs.

jigar-bhati:
@Efrat19 let me know what you think of the last set of change. Thanks!

Efrat19:
> @Efrat19 let me know what you think of the last set of change. Thanks!

Hey, thank you for addressing the [default behavior concern](https://github.com/apache/flink-connector-kafka/pull/267#discussion_r3448100102) and the [APIs drift](https://github.com/apache/flink-connector-kafka/pull/267#discussion_r3448114634). I think it looks better now.

1. Please update the PR description
2. I am still concerned about the chosen approach. I think the main gap here is that a different `auto.offset.reset` strategy should be used on startup vs recovery.
Previously it was always inferred by `StartingOffsetInitializer`, (incorrect during recovery), and now it will always be read from user's `auto.offset.reset` (when provided), and will be wrong during startup so we are only shifting the problem somewhere else.
Do you think we can distinguish startup and recovery modes, and use `StartingOffsetInitializer. getAutoOffsetResetStrategy` for startup, and `none`/user provided `auto.offset.reset` during recovery?
Or am I missing something and the issue/proposed solution isn't just about recovery?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
```diff
@@ -283,7 +272,10 @@ public void handleSourceEvents(SourceEvent sourceEvent) {
                         clusterMetadataMapEntry.getValue().getProperties(), clusterProperties);
                 OffsetsInitializer startingOffsetsInitializer =
                         clusterMetadataMapEntry.getValue().getStartingOffsetsInitializer();
-                if (startingOffsetsInitializer != null) {
+                if (startingOffsetsInitializer != null
+                        && !properties.containsKey(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG)
```
Savonitar: Does properties always contain `auto.offset.reset` ? Because in this PR in [flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/DynamicKafkaSourceBuilder.java](https://github.com/apache/flink-connector-kafka/pull/267/changes#diff-383813096aaf9fe7c22571d538deba5bd12efea9200d7f1b6c63d77e7e9631cdR260) we use maybeOverride(..., false) . 
If properties always contain auto.offset.reset` , it means, this condition is always false.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -601,12 +534,13 @@ private KafkaSourceEnumerator createEnumeratorWithAssignedTopicPartitions(
         KafkaPropertiesUtil.copyProperties(properties, consumerProps);
         DynamicKafkaSourceOptions.removeRemovedClusterRetentionOption(consumerProps);
         KafkaPropertiesUtil.setClientIdPrefix(consumerProps, kafkaClusterId);
-        consumerProps.setProperty(
+        consumerProps.computeIfAbsent(
```
Savonitar: is it same issue as in https://github.com/apache/flink-connector-kafka/pull/267/changes#r3512234425 ?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/DynamicKafkaSourceBuilder.java
```diff
@@ -257,7 +257,7 @@ private void setRequiredConsumerProperties() {
         maybeOverride(
                 ConsumerConfig.AUTO_OFFSET_RESET_CONFIG,
                 startingOffsetsInitializer.getAutoOffsetResetStrategy().name().toLowerCase(),
-                true);
+                false);
```
Savonitar: does it cause all `contain*` checks can no longer distinguish between user vs builder-default value?

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicTableFactoryTest.java
```diff
@@ -474,15 +474,33 @@ public void testTableSourceSetOffsetReset(final String strategyName) {
         testSetOffsetResetForStartFromGroupOffsets(strategyName);
     }
 
-    @Test
-    void testTableSourceSetOffsetResetWithException() {
```
Savonitar, line 478: Could you please clarify why this test was deleted?

Savonitar (commented):
Hi. Thanks for the PR. I left comments. 
Could you please take a look?

jigar-bhati:
Thanks, Efrat. One clarification on what you mean by recovery: are you thinking specifically about Flink checkpoint restoration, or about any previously established concrete Kafka position becoming unavailable, including for an already-running reader?

The motivating case is not limited to checkpoint restoration. A reader can start successfully and consume normally, then a Kafka leader change/log truncation can make its current position out of range (there is a similar existing report in [FLINK-35630](https://issues.apache.org/jira/browse/FLINK-35630)). In that case, retaining the startup initializer's `earliest`/`latest` reset strategy would still silently reset the running reader.

The current PR intentionally preserves an explicitly configured `auto.offset.reset` for the lifetime of the consumer, while keeping the existing initializer-derived behavior when the property is absent. This allows, for example:

```text
earliest startup + auto.offset.reset=none
  -> fresh source starts from earliest normally
  -> a restored or running concrete position later becomes out of range
  -> fail instead of silently resetting to earliest
```

Fresh earliest startup uses the earliest marker and seeks to the beginning; latest startup resolves and seeks to the current end. So `none` does not prevent those normal startups. It intentionally affects cases where Kafka cannot use the selected position (including a missing committed position or an invalid specific position).

If your suggestion is to use the initializer strategy only for the initial position and the explicit strategy for any later out-of-range position, I think that is the behavior we want conceptually. Since an out-of-range can also occur during normal execution, though, limiting the change to `restoreEnumerator(...)` would not cover the leader-truncation case. Does that match the scenario you had in mind?

I also updated and added a table in the PR summary for clarification on how different cases work after this PR.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
jigar-bhati: Good catch. The dynamic builder no longer materializes an initializer-derived `auto.offset.reset` when it is absent. The reader now receives the global initializer and resolves explicit global > explicit cluster > effective initializer, so a cluster-specific initializer is still honored when no reset property is configured.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
jigar-bhati: Yes, this was the same issue. The enumerator now uses the shared reset-strategy resolver with the raw global properties, fetched cluster properties, and effective initializer, so a generated default can no longer mask the cluster initializer and reader/enumerator precedence stays consistent.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/DynamicKafkaSourceBuilder.java
jigar-bhati: Correct. I removed the eager dynamic-builder default, so property presence again means it was explicitly configured. The reader/enumerator derive the fallback only after cluster metadata is known; I also added coverage for global, cluster, and initializer precedence.

flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicTableFactoryTest.java
jigar-bhati: Restored and expanded this coverage. Invalid `auto.offset.reset` values are now rejected early for all five SQL startup modes (earliest, latest, specific offsets, timestamp, and group offsets), with the existing clear validation error.

Efrat19:
Thanks @jigar-bhati for clarifying, that sounds reasonable then.
So the only question left is will / how would we handle config combinations that are probably a mistake
like `earliest startup + auto.offset.reset=latest`,`latest startup + auto.offset.reset=earliest`

jigar-bhati:
Thanks @Efrat19. I think we should leave these combinations allowed. `scan.startup.mode` selects the initial position, while `auto.offset.reset` controls the fallback when an initialized position becomes unavailable. Since they are independent options, different values can be intentional.

We could add validation later if these combinations prove confusing, but I would prefer to start without restricting valid Kafka configurations. I added a short documentation clarification that the two options control different phases and may intentionally use different values.

jigar-bhati:
@Efrat19 @Savonitar let me know if the PR is good to get merged. Would appreciate a stamp/merge if we are all aligned now!

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaConnectorOptionsUtil.java
```diff
@@ -115,6 +119,28 @@ public static void validateTableSinkOptions(ReadableConfig tableOptions) {
...
 
+    static void validateAutoOffsetResetStrategy(Properties properties) {
+        String resetStrategy = properties.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG);
+        if (resetStrategy == null) {
+            return;
+        }
+
+        boolean valid =
+                Arrays.stream(OffsetResetStrategy.values())
+                        .anyMatch(strategy -> strategy.name().equalsIgnoreCase(resetStrategy));
```
Savonitar: > equalsIgnoreCase

Here we accept values case insensitive. However, Kafka's own auto.offset.reset validator is case-sensitive, so 
```
'properties.auto.offset.reset' = 'EARLIEST' 
```
passes plan-time validation and then every source subtask dies on the TaskManager:
```
Invalid value EARLIEST for configuration auto.offset.reset: ... must be either 'earliest', 'latest', 'none'
```

This looks like a regression, because on main today, scan.startup.mode = 'group-offsets' + 'properties.auto.offset.reset' = 'EARLIEST' works, because the old getResetStrategy compared via toUpperCase(Locale.ROOT) and the builder then overwrote the property with the lower-cased enum name https://github.com/apache/flink-connector-kafka/blob/aa6be4bac8aeab15b471820f037fcf3cecca044c/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSourceBuilder.java#L471 (override flag was **true**)

After this PR the uppercase string survives to the consumer.

I see 2 options:
1. Normalise with toLowerCase
OR
2. Make the validator case-sensitive so at least the failure surfases at plan time with the clear message.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/DynamicKafkaSourceBuilderTest.java
```diff
@@ -0,0 +1,112 @@
...
+    private DynamicKafkaSourceBuilder<Integer> baseBuilder() {
+        return DynamicKafkaSource.<Integer>builder()
+                .setStreamIds(Collections.singleton("stream-1"))
+                .setKafkaMetadataService(NoOpKafkaMetadataService.INSTANCE)
+                .setDeserializer(
+                        KafkaRecordDeserializationSchema.valueOnly(IntegerDeserializer.class));
+    }
+
+    private static Properties extractProperties(DynamicKafkaSource<?> source) throws Exception {
+        Field field = DynamicKafkaSource.class.getDeclaredField("properties");
```
Savonitar: Could you please clarify, why do we use reflection here ? I checked and looks like only this place (and DynamicKafkaSourceReaderTest) use reflection in tests.
Why can't we reuse existing approach: package private accessor, for example?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaPropertiesUtil.java
```diff
@@ -36,6 +39,25 @@ public static void copyProperties(@Nonnull Properties from, @Nonnull Properties
         }
     }
 
+    /** Resolves an explicit global or cluster reset strategy before the initializer default. */
```
Savonitar: According to the PR description, is my understanding correct, that an explicit global auto.offset.reset intentionally wins over a per-cluster value supplied via KafkaMetadataService ( the **opposite** of how cluster properties override global ones for **every other key**)?

Savonitar (changes requested):
Thanks for the rework, I see that all my points I raised earlier are addressed.
I found one blocker before merge (case normalization). Please correct me if im wrong.
I left a few more comments.

flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaConnectorOptionsUtil.java
jigar-bhati: Thanks for catching this; it was a real regression. I normalized valid table reset values to lowercase during validation before they reach the Kafka consumer, and added coverage for uppercase `EARLIEST`. This preserves the existing `group-offsets` behavior and avoids a later TaskManager failure.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/DynamicKafkaSourceBuilderTest.java
jigar-bhati: Thanks, agreed. I replaced the reflection helper with a package-private `@VisibleForTesting` accessor on `DynamicKafkaSource` and removed the reflection import/helper from the test.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaPropertiesUtil.java
jigar-bhati: Thanks for pointing this out; that precedence was inconsistent with the rest of the dynamic source. I changed it to explicit cluster property > explicit global property > effective initializer strategy. An explicit global value still wins over a cluster initializer, because the initializer remains only the fallback when neither property is configured.

Efrat19:
> Thanks @Efrat19. I think we should leave these combinations allowed. `scan.startup.mode` selects the initial position, while `auto.offset.reset` controls the fallback when an initialized position becomes unavailable. Since they are independent options, different values can be intentional.
> 
> We could add validation later if these combinations prove confusing, but I would prefer to start without restricting valid Kafka configurations. I added a short documentation clarification that the two options control different phases and may intentionally use different values.

Outside broker corruption its hard to imagine a case where you want this values to be independent. For years the value was inferred from the offsetInitializer.
I'd still log a warning, at least to signal that this isn't optimal.

jigar-bhati:
Thanks, sounds reasonable. I added a warning for the likely-mistake positional combinations (`earliest` startup with `auto.offset.reset=latest`, and `latest` startup with `auto.offset.reset=earliest`) in both the static and dynamic source paths. The warning explains that startup still follows the initializer while Kafka may use the configured reset strategy if an initialized offset later becomes unavailable. I intentionally do not warn for `auto.offset.reset=none`, since `earliest`/`latest` startup plus fail-fast handling for an unavailable concrete offset is the motivating broker-truncation use case. The combinations remain allowed rather than rejected.

Savonitar (commented):
Hi, thanks for working on this PR. I can continue the review after my vacation. If it’s a blocker, I’m fine with other reviewers approving it.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaPropertiesUtil.java
```diff
@@ -36,6 +39,41 @@ public static void copyProperties(@Nonnull Properties from, @Nonnull Properties
...
+            @Nonnull OffsetsInitializer startingOffsetsInitializer) {
+        String initializerResetStrategy =
+                startingOffsetsInitializer.getAutoOffsetResetStrategy().name();
+        return isPositionalResetStrategy(configuredResetStrategy)
+                && isPositionalResetStrategy(initializerResetStrategy)
+                && !configuredResetStrategy.equalsIgnoreCase(initializerResetStrategy);
+    }
+
+    private static boolean isPositionalResetStrategy(String resetStrategy) {
+        return "earliest".equalsIgnoreCase(resetStrategy)
```
Efrat19: `OffsetResetStrategy.EARLIEST`

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaPropertiesUtil.java
```diff
@@ -36,6 +39,41 @@ public static void copyProperties(@Nonnull Properties from, @Nonnull Properties
...
+                || "latest".equalsIgnoreCase(resetStrategy);
```
Efrat19: `OffsetResetStrategy.LATEST`

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaPropertiesUtil.java
```diff
@@ -36,6 +39,41 @@ public static void copyProperties(@Nonnull Properties from, @Nonnull Properties
...
+            @Nonnull String configuredResetStrategy,
+            @Nonnull OffsetsInitializer startingOffsetsInitializer) {
+        String initializerResetStrategy =
+                startingOffsetsInitializer.getAutoOffsetResetStrategy().name();
+        return isPositionalResetStrategy(configuredResetStrategy)
+                && isPositionalResetStrategy(initializerResetStrategy)
+                && !configuredResetStrategy.equalsIgnoreCase(initializerResetStrategy);
+    }
+
+    private static boolean isPositionalResetStrategy(String resetStrategy) {
```
Efrat19:
```suggestion
    private static boolean isPositionalResetStrategy(OffsetResetStrategy resetStrategy) {
```

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaPropertiesUtil.java
```diff
@@ -36,6 +39,41 @@ public static void copyProperties(@Nonnull Properties from, @Nonnull Properties
...
+        if (globalReset != null) {
+            return globalReset;
+        }
+
+        return startingOffsetsInitializer.getAutoOffsetResetStrategy().name().toLowerCase();
+    }
+
+    /** Returns whether the configured strategy opposes a positional initializer strategy. */
+    public static boolean hasOpposingOffsetResetStrategies(
+            @Nonnull String configuredResetStrategy,
```
Efrat19:
```suggestion
            @Nonnull OffsetResetStrategy configuredResetStrategy,
```

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSourceBuilder.java
```diff
@@ -468,10 +468,25 @@ private void parseAndSetRequiredProperties() {
             maybeOverride(KafkaSourceOptions.COMMIT_OFFSETS_ON_CHECKPOINT.key(), "false", false);
         }
         maybeOverride(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "false", false);
+        String configuredOffsetResetStrategy =
```
Efrat19:
```suggestion
        OffsetResetStrategy configuredOffsetResetStrategy =
```
See [getResetStrategy](https://github.com/apache/flink-connector-kafka/blob/main/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaTableSource.java#L513) (can probably be made a static util)

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaPropertiesUtil.java
```diff
@@ -36,6 +39,41 @@ public static void copyProperties(@Nonnull Properties from, @Nonnull Properties
...
+        if (clusterReset != null) {
+            return clusterReset;
+        }
+
+        String globalReset = globalProperties.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG);
+        if (globalReset != null) {
+            return globalReset;
+        }
+
+        return startingOffsetsInitializer.getAutoOffsetResetStrategy().name().toLowerCase();
```
Efrat19: Nit:
```
return clusterProperties.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG) || globalProperties.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG) || startingOffsetsInitializer.getAutoOffsetResetStrategy()
```

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -534,12 +534,27 @@ private KafkaSourceEnumerator createEnumeratorWithAssignedTopicPartitions(
         KafkaPropertiesUtil.copyProperties(properties, consumerProps);
         DynamicKafkaSourceOptions.removeRemovedClusterRetentionOption(consumerProps);
         KafkaPropertiesUtil.setClientIdPrefix(consumerProps, kafkaClusterId);
+        String effectiveOffsetResetStrategy =
```
Efrat19: ditto

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -534,12 +534,27 @@ private KafkaSourceEnumerator createEnumeratorWithAssignedTopicPartitions(
...
+                KafkaPropertiesUtil.resolveAutoOffsetResetStrategy(
+                        properties, fetchedProperties, effectiveStartingOffsetsInitializer);
+        if (KafkaPropertiesUtil.hasOpposingOffsetResetStrategies(
```
Efrat19: I'd move this check to `DynamicKafkaSourceBuilder` with the rest of the dynamic source config validations, wdyt?

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaPropertiesUtilTest.java
```diff
@@ -63,6 +63,31 @@ void testClusterResetPropertyOverridesGlobalAndInitializerStrategies() {
                 .isEqualTo("earliest");
     }
 
+    @Test
```
Efrat19: I'd switch to `@ParameterizedTest` and cover all 4*4 cases with less code, wdyt

Efrat19 (commented):
Thanks

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
Efrat19: See https://github.com/apache/flink-connector-kafka/pull/267#discussion_r3747264514

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaPropertiesUtil.java
jigar-bhati: Done. `hasOpposingOffsetResetStrategies` now accepts `OffsetResetStrategy`, and the positional check uses `OffsetResetStrategy.EARLIEST` / `OffsetResetStrategy.LATEST` directly.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSourceBuilder.java
jigar-bhati: Done. I added shared `KafkaPropertiesUtil.getResetStrategy(...)`; the static and dynamic paths now parse configured properties through it and use the enum internally.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaPropertiesUtil.java
jigar-bhati: Done. `resolveAutoOffsetResetStrategy(...)` now uses the nested `Properties#getProperty` precedence and returns the parsed `OffsetResetStrategy`.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
jigar-bhati: Done. The dynamic warning now runs from `DynamicKafkaSourceBuilder#sanityCheck`; the enumerator only resolves the effective strategy and passes it to Kafka.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/KafkaPropertiesUtilTest.java
jigar-bhati: Done. The test is now parameterized over every pair from `OffsetResetStrategy.values()`; with the current enum values that covers all 3x3 combinations.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaPropertiesUtil.java
```diff
@@ -40,38 +44,56 @@ public static void copyProperties(@Nonnull Properties from, @Nonnull Properties
...
+                        globalProperties.getProperty(
+                                ConsumerConfig.AUTO_OFFSET_RESET_CONFIG,
+                                startingOffsetsInitializer
+                                        .getAutoOffsetResetStrategy()
+                                        .name())));
+    }
 
-        return startingOffsetsInitializer.getAutoOffsetResetStrategy().name().toLowerCase();
+    /** Parses the configured auto offset reset strategy. */
+    public static OffsetResetStrategy getResetStrategy(@Nonnull String offsetResetConfig) {
```
Efrat19, line 60: Can we now remove the duplicates in [KafkaDynamicSource](https://github.com/apache/flink-connector-kafka/blob/aa6be4bac8aeab15b471820f037fcf3cecca044c/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java#L505) and [DynamicKafkaTableSource](https://github.com/apache/flink-connector-kafka/blob/aa6be4bac8aeab15b471820f037fcf3cecca044c/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/DynamicKafkaTableSource.java#L513)

Efrat19 (commented):
LGTM w note

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaPropertiesUtil.java
jigar-bhati: Makes sense, done!

jigar-bhati:
@Efrat19 let me know if its ok to merge this! And what needs to be done - i believe you need to trigger ci check --> merge?

Efrat19 (approved):
LGTM, thanks

Efrat19:
> @Efrat19 let me know if its ok to merge this! And what needs to be done - i believe you need to trigger ci check --> merge?

I'm afraid I don't have permissions either

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSourceBuilder.java
```diff
@@ -468,10 +469,30 @@ private void parseAndSetRequiredProperties() {
             maybeOverride(KafkaSourceOptions.COMMIT_OFFSETS_ON_CHECKPOINT.key(), "false", false);
         }
         maybeOverride(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "false", false);
+        String configuredOffsetReset = props.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG);
+        if (configuredOffsetReset != null) {
+            OffsetResetStrategy configuredOffsetResetStrategy =
+                    KafkaPropertiesUtil.getResetStrategy(configuredOffsetReset);
```
bowenli86, line 475: This accepts values case-insensitively via `getResetStrategy(...)`, but leaves the original string in `props` because the later `maybeOverride(..., false)` preserves explicit values. For example, `auto.offset.reset=EARLIEST` builds successfully and then reaches Kafka unchanged, where the consumer only accepts lowercase `earliest`, `latest`, or `none`. Could we normalize the validated value here (or reject non-lowercase values early) and add a static-builder regression test, matching the Table path?

jigar-bhati: thanks, handled!

bowenli86 (approved):
lgtm, merging

PR #274, opened 2026-06-22
[FLINK-39944] Make dynamic Kafka metadata refresh resilient

lnbest0707:

`DynamicKafkaSource` schedules top-level stream metadata refresh through the same
`SplitEnumeratorContext.callAsync` worker path used by per-cluster
`KafkaSourceEnumerator` partition discovery. When metadata removes an unreachable cluster,
stale per-cluster Kafka Admin calls can occupy that shared worker until timeout, so the next
top-level metadata refresh may not run in time and readers can keep stale cluster metadata.

This change isolates dynamic stream metadata discovery on a dedicated stoppable worker and
marshals result handling back to the coordinator thread. Because metadata discovery and
stale-cluster activity checks can now run from different workers, metadata service calls are
serialized. Metadata updates are sent to readers before stale enumerator close waits complete,
and shutdown closes the metadata service before waiting for discovery so in-flight calls can
unblock. Asynchronous stale-enumerator close failures are retained and surfaced from close.

## Brief change log

- add an isolated metadata discovery context for top-level dynamic Kafka stream refresh
- serialize `KafkaMetadataService` calls across dynamic source workers
- mark stale sub-enumerator contexts closing before asynchronous close waits
- send reader metadata updates before stale enumerator close completes
- allow metadata service close to unblock in-flight discovery during source shutdown
- retain asynchronous stale-enumerator close failures until source close
- add regression coverage for isolated refresh, metadata update ordering, shutdown behavior, and
  metadata service serialization

## Verifying this change

```bash
DOCKER_HOST=unix://${HOME}/.colima/default/docker.sock \
TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE=/var/run/docker.sock \
TESTCONTAINERS_HOST_OVERRIDE=127.0.0.1 \
mvn -pl flink-connector-kafka \
  -Dtest=DynamicKafkaSourceEnumeratorTest,StoppableKafkaMetadataServiceDiscoveryContextTest,SynchronizedKafkaMetadataServiceTest \
  -Drat.skip=true test

DOCKER_HOST=unix://${HOME}/.colima/default/docker.sock \
TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE=/var/run/docker.sock \
TESTCONTAINERS_HOST_OVERRIDE=127.0.0.1 \
mvn -pl flink-connector-kafka -Drat.skip=true clean test
```

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
@@ -62,6 +62,9 @@ import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 /**
@@ -71,6 +74,7 @@ import java.util.stream.Collectors;
 @Internal
 public class DynamicKafkaSourceEnumerator
         implements SplitEnumerator<DynamicKafkaSourceSplit, DynamicKafkaSourceEnumState> {
+    private static final int ENUMERATOR_ADMIN_CLIENT_TIMEOUT_DIVISOR = 2;
     private static final Logger logger =
             LoggerFactory.getLogger(DynamicKafkaSourceEnumerator.class);
 
@@ -90,6 +94,9 @@ public class DynamicKafkaSourceEnumerator
     private final Boundedness boundedness;
     private final StoppableKafkaEnumContextProxy.StoppableKafkaEnumContextProxyFactory
             stoppableKafkaEnumContextProxyFactory;
+    private final StoppableKafkaMetadataServiceDiscoveryContext
+            kafkaMetadataServiceDiscoveryContext;
+    private final ExecutorService enumeratorClosingExecutor;
 
     // options
     private final long kafkaMetadataServiceDiscoveryIntervalMs;
@@ -123,7 +130,9 @@ public class DynamicKafkaSourceEnumerator
                 boundedness,
                 dynamicKafkaSourceEnumState,
                 StoppableKafkaEnumContextProxy.StoppableKafkaEnumContextProxyFactory
-                        .getDefaultFactory());
+                        .getDefaultFactory(),
+                StoppableKafkaMetadataServiceDiscoveryContext
+                        .StoppableKafkaMetadataServiceDiscoveryContextFactory.getDefaultFactory());
     }
 
     @VisibleForTesting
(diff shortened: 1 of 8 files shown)


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -754,6 +864,8 @@ public void handleSourceEvent(int subtaskId, SourceEvent sourceEvent) {
     @Override
     public void close() throws IOException {
         try {
+            kafkaMetadataServiceDiscoveryContext.close();
```
bowenli86, line 862: DynamicKafkaSourceEnumerator.close() waits for kafkaMetadataServiceDiscoveryContext.close() before closing kafkaMetadataService. Since the discovery context uses shutdownNow() and then waits indefinitely, a custom KafkaMetadataService with an in-flight metadata call that does not respond to interruption, or only unblocks when close() is called, could hang source coordinator shutdown.

Could we either close/unblock the metadata service before waiting for the discovery worker, avoid serializing close() behind active metadata calls, or use a bounded wait with explicit failure/logging?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -587,17 +653,38 @@ private void startAllEnumerators() {
...
+                    closingClusterEnumeratorMap) {
+        closingClusterEnumeratorMap.forEach(
                 (cluster, subEnumerator) -> {
                     try {
-                        clusterEnumContextMap.get(cluster).close();
+                        closingClusterEnumContextMap.get(cluster).close();
                         subEnumerator.close();
                     } catch (Exception e) {
-                        throw new RuntimeException(e);
+                        enumContext.runInCoordinatorThread(
```
bowenli86: The async stale-enumerator close path catches exceptions and posts a throwing runnable back through enumContext.runInCoordinatorThread(). During coordinator shutdown, that posted runnable can race with context shutdown and potentially be dropped/rejected, while DynamicKafkaSourceEnumerator.close() only waits for enumeratorClosingExecutor termination.

Could we retain the first async close failure, for example with a Future or AtomicReference, and rethrow or at least log it from close() after awaiting the close executor? That would make close failures observable even during shutdown.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -725,6 +812,29 @@ private Set<SplitAndAssignmentStatus> filterStateByTopics(
...
+    private void tuneEnumeratorAdminClientTimeout(Properties consumerProps, String propertyKey) {
+        String configuredTimeoutMs = consumerProps.getProperty(propertyKey);
+        if (configuredTimeoutMs == null) {
+            return;
+        }
+
+        long readerTimeoutMs = Long.parseLong(configuredTimeoutMs);
+        long enumeratorTimeoutMs =
+                Math.max(1L, readerTimeoutMs / ENUMERATOR_ADMIN_CLIENT_TIMEOUT_DIVISOR);
+        consumerProps.setProperty(propertyKey, Long.toString(enumeratorTimeoutMs));
```
bowenli86: The sub-enumerator timeout tuning silently rewrites `user/cluster-provided request.timeout.ms` and `default.api.timeout.ms` to half their configured values before constructing the Kafka sub-enumerator. I understand the mitigation goal, but this is a behavior change for jobs that intentionally configured those Kafka client timeouts.

Could we make this opt-in / dynamic-source-specific, or document why it is safe to override user-supplied Kafka timeout settings here?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
lnbest0707: Addressed in 8995be24. The async stale-enumerator close path now retains the first failure, still attempts normal coordinator-thread propagation, and rethrows the retained failure from close() after awaiting the close executor. Added a regression test where coordinator callbacks are dropped.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
lnbest0707: Addressed in 8995be24. I removed the timeout rewriting entirely. The isolated metadata discovery worker remains the correctness fix, while user and cluster request.timeout.ms and default.api.timeout.ms settings remain unchanged.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
lnbest0707: Addressed in 8995be24. Shutdown now marks discovery and sub-enumerator contexts closing, closes KafkaMetadataService before waiting for discovery termination, and leaves SynchronizedKafkaMetadataService.close() unsynchronized so delegate close can unblock an in-flight metadata call. Added regression coverage for that unblock path.

bowenli86:
lgtm, merging

bowenli86 (approved):

PR #276, opened 2026-06-24
[FLINK-39979] DynamicKafkaSource should mark reader idle when metadata removal leaves a subtask with no active subreaders

bowenli86:

`DynamicKafkaSource` can remain active after a metadata update removes the last active Kafka cluster/split from a reader subtask. Since no sub-reader remains to report idleness, downstream watermark progress can stall.

When the reader is already waiting on `isAvailable()`, resetting availability to zero sub-readers also leaves the cached availability future incomplete, so `pollNext()` is not invoked to release pending split outputs or mark the output idle.

## Brief change log

- Treat an actively consuming dynamic reader with zero active sub-readers as idle, and mark it active again when sub-readers return.
- Complete the previous availability future when metadata removal leaves the reader with zero active sub-readers.
- Add no-Docker regression coverage for removing the last active cluster/split while waiting on `isAvailable()`, completing availability, releasing the split output, and marking the reader idle.

## Verifying this change

- `mvn -f flink-connector-kafka/pom.xml -Dtest=DynamicKafkaSourceReaderIdlenessTest test`

- The serializers: no

## Documentation

---

- [x] Yes (OpenAI Codex)

Generated-by: OpenAI Codex

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
@@ -102,6 +102,7 @@ public class DynamicKafkaSourceReader<T> implements SourceReader<T, DynamicKafka
     private int availabilityHelperSize;
     private boolean isActivelyConsumingSplits;
     private boolean isNoMoreSplits;
+    private boolean dynamicOutputIdle;
     private AtomicBoolean restartingReaders;
     private ReaderOutput<T> latestReaderOutput;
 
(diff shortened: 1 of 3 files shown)


wenqiwu-oai (approved):

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
```diff
@@ -592,6 +595,12 @@ private void completeAndResetAvailabilityHelper() {
                         this.availabilityHelperSize = clusterReaderMap.size());
         syncAvailabilityHelperWithReaders();
 
+        if (clusterReaderMap.isEmpty()) {
```
lnbest0707: Checking only `clusterReaderMap.isEmpty()` may not be sufficient. IIUC, the sub-readers map is always created as long as the kafka metadata map is not empty.
The failing case described in the ticket is more related the number of active assigned splits. When the last active assigned split is removed from the reader, we should mark it idle. While in such case, this `clusterReaderMap` may not be empty and code would not fall in this branch.

bowenli86:
thanks for the review, addressed comments
merging

PR #277, opened 2026-06-24
[FLINK-39990] Expose DynamicKafkaSource active split count

lnbest0707:

`DynamicKafkaSource` currently exposes cluster- and split-scoped metrics whose metric groups can
disappear when dynamic metadata removes clusters or partitions. Consumers such as the Flink
Kubernetes Operator autoscaler need a stable per-reader signal for the current active split
assignment.

This change adds a lifetime `DynamicKafkaSource.activeSplitCount` gauge for each dynamic source
reader. It reports `0` before assignment and after all active splits are removed. The count is
derived only from splits assigned to active underlying `KafkaSourceReader`s, so pending restored
state and retained removed-cluster checkpoint offsets are excluded. The cached atomic value keeps
metric reporter threads from traversing mutable reader state.

## Brief change log

- register `DynamicKafkaSource.activeSplitCount` at the stable dynamic reader metric root
- refresh the count after split assignment, metadata rebuild/removal, polling, and reader close
- add coverage for empty readers, assignment, partial/all metadata removal, retained-state
  exclusion, and reactivation
- keep the existing cluster-metric lifecycle test scoped to cluster metrics
- document the gauge in the English and Chinese Dynamic Kafka source docs

## Verifying this change

```bash
DOCKER_HOST=unix://${HOME}/.colima/default/docker.sock \
TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE=/var/run/docker.sock \
TESTCONTAINERS_HOST_OVERRIDE=127.0.0.1 \
mvn -pl flink-connector-kafka \
  -Dtest=DynamicKafkaSourceReaderTest \
  -Drat.skip=true clean test

DOCKER_HOST=unix://${HOME}/.colima/default/docker.sock \
TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE=/var/run/docker.sock \
TESTCONTAINERS_HOST_OVERRIDE=127.0.0.1 \
mvn -pl flink-connector-kafka \
  -Dtest='DynamicKafkaSourceITTest$DynamicKafkaSourceSpecificTests#testMetricsLifecycleManagement' \
  -Drat.skip=true test

DOCKER_HOST=unix://${HOME}/.colima/default/docker.sock \
TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE=/var/run/docker.sock \
TESTCONTAINERS_HOST_OVERRIDE=127.0.0.1 \
mvn -pl flink-connector-kafka -Drat.skip=true clean test
```

### DIFF
--- a/docs/content.zh/docs/connectors/datastream/dynamic-kafka.md
+++ b/docs/content.zh/docs/connectors/datastream/dynamic-kafka.md
@@ -296,7 +296,7 @@ a list of applicable properties.
   </thead>
   <tbody>
     <tr>
-        <th rowspan="8">Operator</th>
+        <th rowspan="6">Operator</th>
         <td>currentEmitEventTimeLag</td>
         <td>n/a</td>
         <td>The time span from the record event timestamp to the time the record is emitted by the source connector¹: <code>currentEmitEventTimeLag = EmitTime - EventTime.</code></td>
@@ -324,6 +324,12 @@ a list of applicable properties.
       <td>kafkaClustersCount</td>
       <td>n/a</td>
       <td>The total number of Kafka clusters read by this reader.</td>
+      <td>Gauge</td>
+    </tr>
+    <tr>
+      <td>activeSplitCount</td>
+      <td>n/a</td>
+      <td>当前分配给此 source reader 的活跃 split 数。该 gauge 在 reader 的整个生命周期内保持注册；当元数据移除所有本地 split 后会报告 <code>0</code>。为已移除集群保留的 checkpoint offset 不计入其中。</td>
       <td>Gauge</td>
     </tr>
   </tbody>
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
@@ -69,6 +69,7 @@ import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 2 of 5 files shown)


flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/DynamicKafkaSourceITTest.java
```diff
@@ -1243,6 +1246,12 @@ private Set<String> findMetrics(InMemoryReporter inMemoryReporter, String groupP
                     .collect(Collectors.toSet());
         }
 
+        private Set<String> findKafkaClusterMetrics(InMemoryReporter inMemoryReporter) {
+            return findMetrics(inMemoryReporter, DYNAMIC_KAFKA_SOURCE_METRIC_GROUP).stream()
```
bowenli-oai: Could we avoid selecting a single arbitrary `DynamicKafkaSource` group here? This PR registers `activeSplitCount` directly on the parent group, so `findGroup(DYNAMIC_KAFKA_SOURCE_METRIC_GROUP)` can return that parent; `getMetricsByGroup()` then contains only the root gauge, and the `.kafkaCluster.` filter yields an empty set. That is consistent with the current 2.2.1/JDK 21 failure at line 1125. Please aggregate `findGroups(...)`/all matching groups or explicitly select kafka-cluster child groups. The same single-group pattern in `hasKafkaClusterMetrics()` can also make disappearance checks pass prematurely.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
```diff
@@ -630,6 +643,20 @@ private void closeAllReadersAndClearState() {
         }
         clusterReaderMap.clear();
         clustersProperties.clear();
+        activeSplitCount.set(0);
```
bowenli-oai: Can we avoid publishing `0` from the intermediate clear step? A normal metadata rebuild can preserve assigned splits, but metric reporters read this `AtomicInteger` concurrently and can sample `0` after this line before `refreshActiveSplitCount()` publishes the rebuilt count at line 357. For an autoscaler-facing stable signal, that transient is indistinguishable from “all local splits removed.” Since this method’s only caller refreshes after rebuilding, it seems safer to publish only the final count and keep the explicit zero in `close()`.

docs/content.zh/docs/connectors/datastream/dynamic-kafka.md
```diff
@@ -296,7 +296,7 @@ a list of applicable properties.
   </thead>
   <tbody>
     <tr>
-        <th rowspan="8">Operator</th>
+        <th rowspan="6">Operator</th>
```
bowenli-oai, line 299: what does it do?

lnbest0707: The old number is not correct, it had 5 before, and with the new one, 6 now.

bowenli86:
lgtm, merging

PR #278, opened 2026-06-24
[FLINK-39980] Implement split rebalance/reassignment support in DynamicKafkaSource after metadata changes

bowenli86:
## What is the purpose of the change

`DynamicKafkaSource` keeps the split ownership inherited from reader state after metadata
removes clusters or topics. In global mode, the remaining active splits can therefore stay skewed
across subtasks even when the number of active partitions is at least the source parallelism.

Flink 2.2.1 provides `SupportsSplitReassignmentOnRecovery` so a source can report restored reader
splits to its enumerator before assigning them. This change adopts that contract and lets
`DynamicKafkaSourceEnumerator` redistribute the active restored split set during recovery.

## Brief change log

- Make `DynamicKafkaSource` implement `SupportsSplitReassignmentOnRecovery`.
- Buffer reader-reported restored splits until all readers are registered and the first current
  metadata refresh has been applied.
- Requeue active reported splits through the existing Kafka sub-enumerators; global mode assigns
  deterministic round-robin recovery owners, while per-cluster mode keeps its existing owner
  selection semantics.
- Exclude removed topic splits from active rebalance, preserve retained removed-cluster offsets on
  their prior reader, and defer metadata events until recovery assignments are queued.
- Compile against Flink 2.2.1 because the split-reassignment marker interface and reported-splits
  reader API are unavailable in Flink 2.1.2.
- Document the global-mode recovery behavior and add no-Docker recovery regressions.

## Verifying this change

- `./mvnw -pl flink-connector-kafka -Dtest=DynamicKafkaSourceEnumeratorRecoveryTest,GlobalSplitOwnerAssignerTest test`
- `./mvnw -pl flink-connector-kafka -Dflink.version=2.3.0 -Dtest=DynamicKafkaSourceEnumeratorRecoveryTest test`
- `./mvnw -pl flink-connector-kafka -DskipTests verify`

The existing Docker-backed `DynamicKafkaSourceEnumeratorTest` was not runnable locally because
Testcontainers could not find a Docker environment.

## Does this pull request potentially affect one of the following parts:

- Dependencies (does it add or upgrade a dependency): yes, default Flink dependency 2.1.2 -> 2.2.1
- The public API, i.e., is any changed class annotated with `@Public(Evolving)`: yes, the source
  opts into the public evolving split-reassignment contract
- The serializers: no
- The runtime per-record code paths (performance sensitive): no
- Anything that affects deployment or recovery: JobManager (and its components), Checkpointing,
[...]

PR #279, opened 2026-06-25
[FLINK-39837] Reassign retained dynamic Kafka offsets upon readd

lnbest0707:
Jira: https://issues.apache.org/jira/browse/FLINK-39837

## Problem

Retained offsets must survive a cluster's removal and re-addition without preserving obsolete reader ownership. Reusing the old assignment can leave subtasks idle after rescaling. Reassigning without coordinating recovery can also let a failed former owner reactivate its stale retained copy while another reader is already consuming that split.

## Changes

- Live re-add of previously assigned retained splits collects retained offsets from all current readers, then waits for a completed checkpoint whose coordinator snapshot followed those reports before assigning the returning splits. Surviving active splits continue consuming during the handoff. Reader replacement or topology changes invalidate the current round; delayed responses cannot authorize a later round.
- Local reader recovery preserves current ownership and ignores another reader's stale retained copy. Full checkpoint restore reconciles all reader reports, prefers active progress over retained shadows, and restores returning clusters directly without another handoff checkpoint.
- Missing required unbounded offsets fail explicitly instead of falling back to an initializer. Completed bounded partitions remain recorded against rediscovery. Expired or disabled retention still uses the configured starting offsets for returning clusters.
- Coordinator-issued retention deadlines govern retained-state cleanup. A return accepted before expiry keeps its offsets while awaiting handoff completion; slow responses no longer trigger independent timeout/retry rounds. Startup failures cannot silently discard retained progress.
- GLOBAL placement assigns returning splits against current active load without moving surviving active splits. Retained-inventory indexing and skipping redundant reader rebuilds reduce metadata/recovery work.

## Dependency and limits

Depends on #295 at `173c31803bfae41b2013084ddc163a02cb905729`; the PR diff currently includes that prerequisite. This change retains #295's V4 checkpoint serializer and adds no further version bump. Full restore normalizes known retained reports for legacy compatibility; live and local recovery retain strict retention-epoch checks. V1–V4 compatibility tests and legacy deadline-skew restore tests pass.

Live handoff requires a completed checkpoint; jobs without periodic checkpoints must trigger one manually. Slow or repeatedly failing readers can delay completion. Returning splits can improve GLOBAL balance, but arbitrary imbalance cannot always be corrected without moving active splits.

## Validation

The final local clean module run passed **496 tests**, with zero failures, errors or skips, including **29 dynamic-source integration cases** and all ten committed-record runtime scenarios. Coverage includes old/new-owner local failures, checkpoint ordering, rescale, repeated retention cycles, expiry, disabled retention, bounded completion and serializer compatibility. Explicit local Checkstyle and Spotless checks passed on Java 17.0.17 / Flink 2.2.1.

Testing note: an earlier isolated run observed an unexplained extra healthy-reader restart without sufficient causal logs. Subsequent focused runs and the final clean suite passed; the initial restart has not been explained.


bowenli86 (commented):
lgtm, one minor comment

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -75,6 +78,8 @@
 @Internal
 public class DynamicKafkaSourceEnumerator
         implements SplitEnumerator<DynamicKafkaSourceSplit, DynamicKafkaSourceEnumState> {
+    private static final int ENUMERATOR_ADMIN_CLIENT_TIMEOUT_DIVISOR = 2;
```
bowenli86: minor: it's not used anywhere? remove it?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -796,26 +857,303 @@ private void retainRemovedClusterEnumeratorStates(
...
+            if (!activeTopics.contains(splitStatus.split().getTopic())) {
+                updatedSplits.add(splitStatus);
+                continue;
+            }
+            if (splitStatus.assignmentStatus().equals(AssignmentStatus.ASSIGNED)) {
+                Long retainedSplitOffset =
+                        retainedSplitOffsets.get(
+                                toDynamicSplitId(kafkaClusterId, splitStatus.split()));
+                if (retainedSplitOffset == null) {
+                    // No reader retains this offset anymore; let normal discovery recreate it.
```
Savonitar: 1. Is my understanding correct, in this case, it will be re-discovered as a "new" partition, so it comes back from newDiscoveryOffsetsInitializer (earliest) rather than the configured starting offset, e.g. a latest(), and source would rewind to the start of the topic.
Could this null branch fall back to the split's own retained offset instead of dropping it, so it resumes from the last known position?
2. In this case, will sink topic have duplicates even in EOS mode?

docs/content/docs/connectors/datastream/dynamic-kafka.md
```diff
@@ -242,6 +242,8 @@ By default, metadata removal also removes that cluster's split offsets from subs
 To keep removed cluster offsets available for a later re-add or restore, set
 `stream-metadata-removed-cluster-retention-ms` to a positive duration. For example,
 `604800000` retains removed cluster state for seven days before the source stops checkpointing it.
+If the cluster is re-added, the source uses the retained offsets but computes fresh reader
+assignments instead of reusing their previous owners.
```
Savonitar: should we mention this branch/reset behavior https://github.com/apache/flink-connector-kafka/pull/279/changes#r3505979200 which sounds like an exception? if we go with the suggested workflow.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -748,6 +796,19 @@ public void addReader(int subtaskId) {
...
                 (cluster, subEnumerator) -> subEnumerator.addReader(subtaskId));
+        if (!retainedSplitOffsetHandoffs.isEmpty()) {
+            pruneExpiredRetainedSplitOffsetHandoffs();
+            // A reader can join while a re-added cluster is waiting for offset handoff. Send
+            // metadata first so the reader has reconciled its local retained state before it
+            // answers the request. Restart the attempt so delayed responses from the reader's
+            // previous attempt cannot count as its replacement report.
+            retainedSplitOffsetHandoffs
+                    .values()
+                    .forEach(handoff -> handoff.offsetsByReader.clear());
```
Savonitar: Can I clarify, a single slow subtask that keeps going down and coming back up can repeatedly clear the other reader's reports and involve a delay?

Savonitar (commented):
Hi.
Thanks for the PR. I have a few questions to understand this behavior/intention better.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
lnbest0707: Yes, a repeatedly re-registering reader restarts the handoff. This is intentional: if a reader is unstable, we should not start the retained cluster from an incomplete or stale offset snapshot. Once all readers stabilize, the next handoff completes.
And normally, in such case, it usually means the slot itself is unhealthy and would anyway block the process.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
lnbest0707: Good catch! Yes, your analysis is correct, it would go back to `earliest` in such case. And just to be clear, this branch would rarely be hit when -- topic and cluster are retained in the enumerator checkpoint but missed in reader checkpoint. This would only happen when the checkpoint is somehow corrupted or in extreme time mismatch between TM/JM.
Falling completely back to `splitStatus.split()` is not safe because enumerator state only contains the original assigned starting offset, not the reader’s latest checkpointed progress. I’ll change this branch to keep the partition as an uninitialized/MIGRATED unassigned split so KafkaSourceEnumerator resolves it through the configured starting-offset initializer.
Once with the above source side patch, it should not be conflict with the sink behavior.

For the similar corner case, if reader has the retained offsets in checkpoint but enumerator does not have it, we would treat then as "new" cluster and use initializer to determine the starting point. That behavior should be correct as we should already pass the TTL in such case.

docs/content/docs/connectors/datastream/dynamic-kafka.md
lnbest0707: good point, will add

lnbest0707:
> Hi. Thanks for the PR. I have a few questions to understand this behavior/intention better.

Thanks for the review! Replied inline and also added a small fix & test for the comment.

lnbest0707:
@Savonitar pls let me know if you have any further comments/concerns on the change, thanks

Savonitar:
@lnbest0707 Sorry, I haven’t had a chance to review it again yet. I’m going to come back to this PR next week. Hope that’s fine.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -833,7 +944,11 @@ private void reassignReportedSplits() {
                 new TreeMap<>(pendingReportedSplitsByReader).entrySet()) {
             int readerId = readerSplits.getKey();
             for (DynamicKafkaSourceSplit split : readerSplits.getValue()) {
-                if (isSplitActive(split)) {
+                boolean retainedClusterWaitingForHandoff =
+                        split.isRetained()
+                                && retainedClusterEnumeratorStates.containsKey(
+                                        split.getKafkaClusterId());
+                if (!retainedClusterWaitingForHandoff && isSplitActive(split)) {
```
Savonitar: 1. Is my understanding correct that a reader drops its retained copy only when the same split is freshly assigned to that same reader? So after a re-add where ownership moved (e.g. rescale), a checkpoint can contain both the retained copy on reader#1 and the active copy on reader#2 until the retention expires? 

2. If the job restores from such checkpoint, both readers report the split here, and the retained copy takes this active branch because the cluster is no longer in `retainedClusterEnumeratorStates`. 
Wouldn't putIfAbsent then hit the duplicate and throw `IllegalStateException`?? 
Wouldn't the job be stuck in a restart loop?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
```diff
@@ -208,19 +213,38 @@ private InputStatus consolidateInputStatus(
...
+            if (split.isRetained()) {
+                if (removedClusterStateRetentionMs > 0
+                        && split.isRetained(currentTimeMillis)
+                        && !activeSplitIds.contains(split.splitId())) {
+                    retainedSplits.add(split);
+                }
+                continue;
+            }
+
+            retainedSplits.removeIf(
```
Savonitar, line 250: if the fresh assignment of this split will be assigned to a different reader, what will remove the retained copy from this reader before its retention expires? if nothing, is the recovery path prepared for the same split id appearing retained in one reader's checkpoint and active in another?

Savonitar (commented):
Hi. Thank you for addressing the feedback and answering questions.
I left a few questions. Could you please take a look?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
lnbest0707: Good catch, this was a real gap. Commit 2f729d9c keeps whether each recovery report was retained before clearing the marker. If the same split is reported as retained and active, recovery now prefers the active report; retained-only duplicates keep the furthest offset, while two active reports still fail. This prevents the stale retained copy from becoming a duplicate active owner and causing a restart loop. I also added testRecoveryPrefersActiveSplitOverRetainedDuplicate with retained offset 123 on reader 0 and active offset 456 on reader 1, asserting 456 wins.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
lnbest0707: Yes. When ownership moves, the old reader can keep its dormant retained copy until TTL; no reader-side assignment reaches that old reader. Commit 2f729d9c makes recovery explicitly tolerate that checkpoint shape: it discards the retained duplicate in favor of the active copy, so it is not reassigned and disappears from subsequent checkpoints. Two genuinely active owners still fail.

lnbest0707:
> Hi. Thank you for addressing the feedback and answering questions. I left a few questions. Could you please take a look?

Thanks for the review and catch. Resolved and replied.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
```diff
@@ -714,7 +716,12 @@ public void testEnumeratorStateRetainsRemovedClusterUntilExpired() throws Throwa
                                             DynamicKafkaSourceOptions
                                                     .STREAM_METADATA_REMOVED_CLUSTER_RETENTION_MS
                                                     .key(),
-                                            "1000");
+                                            "60000");
+                                    properties.setProperty(
+                                            DynamicKafkaSourceOptions.STREAM_ENUMERATOR_MODE.key(),
+                                            DynamicKafkaSourceOptions.EnumeratorMode.GLOBAL
```
Savonitar: Is my understanding correct that this test previously ran in the **default** enumerator mode and is now switched to `GLOBAL`? If so, and given that all the other new handoff tests also set `GLOBAL`, does that leave the default mode without any coverage of the new handoff path, and does this test lose the `PER_CLUSTER` coverage it used to provide?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -77,6 +80,7 @@
 @Internal
 public class DynamicKafkaSourceEnumerator
         implements SplitEnumerator<DynamicKafkaSourceSplit, DynamicKafkaSourceEnumState> {
+    private static final long RETAINED_SPLIT_OFFSET_HANDOFF_MIN_TIMEOUT_MS = 60_000L;
```
Savonitar: should it be configurable?

Savonitar (commented):
Hi. Thanks for addressing the feedback. I left a few non-blocking questions.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
lnbest0707: Good catch—addressed in 4dd586a0. The existing retained-cluster recovery test is now parameterized over both enumerator modes. For `PER_CLUSTER`, the mode property is intentionally left unset in both the initial and restored enumerators, so the test exercises the actual default. For `GLOBAL`, it preserves the global-only fresh-reader assignment assertion. The complete enumerator test suite passes: 33 tests, 0 failures.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
lnbest0707: This is already configurable by finally using max(metadata discovery interval, 60 seconds)

Savonitar:
Hi, thanks for working on this PR. I can continue the review after my vacation. If it’s a blocker, I’m fine with other reviewers approving it.

slachiewicz:
Heads-up on three smaller PRs in flight against the same files, so whoever lands second knows to rebase:

- #289 (approved) extracts the recovery-gating state out of `DynamicKafkaSourceEnumerator` into a new `ReaderRecoveryGate` class — most likely to collide with your +491/-46 in that file.
- #291 changes when `DynamicKafkaSourceReader.handleSourceEvents` re-delivers no-more-splits; different hunk from yours, same method.
- #285 is a two-line change in `DynamicKafkaSourceEnumeratorTest`.

All three are small, so rebasing them onto this is likely easier than the reverse.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -833,13 +944,33 @@ private void reassignReportedSplits() {
                 new TreeMap<>(pendingReportedSplitsByReader).entrySet()) {
             int readerId = readerSplits.getKey();
             for (DynamicKafkaSourceSplit split : readerSplits.getValue()) {
-                if (isSplitActive(split)) {
+                boolean retainedClusterWaitingForHandoff =
+                        split.isRetained()
+                                && retainedClusterEnumeratorStates.containsKey(
+                                        split.getKafkaClusterId());
+                if (!retainedClusterWaitingForHandoff && isSplitActive(split)) {
```
Savonitar: I have a question.
Say we have 1 partition, 2 subtasks, `GLOBAL` enumerator mode, 1 cluster. 
1. The cluster is removed while reader 1 owns the partition at offset 100, so reader 1 (and only reader 1) keeps a retained copy of the split in its state. 
2. The cluster is re-added, the offset handoff completes, ownership is reassigned from scratch, and the round-robin in `GlobalSplitOwnerAssigner` assigns the fresh split to reader 0 `(floorMod(0, 2) = 0). `
3. Reader 0 consumes on from 100, say to 150. Since `retainedSplits.removeIf` only runs on the reader that receives the fresh assignment, reader 1 still carries its **retained** copy(100) checkpointed. 
4. Now reader 1 alone fails over (e,g. task failure). On recovery it reports its restored state, which is just the retained copy at registration. 
5. Reader 0 was never unregistered, so `allReadersRegistered()` is satisfied and `reassignReportedSplits` runs with only reader 1's batch, then `clearRetention()` called, and goes through `addSplitsBack` (which also removes reader 0's entry from the sub-enumerator's `assignedSplits`) . And then assigned as a second active copy at offset 100 while reader 0 keeps consuming at 150. 

Is that correct, or am I missing a reconciliation/protection step somewhere?
If it is correct, then in my understanding EOS sink will commit duplicates and violate EOS guarantee.

Savonitar (commented):
Hi. Sorry for delayed response. I was on vacation.
I left 1 question-comment. If my understanding is correct, we may introduce duplicate producing which will affect EOS. 

p.s. During my vacation I also missed another PR https://github.com/apache/flink-connector-kafka/pull/278 . That's why I was curious and done a bit of deep dive into the Dynamic topic. It looks like we may have a dataloss for DynamicKafka workflow. Worth to check my [PR](https://github.com/apache/flink-connector-kafka/pull/295) (I know you contribute a lot to DynamicKafka, maybe you can review?) because it also fixes that EOS issue. It is not a duplicate PR, it solves a different issue.

slachiewicz:
One more to add to the coordination list from my earlier comment: #295 ([FLINK-40469](https://issues.apache.org/jira/browse/FLINK-40469)) checkpoints `pendingReportedSplitsByReader` into `DynamicKafkaSourceEnumState` (serializer bump to VERSION_4) and modifies `DynamicKafkaSourceEnumeratorRecoveryTest.java` — the same filename as the new recovery test class in this PR, in the same recovery-window area of `DynamicKafkaSourceEnumerator`.

Suggested landing order remains: the small approved ones first — #285, #288, #289 ([FLINK-40137](https://issues.apache.org/jira/browse/FLINK-40137), approved) — then this PR and #295 rebase on top; #289 in particular touches the same recovery-gating state #295 checkpoints.

lnbest0707:
> Hi. Sorry for delayed response. I was on vacation. I left 1 question-comment. If my understanding is correct, we may introduce duplicate producing which will affect EOS.
> 
> p.s. During my vacation I also missed another PR #278 . That's why I was curious and done a bit of deep dive into the Dynamic topic. It looks like we may have a dataloss for DynamicKafka workflow. Worth to check my [PR](https://github.com/apache/flink-connector-kafka/pull/295) (I know you contribute a lot to DynamicKafka, maybe you can review?) because it also fixes that EOS issue. It is not a duplicate PR, it solves a different issue.

The https://github.com/apache/flink-connector-kafka/pull/295 looks valid. And your question also makes sense. The approach in the PR has been tuned a few times due to various corner cases. Let me also rethink about whether we can get some better approaches to make it simpler and catch all issues.

lnbest0707:
> > Hi. Sorry for delayed response. I was on vacation. I left 1 question-comment. If my understanding is correct, we may introduce duplicate producing which will affect EOS.
> > p.s. During my vacation I also missed another PR #278 . That's why I was curious and done a bit of deep dive into the Dynamic topic. It looks like we may have a dataloss for DynamicKafka workflow. Worth to check my [PR](https://github.com/apache/flink-connector-kafka/pull/295) (I know you contribute a lot to DynamicKafka, maybe you can review?) because it also fixes that EOS issue. It is not a duplicate PR, it solves a different issue.
> 
> The #295 looks valid. And your question also makes sense. The approach in the PR has been tuned a few times due to various corner cases. Let me also rethink about whether we can get some better approaches to make it simpler and catch all issues.

The updated method is by holding the new cluster consumption till the next checkpoint to avoid all those corner cases. I also tried another approach by forcing a full rebalance/restart upon the events. The alternative method also works but is more complicated and more interruptive on the jobs. So IMO, the current one might be the path to go. @Savonitar WDYT?

PR #282, opened 2026-06-28
[FLINK-39193][Kafka Connector] Fix Kafka client compatibility for committed offset lookup

qiuyanjun888:

FLINK-39193 reports a `NoClassDefFoundError` for `org/apache/kafka/clients/admin/ListConsumerGroupOffsetsSpec` when an incomplete specific-offset map causes the Kafka source to look up committed offsets.

The committed-offset lookup started referencing `ListConsumerGroupOffsetsSpec` when the connector upgraded to Kafka 4.0. That class and the map-based `listConsumerGroupOffsets(...)` overload are not available in Kafka clients before 3.3, so a runtime classpath containing an older Kafka client can fail before the lookup is executed.

This PR now fixes that library compatibility issue without changing offset-initialization semantics.

## Brief change log

- Use the long-standing `AdminClient#listConsumerGroupOffsets(String)` overload instead of directly referencing `ListConsumerGroupOffsetsSpec`.
- Filter the returned consumer-group offsets to the partitions requested by the initializer.
- Ignore missing or null committed offsets, preserving the existing fallback behavior.
- Add a focused unit test for requested-partition filtering.
- Restore the existing specified-offset behavior: missing specified offsets still prefer committed offsets before applying the configured reset strategy.

The stable overload retrieves all committed offsets for the consumer group and filters them client-side. This is a deliberate compatibility trade-off that avoids reflection and keeps the production code linkable with older Kafka client runtimes.

## Verifying this change

- Verified with actual `kafka-clients` JARs that:
  - `ListConsumerGroupOffsetsSpec` is absent through 3.2.3 and present from 3.3.0.
  - `AdminClient#listConsumerGroupOffsets(String)` and `ListConsumerGroupOffsetsResult#partitionsToOffsetAndMetadata()` remain linkable across representative Kafka client versions from 2.1.1 through 4.2.x.
- Reproduced the original `NoClassDefFoundError` by compiling the map/spec call against Kafka 4.2.0 and running it with Kafka 2.3.1.
- Confirmed the stable string-overload probe links with representative Kafka client versions from 2.1.1 through 4.2.0.
- Focused regression test:
  - `mvn -pl flink-connector-kafka -Dtest=PartitionOffsetsRetrieverImplTest -DfailIfNoTests=false test`
  - `Tests run: 1, Failures: 0, Errors: 0, Skipped: 0`
  - Checkstyle: 0 violations
  - Spotless: 0 files need changes
- Confirmed the compiled `PartitionOffsetsRetrieverImpl` bytecode no longer references `ListConsumerGroupOffsetsSpec`.

Docker-backed Kafka integration tests were not run locally because this environment has no Docker daemon.

## AI Assistance

This pull request was prepared with AI assistance under human-directed scope for FLINK-39193. The root-cause analysis, compatibility matrix, code changes, focused tests, and final submission were reviewed and directed by the contributor.

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
@@ -32,10 +32,10 @@ import org.apache.flink.util.FlinkRuntimeException;
 
 import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.KafkaAdminClient;
-import org.apache.kafka.clients.admin.ListConsumerGroupOffsetsSpec;
 import org.apache.kafka.clients.admin.ListOffsetsResult;
 import org.apache.kafka.clients.admin.OffsetSpec;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.clients.consumer.OffsetAndTimestamp;
 import org.apache.kafka.common.TopicPartition;
 import org.slf4j.Logger;
@@ -685,23 +685,11 @@ public class KafkaSourceEnumerator
 
         @Override
         public Map<TopicPartition, Long> committedOffsets(Collection<TopicPartition> partitions) {
-            ListConsumerGroupOffsetsSpec offsetsSpec =
-                    new ListConsumerGroupOffsetsSpec().topicPartitions(partitions);
             try {
                 return adminClient
-                        .listConsumerGroupOffsets(Collections.singletonMap(groupId, offsetsSpec))
+                        .listConsumerGroupOffsets(groupId)
                         .partitionsToOffsetAndMetadata()
-                        .thenApply(
-                                result -> {
-                                    Map<TopicPartition, Long> offsets = new HashMap<>();
-                                    result.forEach(
-                                            (tp, oam) -> {
-                                                if (oam != null) {
-                                                    offsets.put(tp, oam.offset());
-                                                }
-                                            });
-                                    return offsets;
-                                })
+                        .thenApply(result -> filterCommittedOffsets(partitions, result))
                         .get();
             } catch (InterruptedException e) {
                 Thread.currentThread().interrupt();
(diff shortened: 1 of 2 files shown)


flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/enumerator/initializer/SpecifiedOffsetsInitializerTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+/** Unit tests for {@link SpecifiedOffsetsInitializer}. */
+class SpecifiedOffsetsInitializerTest {
+
+    @Test
+    void incompleteSpecificOffsetsUsesResetStrategyWithoutCommittedOffsetLookup() {
+        TopicPartition specifiedPartition = new TopicPartition("topic", 0);
+        TopicPartition unspecifiedPartition = new TopicPartition("topic", 1);
+        OffsetsInitializer initializer =
+                OffsetsInitializer.offsets(Map.of(specifiedPartition, 111L));
```
Efrat19: Nit:
```suggestion
                OffsetsInitializer.offsets(Map.of(specifiedPartition, expectedSpecifiedOffset));
```

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/enumerator/initializer/SpecifiedOffsetsInitializerTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+        Map<TopicPartition, Long> offsets =
+                initializer.getPartitionOffsets(
+                        List.of(specifiedPartition, unspecifiedPartition),
+                        new OffsetsInitializer.PartitionOffsetsRetriever() {
```
Efrat19: Nit: Create a static class?

Efrat19 (commented):
+1 on the fix, I think this matches user expectation for `specifiedOffsetInitializer`.
LGTM % test readability could be improved

bowenli86:
thanks for the contribution

I opened FLINK-39193 mainly for fixing the lib incompatibility error of `java.lang.NoClassDefFoundError: org/apache/kafka/clients/admin/ListConsumerGroupOffsetsSpec`

I think your PR can be on another ticket to update the behavior change?

qiuyanjun888:
@bowenli86 Thanks for the clarification. I re-investigated the failure and updated this PR to focus on the library compatibility issue in FLINK-39193.

The root cause is the direct `ListConsumerGroupOffsetsSpec` reference introduced with the Kafka 4.0 client upgrade. The class and map-based overload are absent in Kafka clients before 3.3, which reproduces the reported `NoClassDefFoundError` when an older client is present at runtime.

The PR now uses the long-standing `AdminClient#listConsumerGroupOffsets(String)` overload and filters the returned offsets to the requested partitions. I removed the previous behavior-change approach, so incomplete specified offsets continue to prefer committed offsets before falling back to the configured reset strategy.

I also added a focused filtering regression test and verified the relevant API linkage with representative Kafka client JARs from 2.1.1 through 4.2.x. The focused Maven test, checkstyle, and spotless checks pass locally.

Efrat19:
> The root cause is the direct ListConsumerGroupOffsetsSpec reference introduced with the Kafka 4.0 client upgrade. The class and map-based overload are absent in Kafka clients before 3.3, which reproduces the reported NoClassDefFoundError when an older client is present at runtime.

Thank you for the change.
Could you please clarify why should we still maintain <3.3 kafka client support in the main branch?

PR #283, opened 2026-07-07
[FLINK-40094][Connectors/Kafka] Mark removed DynamicKafkaSource split outputs idle before release

lnbest0707:

This implements [FLINK-40094](https://issues.apache.org/jira/browse/FLINK-40094) as the mandatory correctness follow-up to [FLINK-39979](https://issues.apache.org/jira/browse/FLINK-39979) and [#276](https://github.com/apache/flink-connector-kafka/pull/276).

#276 correctly marks the `DynamicKafkaSource` reader's main output idle when metadata removal leaves a source subtask without active subreaders. Flink also maintains split-local outputs for source watermark aggregation, however, and directly calling `releaseOutputForSplit` does not recompute or publish their aggregate watermark/idleness. A removed split that was active can therefore leave the subtask output active and pin the downstream watermark even after the main reader output becomes idle.

This change marks each removed split output idle immediately before releasing it. That forces Flink to publish the updated aggregate state before the split-local output is unregistered.

## Brief change log

- Mark removed split outputs idle before both immediate and deferred release.
- Synchronize the reader-level idle state so a replacement split is reactivated before polling.
- Add no-Docker regression coverage for immediate/deferred release, `idle` then `release`
  ordering, and split reassignment before the next poll.

## Verifying this change

- `mvn -pl flink-connector-kafka -Dtest=DynamicKafkaSourceReaderIdlenessTest -DskipITs -Drat.skip=true test`
- `mvn -pl flink-connector-kafka -Drat.skip=true clean test`

- The serializers: no

## Documentation

---

- [x] Yes (OpenAI Codex)

Generated-by: OpenAI Codex

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
@@ -395,7 +395,7 @@ public class DynamicKafkaSourceReader<T> implements SourceReader<T, DynamicKafka
         if (latestReaderOutput == null) {
             pendingSplitOutputReleases.add(splitId);
         } else {
-            latestReaderOutput.releaseOutputForSplit(splitId);
+            markSplitIdleAndReleaseOutput(latestReaderOutput, splitId);
         }
     }
 
@@ -405,9 +405,16 @@ public class DynamicKafkaSourceReader<T> implements SourceReader<T, DynamicKafka
         }
 
         for (String splitId : pendingSplitOutputReleases) {
-            readerOutput.releaseOutputForSplit(splitId);
+            markSplitIdleAndReleaseOutput(readerOutput, splitId);
         }
         pendingSplitOutputReleases.clear();
+    }
+
+    private void markSplitIdleAndReleaseOutput(ReaderOutput<T> readerOutput, String splitId) {
+        // Unregistering a split output does not recompute Flink's split-local watermark or
+        // idleness. Mark it idle first so a removed split cannot keep the source output active.
+        readerOutput.createOutputForSplit(splitId).markIdle();
+        readerOutput.releaseOutputForSplit(splitId);
     }
 
     private static boolean isSplitForActiveClusters(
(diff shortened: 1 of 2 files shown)


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
```diff
@@ -405,11 +405,18 @@ private void releasePendingSplitOutputs(ReaderOutput<T> readerOutput) {
...
-            readerOutput.releaseOutputForSplit(splitId);
+            markSplitIdleAndReleaseOutput(readerOutput, splitId);
         }
         pendingSplitOutputReleases.clear();
     }
 
+    private void markSplitIdleAndReleaseOutput(ReaderOutput<T> readerOutput, String splitId) {
+        // Unregistering a split output does not recompute Flink's split-local watermark or
+        // idleness. Mark it idle first so a removed split cannot keep the source output active.
+        readerOutput.createOutputForSplit(splitId).markIdle();
```
bowenli86, line 422: Could we add a regression test for the case where the last split is removed and then a split is re-added before the next `pollNext()`? In that sequence, this `markIdle()` may already expose the source as idle while `dynamicOutputIdle` remains `false`, so the active path may not call `markActive()` again. I wonder if explicitly reactivating on split re-add, or covering this ordering in a test, would help avoid records flowing while downstream still sees the source as idle.

lnbest0707: good catch! Added a fix with a test to catch it.

lnbest0707:
The Java 17 CI failure is unrelated to this change: all connector modules completed, then `flink-connector-kafka-python` failed while Maven Wagon downloaded `install_command.sh` with `Attempted read from closed stream`. Matrix fail-fast subsequently canceled the Java 11 and 21 jobs; all standalone Python jobs passed. I do not have repository permission to rerun the workflow. @bowenli86, could you please rerun the failed jobs?

bowenli86:
lgtm, merging

PR #285, opened 2026-07-12
[FLINK-40136][Connectors/Kafka] Fix flaky ConcurrentModificationException in DynamicKafkaSourceEnumeratorTest

slachiewicz:

`DynamicKafkaSourceEnumeratorTest` fails intermittently with `ConcurrentModificationException` instead of the `AssertionError` that its polling loop expects and retries on ([FLINK-40136](https://issues.apache.org/jira/browse/FLINK-40136)).

## Brief change log

- In `MockSplitEnumeratorContext`, the inner `List<SourceEvent>` instances of `sentSourceEvent` are shared, and the main executor appends to them while `hasLatestMetadataUpdateEvent()` streams them on the test thread ([FLINK-40543](https://issues.apache.org/jira/browse/FLINK-40543)). That interleaving throws `ConcurrentModificationException`.
- Catch `ConcurrentModificationException` alongside `AssertionError` in `hasLatestMetadataUpdateEvent()`, so the existing `CommonTestUtils.waitUtil(...)` polling loop retries instead of failing the test.
- Keep the reflection in `getLatestMetadataUpdateEventWithoutContextSync()`. The public `MockSplitEnumeratorContext.getSentSourceEvent()` dispatches to `workerExecutor` and calls `Future.get()`, so a test that blocks `workerExecutor` on purpose (for example, `testProductionMetadataRefreshBypassesBlockedSourceCoordinatorAsyncCallable`) would deadlock.

## Verifying this change

This change is test-only.

Verified on the rebased branch: `mvn -pl flink-connector-kafka test -Dtest=DynamicKafkaSourceEnumeratorTest` → 30 tests, 0 failures, 14.8 s, including `testProductionMetadataRefreshBypassesBlockedSourceCoordinatorAsyncCallable` without hanging. The original failure is intermittent, so one green run does not prove it gone; CI over time is the stronger signal.

  - The serializers: no

## Documentation

---

Generated-by: Claude Code (Claude Fable 5)

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
@@ -385,9 +385,14 @@ public class DynamicKafkaSourceReader<T> implements SourceReader<T, DynamicKafka
 
             addSplits(validPendingSplits);
             pendingSplits.clear();
-            if (isNoMoreSplits) {
-                notifyNoMoreSplits();
-            }
+        }
+
+        // Re-deliver a no-more-splits signal that arrived before this metadata update, since
+        // sub-readers created above have not seen it. This must also happen when there are no
+        // pending splits at all: an idle reader that registers after all bounded sub-enumerators
+        // finished discovery receives no-more-splits first and would otherwise never finish.
+        if (isNoMoreSplits) {
+            notifyNoMoreSplits();
         }
     }
 
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
@@ -61,6 +61,7 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.ConcurrentModificationException;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -2438,7 +2439,7 @@ public class DynamicKafkaSourceEnumeratorTest {
             return getLatestMetadataUpdateEventWithoutContextSync(context, readerId)
                     .getKafkaStreams()
                     .equals(Collections.singleton(expectedKafkaStream));
-        } catch (AssertionError e) {
+        } catch (AssertionError | ConcurrentModificationException e) {
             return false;
         }
     }
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
@@ -61,6 +61,7 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.ConcurrentModificationException;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -2411,6 +2412,11 @@ public class DynamicKafkaSourceEnumeratorTest {
     @SuppressWarnings("unchecked")
     private MetadataUpdateEvent getLatestMetadataUpdateEventWithoutContextSync(
             MockSplitEnumeratorContext<DynamicKafkaSourceSplit> context, int readerId) {
+        // Reflection is required here because MockSplitEnumeratorContext#getSentSourceEvent()
+        // dispatches to workerExecutor and blocks on Future#get(). In tests that intentionally
+        // block workerExecutor (e.g.
+        // testProductionMetadataRefreshBypassesBlockedSourceCoordinatorAsyncCallable),
+        // calling getSentSourceEvent() deadlocks.
         Map<Integer, List<SourceEvent>> sentSourceEvents =
                 (Map<Integer, List<SourceEvent>>)
                         Whitebox.getInternalState(context, "sentSourceEvent");
@@ -2430,6 +2436,9 @@ public class DynamicKafkaSourceEnumeratorTest {
                                                 readerId)));
     }
 
+    // Polling predicate: retries on AssertionError ("not ready yet") as well as
+    // ConcurrentModificationException because sentSourceEvent's inner event lists are modified
+    // concurrently by the main executor thread (see FLINK-40543).
     private boolean hasLatestMetadataUpdateEvent(
             MockSplitEnumeratorContext<DynamicKafkaSourceSplit> context,
             int readerId,
@@ -2438,7 +2447,7 @@ public class DynamicKafkaSourceEnumeratorTest {
             return getLatestMetadataUpdateEventWithoutContextSync(context, readerId)
                     .getKafkaStreams()
                     .equals(Collections.singleton(expectedKafkaStream));
-        } catch (AssertionError e) {
+        } catch (AssertionError | ConcurrentModificationException e) {
             return false;
         }
     }
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReaderTest.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReaderTest.java
@@ -370,6 +370,34 @@ public class DynamicKafkaSourceReaderTest extends SourceReaderTestBase<DynamicKa
     }
 
     @Test
+    void testIdleReaderFinishesWhenNoMoreSplitsArrivesBeforeMetadata() throws Exception {
+        TestingReaderContext context = new TestingReaderContext();
+        try (DynamicKafkaSourceReader<Integer> reader = createReaderWithoutStart(context)) {
+            TrackingReaderOutput<Integer> readerOutput = new TrackingReaderOutput<>();
+            reader.start();
+
+            // The enumerator signals no more splits before the reader receives the metadata
+            // update event, e.g. when an idle reader registers after all bounded
+            // sub-enumerators have already finished split discovery.
+            reader.notifyNoMoreSplits();
+
+            MetadataUpdateEvent metadata =
+                    DynamicKafkaSourceTestHelper.getMetadataUpdateEvent(TOPIC);
+            reader.handleSourceEvents(metadata);
+
+            long deadline = System.currentTimeMillis() + 10_000;
+            InputStatus status = reader.pollNext(readerOutput);
+            while (status != InputStatus.END_OF_INPUT && System.currentTimeMillis() < deadline) {
+                status = reader.pollNext(readerOutput);
+            }
+            assertThat(status)
+                    .as(
+                            "idle reader must reach END_OF_INPUT even when no-more-splits precedes the metadata update event")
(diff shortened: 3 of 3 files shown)


flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
```diff
@@ -2440,6 +2441,13 @@ private boolean hasLatestMetadataUpdateEvent(
                     .equals(Collections.singleton(expectedKafkaStream));
         } catch (AssertionError e) {
             return false;
+        } catch (ConcurrentModificationException e) {
```
spuru9: Might just be one multi-catch instead of duplicating return false.

```suggestion
        } catch (AssertionError | ConcurrentModificationException e) {
```

spuru9:
@slachiewicz I dont think it anymore falls in the domain of a hotfix. Can you file a JIRA for this. Also can you rework on the PR description as per the guidlines.

slachiewicz:
@spuru9 Filed https://issues.apache.org/jira/browse/FLINK-40136 and updated the PR title/description accordingly.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
```diff
@@ -385,9 +385,14 @@ public void handleSourceEvents(SourceEvent sourceEvent) {
...
-            if (isNoMoreSplits) {
-                notifyNoMoreSplits();
-            }
+        }
+
+        // Re-deliver a no-more-splits signal that arrived before this metadata update, since
+        // sub-readers created above have not seen it. This must also happen when there are no
+        // pending splits at all: an idle reader that registers after all bounded sub-enumerators
+        // finished discovery receives no-more-splits first and would otherwise never finish.
+        if (isNoMoreSplits) {
```
Efrat19: It seems like this fix addresses a different issue than the one described in the ticket?
were you able to reproduce this in actual env / ITCase?
Otherwise we should probably avoid this cross components notification on the checkpoint interval 🤔

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReaderTest.java
```diff
@@ -369,6 +369,34 @@ void testNotifyNoMoreSplits() throws Exception {
         }
     }
 
+    @Test
+    void testIdleReaderFinishesWhenNoMoreSplitsArrivesBeforeMetadata() throws Exception {
```
Efrat19: ditto - maybe consider a different ticket

Efrat19 (commented):
Hi, thank you for the fix

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
slachiewicz: You were right — moved to #291 under FLINK-40362, where the reproduction and the notification-frequency question are answered.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReaderTest.java
slachiewicz: You were right — moved to #291 under FLINK-40362, where the reproduction and the notification-frequency question are answered.

slachiewicz:
The reader fix that had been sitting on this branch is now #291 (FLINK-40362). #285 is back to the two-line test fix it is titled after, rebased on main.

Note that CI here can still hit the `testIdleReader` hang until #291 lands, since main keeps the guard that causes it.

slachiewicz:
@spuru9 this is ready for another look — FLINK-40136 is filed with the description reworked, your multi-catch suggestion is applied, and the unrelated reader fix that had been sitting on the branch is now #291. What is left here is the two-line test change.

Efrat19 (commented):
Hey @slachiewicz thanks for splitting the pr. 
Rereading the ticket, I actually don't understand why do we need reflection at all?
Instead of 
```
    private MetadataUpdateEvent getLatestMetadataUpdateEventWithoutContextSync(
            MockSplitEnumeratorContext<DynamicKafkaSourceSplit> context, int readerId) {
        Map<Integer, List<SourceEvent>> sentSourceEvents =
                (Map<Integer, List<SourceEvent>>)
                        Whitebox.getInternalState(context, "sentSourceEvent");
```
Can't we do
```
    private MetadataUpdateEvent getLatestMetadataUpdateEventWithoutContextSync(
            MockSplitEnumeratorContext<DynamicKafkaSourceSplit> context, int readerId) {
        Map<Integer, List<SourceEvent>> sentSourceEvents = context.getSentSourceEvent();
```
Should probably solve the `ConcurrentModificationException` as well

Efrat19 (changes requested):

slachiewicz:
@Efrat19 good catch on the reflection — `getSentSourceEvent()` is public and I will switch to it.

It will not remove the `ConcurrentModificationException` though. In `MockSplitEnumeratorContext`, `getSentSourceEvent()` copies on `workerExecutor` while `sendEventToSourceReader` writes on `mainExecutor`, and those are two separate single-thread executors created in the constructor, so the copy is not serialised against the writer. The copy is also shallow: `new HashMap<>(sentSourceEvent)` shares the `List<SourceEvent>` values, and it is that inner list this helper streams — which is where the CME actually comes from.

So I would like to keep the catch alongside the accessor change. `hasLatestMetadataUpdateEvent` is a poll predicate, so returning false and retrying is the right response to a mid-write read.

Removing the race properly means changing `MockSplitEnumeratorContext` in flink-core to copy on the main executor and deep-copy the value lists. Happy to file that separately if you think it is worth doing.

Efrat19:
> getSentSourceEvent() is public and I will switch to it.

Thanks @slachiewicz  - please update commit description accordingly

> Removing the race properly means changing `MockSplitEnumeratorContext` in [flink-core](https://issues.apache.org/jira/browse/FLINK-core) to copy on the main executor and deep-copy the value lists. Happy to file that separately if you think it is worth doing.

Sounds like a simple fix through `ConcurrentHashMap` - can we at least have a ticket?


> It will not remove the ConcurrentModificationException though

Understood, would be nice to link to the ticket for the flink-core fix

slachiewicz:
@Efrat19 Done:
- Switched `getLatestMetadataUpdateEventWithoutContextSync()` from reflection to `context.getSentSourceEvent()`, removing `@SuppressWarnings("unchecked")`.
- `hasLatestMetadataUpdateEvent()` now catches `ConcurrentModificationException` directly as well as unwrapped from `ExecutionException` (in case CME occurs during `getSentSourceEvent()`'s `workerExecutor` copy).
- Updated commit message and PR description.

Regarding the `MockSplitEnumeratorContext` ticket for `flink-core`: I have prepared the ticket details for `flink-core` to track synchronizing the copy on `mainExecutor` and deep-copying event lists, and will link it once filed.

slachiewicz:
@Efrat19 I have created the flink-core ticket to track the race condition in `MockSplitEnumeratorContext`: https://issues.apache.org/jira/browse/FLINK-40543 and linked it in the PR description and code comment.

Efrat19:
> @Efrat19 I have created the [flink-core](https://issues.apache.org/jira/browse/FLINK-core) ticket to track the race condition in `MockSplitEnumeratorContext`: https://issues.apache.org/jira/browse/FLINK-40543 and linked it in the PR description and code comment.
> 
> _This comment was created with AI assistance._

Thank you!
Is the `getSentSourceEvent` call causing a CI deadlock now?

slachiewicz:
@Efrat19 Yes, that is exactly what happened.

### Root Cause Analysis
In `MockSplitEnumeratorContext`, `getSentSourceEvent()` is implemented by submitting a task to `workerExecutor`:
```java
workerExecutor.submit(() -> new HashMap<>(sentSourceEvent)).get();
```
`workerExecutor` is a single-threaded executor.

In `DynamicKafkaSourceEnumeratorTest#testProductionMetadataRefreshBypassesBlockedSourceCoordinatorAsyncCallable`, the test deliberately blocks `workerExecutor` with an async callable waiting on a `CountDownLatch` (`allowSourceCoordinatorCallableToFinish`), specifically to assert that metadata discovery bypasses a blocked coordinator callable.

When `hasLatestMetadataUpdateEvent()` called `context.getSentSourceEvent()`, it queued onto the blocked `workerExecutor` and waited indefinitely on `FutureTask.get()`, causing the 50-minute CI timeout.

### Resolution
In `5df47434`:
- Retained reflection via `Whitebox.getInternalState(context, "sentSourceEvent")` in `getLatestMetadataUpdateEventWithoutContextSync()` with an explanatory comment documenting why `getSentSourceEvent()` deadlocks here.
- Kept the `catch (AssertionError | ConcurrentModificationException e)` in `hasLatestMetadataUpdateEvent()` so concurrent modifications during polling retry cleanly.

Verified locally: `testProductionMetadataRefreshBypassesBlockedSourceCoordinatorAsyncCallable` now completes in ~17 seconds without hanging.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
```diff
@@ -2411,6 +2412,11 @@ private MetadataUpdateEvent getLatestMetadataUpdateEvent(
     @SuppressWarnings("unchecked")
     private MetadataUpdateEvent getLatestMetadataUpdateEventWithoutContextSync(
             MockSplitEnumeratorContext<DynamicKafkaSourceSplit> context, int readerId) {
+        // Reflection is required here because MockSplitEnumeratorContext#getSentSourceEvent()
+        // dispatches to workerExecutor and blocks on Future#get(). In tests that intentionally
+        // block workerExecutor (e.g.
+        // testProductionMetadataRefreshBypassesBlockedSourceCoordinatorAsyncCallable),
+        // calling getSentSourceEvent() deadlocks.
```
Efrat19, line 2419: Will this also be solved by FLINK-40543?

Efrat19 (approved):

Savonitar:
Hi. Thanks for the PR!
I see it is "DRAFT". Could you please clarify, Is it still Draft or ready for a review (I see there is an approval already)?
The flaky issue blocks some other PRs.

slachiewicz:
Ready for review now, marked as such. Approved by @Efrat19 on 09-15; CI green on the 09-11 rebase.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
slachiewicz: Partly. FLINK-40543 (still open, Qiu Yanjun asked to take it on 09-13) removes the `ConcurrentModificationException`, so the catch in `hasLatestMetadataUpdateEvent()` becomes redundant. The reflection here only goes away if that fix also stops `getSentSourceEvent()` blocking on `workerExecutor`, which the ticket recommends but does not require. Either way the connector builds against the lowest supported `flink.version`, so both stay until that minimum carries the fix.

MartijnVisser (approved):
Reproduced the mechanism locally with a forced appender on `sentSourceEvent`: the old predicate lets the `ConcurrentModificationException` escape `waitUtil`, and this one retries until the expected event arrives. 1000 runs of the test itself pass on both main and this branch, so that alone proves little. Approving; I'll merge.

PR #289, opened 2026-07-13
[FLINK-40137][Connectors/Kafka] Extract ReaderRecoveryGate from DynamicKafkaSourceEnumerator

slachiewicz:

`DynamicKafkaSourceEnumerator` accumulated three interacting pieces of recovery-gating state (initial registration pending, reported splits pending redistribution, deferred metadata-update readers) inline, making the gating semantics hard to follow. This change extracts that state and its transitions into a dedicated `ReaderRecoveryGate` class so the gating state and its lifecycle have an explicit name and structure.

## Brief change log

- Extract the recovery-gating state and its logic into a dedicated `ReaderRecoveryGate` class.
- Add `ReaderRecoveryGateTest` covering the gating state transitions and their interaction directly in isolation.
- No behavior change to `DynamicKafkaSourceEnumerator`.

## Verifying this change

This change is covered by the new `ReaderRecoveryGateTest` unit test suite, which exercises the extracted gating semantics directly in isolation. Existing `DynamicKafkaSourceEnumeratorTest` and `DynamicKafkaSourceEnumeratorRecoveryTest` test suites continue to pass unchanged, confirming no behavior change.

  - The serializers: no
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Kafka connector, etc: yes, touches `DynamicKafkaSourceEnumerator` recovery gating logic, but this is a pure refactor with no behavior change

## Documentation

---

Claude Code

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
@@ -113,9 +113,7 @@ public class DynamicKafkaSourceEnumerator
     private Map<String, DynamicKafkaSourceEnumState.RetainedClusterState>
             retainedClusterEnumeratorStates;
     private boolean firstDiscoveryComplete;
-    private boolean initialReaderRegistrationPending;
-    private final Map<Integer, List<DynamicKafkaSourceSplit>> pendingReportedSplitsByReader;
-    private final Set<Integer> pendingMetadataUpdateReaders;
+    private final ReaderRecoveryCoordinator readerRecoveryCoordinator;
 
     public DynamicKafkaSourceEnumerator(
             KafkaStreamSubscriber kafkaStreamSubscriber,
@@ -217,10 +215,9 @@ public class DynamicKafkaSourceEnumerator
                                         runnable, "dynamic-kafka-enumerator-closing-worker"));
         this.asynchronousEnumeratorCloseFailure = new AtomicReference<>();
         this.splitAssignmentStrategy = createSplitAssignmentStrategy(properties);
-        this.initialReaderRegistrationPending =
-                hasRestoredEnumeratorState(dynamicKafkaSourceEnumState);
-        this.pendingReportedSplitsByReader = new HashMap<>();
-        this.pendingMetadataUpdateReaders = new HashSet<>();
+        this.readerRecoveryCoordinator =
+                new ReaderRecoveryCoordinator(
+                        hasRestoredEnumeratorState(dynamicKafkaSourceEnumState));
 
         if (!dynamicKafkaSourceEnumState.getClusterEnumeratorStates().isEmpty()) {
             logger.info("Dynamic Kafka source restored from checkpointed enumerator state");
@@ -558,7 +555,8 @@ public class DynamicKafkaSourceEnumerator
     /** NOTE: Must run on coordinator thread. */
     private void sendMetadataUpdateEventToAvailableReaders() {
         if (shouldDeferMetadataUpdateEvents()) {
-            pendingMetadataUpdateReaders.addAll(enumContext.registeredReaders().keySet());
+            readerRecoveryCoordinator.deferMetadataUpdates(
+                    enumContext.registeredReaders().keySet());
             return;
         }
 
@@ -768,11 +766,8 @@ public class DynamicKafkaSourceEnumerator
         logger.debug("Adding reader {}", subtaskId);
         ReaderInfo readerInfo = enumContext.registeredReaders().get(subtaskId);
         if (readerInfo != null) {
-            List<DynamicKafkaSourceSplit> reportedSplits =
-                    readerInfo.getReportedSplitsOnRegistration();
-            if (!reportedSplits.isEmpty()) {
-                pendingReportedSplitsByReader.put(subtaskId, new ArrayList<>(reportedSplits));
-            }
+            readerRecoveryCoordinator.recordReportedSplits(
+                    subtaskId, readerInfo.getReportedSplitsOnRegistration());
         }
 
         if (tryCompletePendingReaderRegistration()) {
@@ -784,19 +779,15 @@ public class DynamicKafkaSourceEnumerator
     }
 
     private boolean tryCompletePendingReaderRegistration() {
-        boolean hasPendingRecovery =
-                initialReaderRegistrationPending || !pendingReportedSplitsByReader.isEmpty();
-        if (!hasPendingRecovery) {
+        if (!readerRecoveryCoordinator.hasPendingRecovery()) {
             return false;
         }
         if (!firstDiscoveryComplete || !allReadersRegistered()) {
             return true;
         }
 
-        if (initialReaderRegistrationPending) {
-            initialReaderRegistrationPending = false;
-        }
-        if (!pendingReportedSplitsByReader.isEmpty()) {
+        readerRecoveryCoordinator.markInitialRegistrationComplete();
+        if (readerRecoveryCoordinator.hasReportedSplits()) {
             reassignReportedSplits();
         } else {
             flushPendingSplitAssignmentsForRegisteredReaders();
@@ -830,7 +821,7 @@ public class DynamicKafkaSourceEnumerator
         long currentTimeMillis = System.currentTimeMillis();
 
         for (Entry<Integer, List<DynamicKafkaSourceSplit>> readerSplits :
-                new TreeMap<>(pendingReportedSplitsByReader).entrySet()) {
+                readerRecoveryCoordinator.drainReportedSplits().entrySet()) {
             int readerId = readerSplits.getKey();
             for (DynamicKafkaSourceSplit split : readerSplits.getValue()) {
                 if (isSplitActive(split)) {
@@ -875,19 +866,14 @@ public class DynamicKafkaSourceEnumerator
         if (!retainedSplitsByReader.isEmpty()) {
             enumContext.assignSplits(new SplitsAssignment<>(retainedSplitsByReader));
         }
-        pendingReportedSplitsByReader.clear();
     }
 
     private boolean shouldDeferMetadataUpdateEvents() {
-        return initialReaderRegistrationPending
-                || (!pendingReportedSplitsByReader.isEmpty() && !allReadersRegistered());
+        return readerRecoveryCoordinator.shouldDeferMetadataUpdateEvents(allReadersRegistered());
     }
 
     private void flushPendingMetadataUpdateEvents() {
-        List<Integer> readers = new ArrayList<>(pendingMetadataUpdateReaders);
-        Collections.sort(readers);
-        pendingMetadataUpdateReaders.clear();
-        for (int readerId : readers) {
+        for (int readerId : readerRecoveryCoordinator.drainDeferredMetadataUpdateReaders()) {
             if (enumContext.registeredReaders().containsKey(readerId)) {
                 sendMetadataUpdateEvent(readerId);
             }
@@ -1043,7 +1029,7 @@ public class DynamicKafkaSourceEnumerator
 
         if (enumContext.registeredReaders().containsKey(subtaskId)) {
             if (shouldDeferMetadataUpdateEvents()) {
-                pendingMetadataUpdateReaders.add(subtaskId);
+                readerRecoveryCoordinator.deferMetadataUpdate(subtaskId);
             } else {
                 sendMetadataUpdateEvent(subtaskId);
             }
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
@@ -113,9 +113,7 @@ public class DynamicKafkaSourceEnumerator
     private Map<String, DynamicKafkaSourceEnumState.RetainedClusterState>
             retainedClusterEnumeratorStates;
     private boolean firstDiscoveryComplete;
-    private boolean initialReaderRegistrationPending;
-    private final Map<Integer, List<DynamicKafkaSourceSplit>> pendingReportedSplitsByReader;
-    private final Set<Integer> pendingMetadataUpdateReaders;
+    private final ReaderRecoveryGate readerRecoveryGate;
 
     public DynamicKafkaSourceEnumerator(
             KafkaStreamSubscriber kafkaStreamSubscriber,
@@ -217,10 +215,8 @@ public class DynamicKafkaSourceEnumerator
                                         runnable, "dynamic-kafka-enumerator-closing-worker"));
         this.asynchronousEnumeratorCloseFailure = new AtomicReference<>();
         this.splitAssignmentStrategy = createSplitAssignmentStrategy(properties);
-        this.initialReaderRegistrationPending =
-                hasRestoredEnumeratorState(dynamicKafkaSourceEnumState);
-        this.pendingReportedSplitsByReader = new HashMap<>();
-        this.pendingMetadataUpdateReaders = new HashSet<>();
+        this.readerRecoveryGate =
+                new ReaderRecoveryGate(hasRestoredEnumeratorState(dynamicKafkaSourceEnumState));
 
         if (!dynamicKafkaSourceEnumState.getClusterEnumeratorStates().isEmpty()) {
             logger.info("Dynamic Kafka source restored from checkpointed enumerator state");
@@ -558,7 +554,7 @@ public class DynamicKafkaSourceEnumerator
     /** NOTE: Must run on coordinator thread. */
     private void sendMetadataUpdateEventToAvailableReaders() {
         if (shouldDeferMetadataUpdateEvents()) {
-            pendingMetadataUpdateReaders.addAll(enumContext.registeredReaders().keySet());
+            readerRecoveryGate.deferMetadataUpdates(enumContext.registeredReaders().keySet());
             return;
         }
 
@@ -768,11 +764,8 @@ public class DynamicKafkaSourceEnumerator
         logger.debug("Adding reader {}", subtaskId);
         ReaderInfo readerInfo = enumContext.registeredReaders().get(subtaskId);
         if (readerInfo != null) {
-            List<DynamicKafkaSourceSplit> reportedSplits =
-                    readerInfo.getReportedSplitsOnRegistration();
-            if (!reportedSplits.isEmpty()) {
-                pendingReportedSplitsByReader.put(subtaskId, new ArrayList<>(reportedSplits));
-            }
+            readerRecoveryGate.recordReportedSplits(
+                    subtaskId, readerInfo.getReportedSplitsOnRegistration());
         }
 
         if (tryCompletePendingReaderRegistration()) {
@@ -784,19 +777,15 @@ public class DynamicKafkaSourceEnumerator
     }
 
     private boolean tryCompletePendingReaderRegistration() {
-        boolean hasPendingRecovery =
-                initialReaderRegistrationPending || !pendingReportedSplitsByReader.isEmpty();
-        if (!hasPendingRecovery) {
+        if (!readerRecoveryGate.hasPendingRecovery()) {
             return false;
         }
         if (!firstDiscoveryComplete || !allReadersRegistered()) {
             return true;
         }
 
-        if (initialReaderRegistrationPending) {
-            initialReaderRegistrationPending = false;
-        }
-        if (!pendingReportedSplitsByReader.isEmpty()) {
+        readerRecoveryGate.markInitialRegistrationComplete();
+        if (readerRecoveryGate.hasReportedSplits()) {
             reassignReportedSplits();
         } else {
             flushPendingSplitAssignmentsForRegisteredReaders();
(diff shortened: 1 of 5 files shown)


spuru9:
Same as the other PR
@slachiewicz I dont think it anymore falls in the domain of a hotfix. Can you file a JIRA for this. Also can you rework on the PR description as per the guidlines.

Efrat19 (approved):
Confirmed no logic change was introduced
Thanks for the refactor and coverage

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/ReaderRecoveryCoordinator.java
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Tracks the recovery-time reader registration state of the {@link DynamicKafkaSourceEnumerator}.
+ *
+ * <p>When the enumerator is restored from checkpointed state, split assignment and metadata update
+ * events must be deferred until the first metadata discovery has completed and every reader has
+ * (re-)registered, so that restored reader splits can be redistributed consistently. This class
+ * owns that gating state; the enumerator remains responsible for acting on it.
+ */
+@Internal
+class ReaderRecoveryCoordinator {
```
Efrat19: Nit: `Coordinator` suffix in usually for `OperatorCoordinator` implementations, associated with
job vertex (i.e `SourceCoordinator`, `CheckpointCoordinator`)
I'd consider a different naming

slachiewicz: Renamed to `ReaderRecoveryGate` — agreed, `Coordinator` reads as an `OperatorCoordinator` here and this is not one. Title, body and FLINK-40137 updated to match.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -555,10 +551,10 @@ private Set<KafkaStream> handleFetchSubscribedStreamsError(
         }
     }
 
-    /** NOTE: Must run on coordinator thread. */
+    /** NOTE: Must run on gate thread. */
```
Efrat19:
```suggestion
    /** NOTE: Must run on coordinator thread. */
```

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -574,7 +570,7 @@ private void sendMetadataUpdateEventToAvailableReaders() {
      * not assigning to the offsets prior to reader restart). Split offset resolution is mostly
      * managed by the readers.
      *
-     * <p>NOTE: Must run on coordinator thread
+     * <p>NOTE: Must run on gate thread
```
Efrat19:
```suggestion
     * <p>NOTE: Must run on coordinator thread
```

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -706,11 +702,11 @@ private void handleAsynchronousEnumeratorCloseFailure(Exception e) {
                     () -> {
                         throw new RuntimeException(e);
                     });
-        } catch (Throwable coordinatorFailure) {
+        } catch (Throwable gateFailure) {
```
Efrat19:
```suggestion
        } catch (Throwable coordinatorFailure) {
```

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -706,11 +702,11 @@ private void handleAsynchronousEnumeratorCloseFailure(Exception e) {
...
             logger.warn(
                     "Unable to propagate asynchronous dynamic Kafka enumerator close failure to "
-                            + "the coordinator thread. The failure will be rethrown during close.",
-                    coordinatorFailure);
+                            + "the gate thread. The failure will be rethrown during close.",
```
Efrat19:
```suggestion
                            + "the coordinator thread. The failure will be rethrown during close.",
```

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -706,11 +702,11 @@ private void handleAsynchronousEnumeratorCloseFailure(Exception e) {
...
+                    gateFailure);
```
Efrat19:
```suggestion
                    coordinatorFailure);
```

Efrat19 (commented):
Thanks. Looks like a few coordinator references were incorrectly renamed

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/ReaderRecoveryCoordinator.java
Efrat19: LGTM % https://github.com/apache/flink-connector-kafka/pull/289#pullrequestreview-5002009224

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
slachiewicz: Good catch — the rename overreached into the `coordinator thread` references from `runInCoordinatorThread`. All five spots restored in 3eaf13fc (the two javadoc notes, the `coordinatorFailure` catch variable, and the log message + argument); everything else named `gate` is the `ReaderRecoveryGate` itself.

slachiewicz:
Pushed `3eaf13fc` addressing the review: the rename to `ReaderRecoveryGate` had overreached into Flink's coordinator-thread references in `DynamicKafkaSourceEnumerator`. All five spots are restored —

* the two `Must run on coordinator thread` javadoc notes (`sendMetadataUpdateEventToAvailableReaders`, `createEnumeratorWithAssignedTopicPartitions`)
* the `coordinatorFailure` catch variable in `handleAsynchronousEnumeratorCloseFailure`
* the close-failure log message and its argument

Everything else named `gate` refers to `ReaderRecoveryGate` itself. Verified with a repo-wide grep (no `gate thread`/`gateFailure` left) plus `spotless:check` and a module compile. Conversations resolved; FLINK-40137 reopened.

slachiewicz:
The previous CI run (`compile_and_test (17)`) hit a 50-minute timeout due to an intermittent hang in `SourceTestSuiteBase.testIdleReader`:

```text
at org.apache.flink.streaming.api.operators.collect.CollectResultIterator.hasNext(CollectResultIterator.java:98)
at org.apache.flink.connector.testframe.utils.CollectIteratorAssert.compareWithExactlyOnceSemantic(CollectIteratorAssert.java:130)
at org.apache.flink.connector.testframe.utils.CollectIteratorAssert.matchesRecordsFromSource(CollectIteratorAssert.java:85)
at org.apache.flink.connector.testframe.testsuites.SourceTestSuiteBase.checkResultWithSemantic(SourceTestSuiteBase.java:749)
at org.apache.flink.connector.testframe.testsuites.SourceTestSuiteBase.testIdleReader(SourceTestSuiteBase.java:526)
```

This is the known idle-reader hang race (where `no-more-splits` arrives before the metadata update) addressed in #291 ([FLINK-40362](https://issues.apache.org/jira/browse/FLINK-40362)). The failure is unrelated to the `ReaderRecoveryGate` refactoring in this PR.

Re-pushed commit to trigger a clean CI run.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/ReaderRecoveryGate.java
```diff
@@ -0,0 +1,112 @@
...
+
+    /** Readers whose metadata update events were deferred during recovery. */
+    private final Set<Integer> pendingMetadataUpdateReaders = new HashSet<>();
+
+    ReaderRecoveryGate(boolean restoredFromCheckpoint) {
+        this.initialReaderRegistrationPending = restoredFromCheckpoint;
+    }
+
+    /** Records splits a reader reported on registration; an empty report is ignored. */
+    void recordReportedSplits(int subtaskId, List<DynamicKafkaSourceSplit> reportedSplits) {
```
Savonitar, line 61: Is my understanding correct, that even if this method called twice for same subtask (e.g. during failover), it does not bring any issues? if so, do we have a test for that case?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/ReaderRecoveryGate.java
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Tracks the recovery-time reader registration state of the {@link DynamicKafkaSourceEnumerator}.
+ *
+ * <p>When the enumerator is restored from checkpointed state, split assignment and metadata update
+ * events must be deferred until the first metadata discovery has completed and every reader has
+ * (re-)registered, so that restored reader splits can be redistributed consistently. This class
+ * owns that gating state; the enumerator remains responsible for acting on it.
+ */
+@Internal
+class ReaderRecoveryGate {
```
Savonitar, line 44: this class is not thread-safe, right? but it is called only in coordinator thread, that's why it is fine. am I right?

Savonitar (commented):
Hi. Thanks for the PR.
Worth to mention, there is https://github.com/apache/flink-connector-kafka/pull/295 which fixes one of the pre-existing issues with DynamicKafka workflow.
I left a few non-blocking questions.

Efrat19:
> Pushed `3eaf13fc` addressing the review: the rename to `ReaderRecoveryGate` had overreached into Flink's coordinator-thread references in `DynamicKafkaSourceEnumerator`. All five spots are restored —
> 
> * the two `Must run on coordinator thread` javadoc notes (`sendMetadataUpdateEventToAvailableReaders`, `createEnumeratorWithAssignedTopicPartitions`)
> * the `coordinatorFailure` catch variable in `handleAsynchronousEnumeratorCloseFailure`
> * the close-failure log message and its argument
> 
> Everything else named `gate` refers to `ReaderRecoveryGate` itself. Verified with a repo-wide grep (no `gate thread`/`gateFailure` left) plus `spotless:check` and a module compile. Conversations resolved; [FLINK-40137](https://issues.apache.org/jira/browse/FLINK-40137) reopened.

Thanks, LGTM

MartijnVisser (changes requested):
Looks good: a straight extraction with no behaviour change, and the gate test covers all the transitions. One thing before I merge: please replace the `Co-Authored-By` trailer with `Generated-by: <tool>` per the ASF generative tooling guidance (same as your #291 commit).

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/ReaderRecoveryGate.java
MartijnVisser: Correct, everything that touches the gate runs on the coordinator thread: the enumerator methods the SourceCoordinator calls, and the metadata refresh result is handed back through `runInCoordinatorThread` in `StoppableKafkaMetadataServiceDiscoveryContext`. Same as the three fields before this extraction. Worth a one-line note in the class Javadoc that the gate is coordinator-thread only, so nobody adds synchronisation or calls it from the discovery worker later.

slachiewicz: Added in `586295b1` as a Javadoc paragraph on `ReaderRecoveryGate`: not thread-safe, coordinator-thread only, naming the `SourceCoordinator` and `runInCoordinatorThread` paths it arrives through.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/ReaderRecoveryGate.java
slachiewicz: Correct, no issue. The reports are held in a map keyed by subtask id, so a second `addReader` for the same subtask replaces that reader entry instead of accumulating, and the duplicate-owner check in `reassignReportedSplits` cannot be tripped by a reader's own re-report.

There was no test, so `586295b1` adds two: `testRepeatedReportForSameReaderReplacesPreviousReport` and `testEmptyRepeatedReportRetainsPreviousReport`. The second pins a pre-existing edge — an empty re-report is ignored, so an earlier non-empty report survives until the drain. `addReader` had the same guard around the same `put` before this extraction, so that is unchanged behaviour, not something introduced here.

slachiewicz:
Done in `586295b1`: `Co-Authored-By` replaced with `Generated-by: Claude Fable 5`, per the ASF generative tooling guidance and matching #291. The same push adds the coordinator-thread Javadoc note and two gate tests answering @Savonitar's re-registration question; production code is otherwise untouched since your review.

#291 is updated too (`4eded31c`), so the #291 -> #289 -> #300 order you set out is unblocked from my side.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/ReaderRecoveryGate.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.Set;
+import java.util.TreeMap;
+
+/**
+ * Tracks the recovery-time reader registration state of the {@link DynamicKafkaSourceEnumerator}.
+ *
+ * <p>When the enumerator is restored from checkpointed state, split assignment and metadata update
+ * events must be deferred until the first metadata discovery has completed and every reader has
+ * (re-)registered, so that restored reader splits can be redistributed consistently. This class
+ * owns that gating state; the enumerator remains responsible for acting on it.
```
MartijnVisser: This describes the restore trigger only. The gate is also armed on a running enumerator when a reader re-registers after a partial failover and reports its checkpointed splits (`recordReportedSplits` with a non-empty list while `initialReaderRegistrationPending` is already false). Since making the gating explicit is the point of this class, please name both triggers and what each waits for: a restore waits for the first discovery plus all readers; a reported-splits registration waits for all readers to be registered again.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -830,7 +819,7 @@ private void reassignReportedSplits() {
         long currentTimeMillis = System.currentTimeMillis();
 
         for (Entry<Integer, List<DynamicKafkaSourceSplit>> readerSplits :
-                new TreeMap<>(pendingReportedSplitsByReader).entrySet()) {
+                readerRecoveryGate.drainReportedSplits().entrySet()) {
```
MartijnVisser, line 822: This is the one spot where the extraction is not a pure move: the pending map used to be cleared at the end of `reassignReportedSplits`, now it is cleared before the loop runs. Not observable today, nothing between here and the end of the method reads the gate, and an exception in here fails the job through the coordinator anyway. It does mean that anyone who later consults the gate from `handleNoMoreSplits`, which runs re-entrantly from inside this loop via the sub-enumerator's no-more-splits callback, sees an empty map. A one-line note on `drainReportedSplits` that the state is cleared eagerly and must not be consulted while reassigning is enough.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/ReaderRecoveryGateTest.java
```diff
@@ -0,0 +1,138 @@
...
+    @Test
+    void testReportedSplitsGateUntilAllReadersRegistered() {
+        ReaderRecoveryGate gate = new ReaderRecoveryGate(false);
+        gate.recordReportedSplits(1, Collections.singletonList(split("topic", 0)));
+
+        assertThat(gate.hasPendingRecovery()).isTrue();
+        assertThat(gate.hasReportedSplits()).isTrue();
+        assertThat(gate.shouldDeferMetadataUpdateEvents(false)).isTrue();
+        assertThat(gate.shouldDeferMetadataUpdateEvents(true)).isFalse();
+    }
```
MartijnVisser, line 88: These cover each field on its own but not the interaction between the restore flag and the reported splits, which is the part the description calls hard to follow. One more test: `new ReaderRecoveryGate(true)`, record splits for one reader, `markInitialRegistrationComplete()`; then `hasPendingRecovery()` is still true, `shouldDeferMetadataUpdateEvents(false)` is true and `shouldDeferMetadataUpdateEvents(true)` is false; after `drainReportedSplits()` all three are false. That pins that completing the initial registration does not release the gate while reported splits are still pending.

MartijnVisser (commented):
"untestable without a full container-based test setup" is not accurate: `DynamicKafkaSourceEnumeratorRecoveryTest`, added in #278 together with these three fields, already drives the deferral, the reassignment and the retained-split path through `MockSplitEnumeratorContext` and `MockKafkaMetadataService` without containers. The value here is that the gating state has a name and a home, which is fine, but the commit message and the PR body should say that instead. While amending, please use the component tag: `[FLINK-40137][Connectors/Kafka]`.

Verified locally on 586295b1: `ReaderRecoveryGateTest`, `DynamicKafkaSourceEnumeratorRecoveryTest`, `DynamicKafkaSourceEnumeratorTest` and `DynamicKafkaSourceITTest` are all green. The JDK 21 CI failure is the FLINK-40362 `testIdleReader` hang that #291 fixes, so the order stays #291, then this, then #300. I will approve and merge on the amended push.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/ReaderRecoveryGate.java
MartijnVisser: Settled, thanks for the tests. For the record: on a subtask failover Flink runs `subtaskReset` (which calls `addSplitsBack` and drops the reader from `registeredReaders`) and then a fresh `addReader`, and, while the gate is still pending, the reader reports the same checkpointed split set both times, so replacing the entry is the right semantics and the duplicate-owner check only ever compares different readers.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/ReaderRecoveryGate.java
slachiewicz: Updated class Javadoc in `a9ec7c19` to detail both recovery triggers (restore and partial failover re-registration with reported splits) and their respective gating conditions.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
slachiewicz: Added a note on `drainReportedSplits` in `a9ec7c19` that the state is cleared eagerly and must not be consulted while reassigning.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/ReaderRecoveryGateTest.java
slachiewicz: Added `testCompletingInitialRegistrationDoesNotReleaseGateWithPendingReportedSplits` in `a9ec7c19` to test this interaction.

slachiewicz:
@MartijnVisser Addressed all review items in `a9ec7c19` and updated the PR title and description:
- Added component tag `[FLINK-40137][Connectors/Kafka]` to commit subject and PR title.
- Updated commit message and PR description to reflect that the value is giving the gating state and its lifecycle an explicit name and structure.
- Updated `ReaderRecoveryGate` class Javadoc to explain both recovery triggers (restore vs. partial failover re-registration with reported splits) and their respective gating conditions.
- Added Javadoc note on `drainReportedSplits` explaining that state is cleared eagerly.
- Added `testCompletingInitialRegistrationDoesNotReleaseGateWithPendingReportedSplits` in `ReaderRecoveryGateTest` testing the interaction between the restore flag and pending reported splits.

MartijnVisser (approved):
All addressed, thanks.
CI on this head failed only on FLINK-40398, FLINK-40615 and an unticketed flake in FlinkKafkaInternalProducerITCase (Edit: filed FLINK-40621), none of which this PR touches, so I'm merging on that run deliberately.
Approving; I'll merge.

PR #290, opened 2026-08-04
[FLINK-36434][Connectors/Kafka] Create the KafkaConsumer lazily on the split fetcher thread

Zhuoxi2000:

Fixes [FLINK-36434](https://issues.apache.org/jira/browse/FLINK-36434): the `KafkaPartitionSplitReader` is constructed on the source-reader thread — or, when `KafkaSourceFetcherManager#commitOffsets` re-creates a fetcher after an idle shutdown, on the checkpoint thread — and its constructor eagerly created the `KafkaConsumer` (and registered consumer metrics) there. The consumer is then used almost exclusively on the split fetcher thread, but `KafkaConsumer` is not thread-safe; as described in the ticket, this broken threading model blocks Kafka client upgrades (sporadic `CorrelationIdMismatchException`).

This PR makes the consumer lazy inside the reader, so it is created on the first consumer-touching call — which by construction happens on the split fetcher thread (`fetch`, `handleSplitsChanges`, `pauseOrResumeSplits`, the offset-commit task, and `close()` in the fetcher's shutdown path all run there). Design points, as discussed on the ticket and in review:

- The constructor only computes `consumerProps` (client-id derivation unchanged) and stores them; no consumer, no metric registration.
- A private `ensureConsumer()` runs at the entry of every consumer-touching method; consumer metric registration moves there as well. The creation branch needs no guard: only the fetcher thread calls it, so the field is single-writer.
- `wakeUp()` remains the only cross-thread entry point (the one call `KafkaConsumer` documents as thread-safe). No lock is needed: `wakeUp()` only ever targets a running fetch task (see `SplitReader#wakeUp`'s javadoc), and the base `SplitFetcher` can only schedule a fetch task after `handleSplitsChanges()` — and therefore `ensureConsumer()` — has run at least once on the fetcher thread. So a plain `volatile` field covers the cross-thread reads, and a pre-creation wakeup has nothing to interrupt and is safely dropped. A code comment records that this leans on a `SplitFetcher`/`AddSplitsTask` implementation invariant rather than a documented contract.
- `close()` before first use is a no-op.

`KafkaPartitionSplitReaderWrapper` (DynamicKafkaSource) subclasses this reader and only wraps `fetch()` output, so the fix covers the `DynamicKafkaSource` path as well.

Deferring `splitReaderFactory.get()` itself onto the fetcher thread in flink-connector-base would fix this class of issue for all connectors; with the lazy consumer this connector no longer depends on that, so I will file it as a separate ticket (as discussed on FLINK-36434, it is orthogonal).

## Brief change log

  - `KafkaPartitionSplitReader` no longer creates the `KafkaConsumer` (nor registers consumer metrics) in its constructor; both happen in a new `ensureConsumer()` on the first consumer-touching call, i.e. on the split fetcher thread
  - `wakeUp()` before the consumer exists is safely dropped — it can only fire against a running fetch task, which cannot exist before the consumer does; `close()` before first use is a no-op
  - A `@VisibleForTesting createConsumer(Properties)` seam allows tests to observe consumer creation

## Verifying this change

Please make sure both new and modified tests in this PR follow [the conventions for tests defined in our code quality guide](https://flink.apache.org/how-to-contribute/code-style-and-quality-common/#7-testing).

  - Added `KafkaPartitionSplitReaderLazyConsumerCreationTest` (pure unit tests, no Kafka cluster needed): the consumer is not created at construction time and is created on the thread that first uses the reader; a `wakeUp()` arriving before creation neither throws nor creates a consumer, and is not replayed after creation; a `wakeUp()` after creation delegates directly; `close()` without prior use never creates a consumer
  - Existing `KafkaPartitionSplitReaderTest` covers the eager-path behaviors against a real cluster and passes unchanged in CI (not run locally: no Docker environment on this machine); `mvn test` for the module (compilation, checkstyle, spotless, and the new tests) passes locally

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (no — `KafkaPartitionSplitReader` is `@Internal`)
  - The serializers: (no)
  - The runtime per-record code paths (performance sensitive): (no — `fetch()` gains a single volatile read on its fast path; consumer creation is one-time)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Kubernetes/Yarn, ZooKeeper: (no)
  - The S3 file system connector: (no)

## Documentation

---

Claude Code

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
@@ -66,12 +66,27 @@ public class KafkaPartitionSplitReader
     private static final Logger LOG = LoggerFactory.getLogger(KafkaPartitionSplitReader.class);
     private static final long POLL_TIMEOUT = 10000L;
 
-    private final KafkaConsumer<byte[], byte[]> consumer;
+    private final Properties consumerProps;
     private final Map<TopicPartition, Long> stoppingOffsets;
     private final String groupId;
     private final int subtaskId;
 
     private final KafkaSourceReaderMetrics kafkaSourceReaderMetrics;
+
+    /** Guards lazy creation of {@link #consumer} and the {@link #pendingWakeup} flag. */
+    private final Object consumerLock = new Object();
+
+    /**
+     * The Kafka consumer. Created lazily on the first thread that actually uses it — the split
+     * fetcher thread — rather than on the thread constructing this reader (the source-reader or
+     * checkpoint thread). {@link KafkaConsumer} is not thread-safe, so it must live on the thread
+     * that uses it (FLINK-36434). Access outside {@link #ensureConsumer()} / {@link #wakeUp()} is
+     * only safe after a call to {@link #ensureConsumer()} on the same thread.
+     */
+    @Nullable private volatile KafkaConsumer<byte[], byte[]> consumer;
+
+    /** Set when {@link #wakeUp()} is called before the consumer exists; guarded by the lock. */
+    private boolean pendingWakeup;
 
     // Tracking empty splits that has not been added to finished splits in fetch()
     private final Set<String> emptySplits = new HashSet<>();
@@ -94,17 +109,56 @@ public class KafkaPartitionSplitReader
         consumerProps.putAll(props);
         consumerProps.setProperty(ConsumerConfig.CLIENT_ID_CONFIG, createConsumerClientId(props));
         setConsumerClientRack(consumerProps, rackIdSupplier);
-        this.consumer = new KafkaConsumer<>(consumerProps);
+        this.consumerProps = consumerProps;
         this.stoppingOffsets = new HashMap<>();
         this.groupId = consumerProps.getProperty(ConsumerConfig.GROUP_ID_CONFIG);
+    }
 
-        // Metric registration
-        maybeRegisterKafkaConsumerMetrics(props, kafkaSourceReaderMetrics, consumer);
-        this.kafkaSourceReaderMetrics.registerNumBytesIn(consumer);
+    /**
+     * Returns the consumer, creating it on the calling thread on first use.
+     *
+     * <p>The reader is constructed on the source-reader (or, when a fetcher is re-created for an
+     * offset commit, the checkpoint) thread, while the consumer is used almost exclusively on the
+     * split fetcher thread. Creating the consumer eagerly in the constructor therefore puts it on
+     * the wrong thread. All consumer-touching {@link SplitReader} methods run on the fetcher
+     * thread, so deferring creation to the first such call keeps construction and use on the same
+     * thread. The only cross-thread entry point remains {@link #wakeUp()}, which is the one call
+     * {@link KafkaConsumer} documents as thread-safe.
+     */
+    private KafkaConsumer<byte[], byte[]> ensureConsumer() {
+        KafkaConsumer<byte[], byte[]> currentConsumer = this.consumer;
+        if (currentConsumer != null) {
+            return currentConsumer;
+        }
+        synchronized (consumerLock) {
+            currentConsumer = this.consumer;
+            if (currentConsumer == null) {
+                currentConsumer = createConsumer(consumerProps);
+                maybeRegisterKafkaConsumerMetrics(
+                        consumerProps, kafkaSourceReaderMetrics, currentConsumer);
+                kafkaSourceReaderMetrics.registerNumBytesIn(currentConsumer);
+                if (pendingWakeup) {
+                    // A wakeUp() arrived before the consumer existed. Apply it now so that the
+                    // first blocking call still observes it, matching the behavior of a wakeup
+                    // against an eagerly-created consumer.
+                    currentConsumer.wakeup();
+                    pendingWakeup = false;
+                }
+                this.consumer = currentConsumer;
+            }
+            return currentConsumer;
+        }
+    }
+
+    /** Creates the {@link KafkaConsumer}. Overridable for tests to observe the creation. */
+    @VisibleForTesting
+    protected KafkaConsumer<byte[], byte[]> createConsumer(Properties consumerProps) {
+        return new KafkaConsumer<>(consumerProps);
     }
 
     @Override
     public RecordsWithSplitIds<ConsumerRecord<byte[], byte[]>> fetch() throws IOException {
+        final KafkaConsumer<byte[], byte[]> consumer = ensureConsumer();
         ConsumerRecords<byte[], byte[]> consumerRecords;
         try {
             consumerRecords = consumer.poll(Duration.ofMillis(POLL_TIMEOUT));
@@ -180,6 +234,7 @@ public class KafkaPartitionSplitReader
                             "The SplitChange type of %s is not supported.",
                             splitsChange.getClass()));
         }
+        final KafkaConsumer<byte[], byte[]> consumer = ensureConsumer();
 
         // Assignment.
         List<TopicPartition> newPartitionAssignments = new ArrayList<>();
@@ -228,18 +283,34 @@ public class KafkaPartitionSplitReader
 
     @Override
     public void wakeUp() {
-        consumer.wakeup();
+        // The only method called from a thread other than the fetcher thread, relying on
+        // KafkaConsumer.wakeup() being the one documented thread-safe consumer call. If the
+        // consumer does not exist yet, record the wakeup and apply it on creation.
+        synchronized (consumerLock) {
+            KafkaConsumer<byte[], byte[]> currentConsumer = this.consumer;
+            if (currentConsumer != null) {
+                currentConsumer.wakeup();
+            } else {
+                pendingWakeup = true;
+            }
+        }
     }
 
     @Override
     public void close() throws Exception {
-        consumer.close();
+        synchronized (consumerLock) {
+            KafkaConsumer<byte[], byte[]> currentConsumer = this.consumer;
+            if (currentConsumer != null) {
+                currentConsumer.close();
+            }
+        }
     }
 
     @Override
     public void pauseOrResumeSplits(
             Collection<KafkaPartitionSplit> splitsToPause,
             Collection<KafkaPartitionSplit> splitsToResume) {
+        final KafkaConsumer<byte[], byte[]> consumer = ensureConsumer();
         // Filter against current assignment to avoid IllegalStateException when a partition
         // was concurrently unassigned by fetch() or removeEmptySplits().
         Set<TopicPartition> assigned = consumer.assignment();
@@ -270,12 +341,12 @@ public class KafkaPartitionSplitReader
     public void notifyCheckpointComplete(
             Map<TopicPartition, OffsetAndMetadata> offsetsToCommit,
             OffsetCommitCallback offsetCommitCallback) {
-        consumer.commitAsync(offsetsToCommit, offsetCommitCallback);
+        ensureConsumer().commitAsync(offsetsToCommit, offsetCommitCallback);
     }
 
     @VisibleForTesting
     KafkaConsumer<byte[], byte[]> consumer() {
-        return consumer;
+        return ensureConsumer();
     }
 
     // --------------- private helper method ----------------------
@@ -295,7 +366,7 @@ public class KafkaPartitionSplitReader
     }
 
     long getConsumerPosition(TopicPartition tp, String msg) {
-        return retryOnWakeup(() -> consumer.position(tp), msg);
+        return retryOnWakeup(() -> ensureConsumer().position(tp), msg);
     }
 
     private void parseStartingOffsets(
@@ -348,30 +419,31 @@ public class KafkaPartitionSplitReader
 
         if (!partitionsStartingFromEarliest.isEmpty()) {
             LOG.trace("Seeking starting offsets to beginning: {}", partitionsStartingFromEarliest);
-            consumer.seekToBeginning(partitionsStartingFromEarliest);
+            ensureConsumer().seekToBeginning(partitionsStartingFromEarliest);
         }
 
         if (!partitionsStartingFromLatest.isEmpty()) {
             LOG.trace("Seeking starting offsets to end: {}", partitionsStartingFromLatest);
-            consumer.seekToEnd(partitionsStartingFromLatest);
+            ensureConsumer().seekToEnd(partitionsStartingFromLatest);
         }
 
         if (!partitionsStartingFromSpecifiedOffsets.isEmpty()) {
             LOG.trace(
                     "Seeking starting offsets to specified offsets: {}",
                     partitionsStartingFromSpecifiedOffsets);
-            partitionsStartingFromSpecifiedOffsets.forEach(consumer::seek);
+            partitionsStartingFromSpecifiedOffsets.forEach(ensureConsumer()::seek);
         }
     }
 
     private void acquireAndSetStoppingOffsets(
             List<TopicPartition> partitionsStoppingAtLatest,
             Set<TopicPartition> partitionsStoppingAtCommitted) {
-        Map<TopicPartition, Long> endOffset = consumer.endOffsets(partitionsStoppingAtLatest);
+        Map<TopicPartition, Long> endOffset =
+                ensureConsumer().endOffsets(partitionsStoppingAtLatest);
         stoppingOffsets.putAll(endOffset);
         if (!partitionsStoppingAtCommitted.isEmpty()) {
             retryOnWakeup(
-                            () -> consumer.committed(partitionsStoppingAtCommitted),
+                            () -> ensureConsumer().committed(partitionsStoppingAtCommitted),
                             "getting committed offset as stopping offsets")
                     .forEach(
                             (tp, offsetAndMetadata) -> {
@@ -389,7 +461,7 @@ public class KafkaPartitionSplitReader
     private void removeEmptySplits() {
         List<TopicPartition> emptyPartitions = new ArrayList<>();
         // If none of the partitions have any records,
-        for (TopicPartition tp : consumer.assignment()) {
+        for (TopicPartition tp : ensureConsumer().assignment()) {
             long startingOffset =
                     getConsumerPosition(tp, "getting starting offset to check if split is empty");
             if (startingOffset >= getStoppingOffset(tp)) {
@@ -414,7 +486,7 @@ public class KafkaPartitionSplitReader
             SplitsChange<KafkaPartitionSplit> splitsChange) {
         if (LOG.isDebugEnabled()) {
             StringJoiner splitsInfo = new StringJoiner(",");
-            Set<TopicPartition> assginment = consumer.assignment();
+            Set<TopicPartition> assginment = ensureConsumer().assignment();
             for (KafkaPartitionSplit split : splitsChange.splits()) {
                 if (!assginment.contains(split.getTopicPartition())) {
                     continue;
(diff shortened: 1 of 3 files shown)


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -66,13 +66,28 @@ public class KafkaPartitionSplitReader
...
+     * The Kafka consumer. Created lazily on the first thread that actually uses it — the split
+     * fetcher thread — rather than on the thread constructing this reader (the source-reader or
+     * checkpoint thread). {@link KafkaConsumer} is not thread-safe, so it must live on the thread
+     * that uses it (FLINK-36434). Access outside {@link #ensureConsumer()} / {@link #wakeUp()} is
+     * only safe after a call to {@link #ensureConsumer()} on the same thread.
+     */
+    @Nullable private volatile KafkaConsumer<byte[], byte[]> consumer;
+
+    /** Set when {@link #wakeUp()} is called before the consumer exists; guarded by the lock. */
+    private boolean pendingWakeup;
```
Efrat19: `@GuardedBy("lock")`

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -66,13 +66,28 @@ public class KafkaPartitionSplitReader
...
+    private final Object consumerLock = new Object();
+
+    /**
+     * The Kafka consumer. Created lazily on the first thread that actually uses it — the split
+     * fetcher thread — rather than on the thread constructing this reader (the source-reader or
+     * checkpoint thread). {@link KafkaConsumer} is not thread-safe, so it must live on the thread
+     * that uses it (FLINK-36434). Access outside {@link #ensureConsumer()} / {@link #wakeUp()} is
+     * only safe after a call to {@link #ensureConsumer()} on the same thread.
+     */
+    @Nullable private volatile KafkaConsumer<byte[], byte[]> consumer;
```
Efrat19, line 84: `@GuardedBy("lock")`

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -94,17 +109,56 @@ public KafkaPartitionSplitReader(
...
+            return currentConsumer;
+        }
+    }
 
-        // Metric registration
-        maybeRegisterKafkaConsumerMetrics(props, kafkaSourceReaderMetrics, consumer);
-        this.kafkaSourceReaderMetrics.registerNumBytesIn(consumer);
+    /** Creates the {@link KafkaConsumer}. Overridable for tests to observe the creation. */
+    @VisibleForTesting
+    protected KafkaConsumer<byte[], byte[]> createConsumer(Properties consumerProps) {
```
Efrat19: It feels a bit weird to have the `ensureConsumer` wrapper private and `createConsumer` inner call `protected`, wdyt?
If this is just for the tests you could probably call consumer.poll with a long timeout and expect it to throw a wakeUpException
Or at least make package-private

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReaderConsumerThreadTest.java
```diff
@@ -0,0 +1,165 @@
...
+ * Unit tests for FLINK-36434: the {@link KafkaConsumer} inside {@link KafkaPartitionSplitReader}
+ * must be created lazily on the thread that uses it (the split fetcher thread), not on the thread
+ * that constructs the reader (the source-reader or checkpoint thread), because the consumer is not
+ * thread-safe.
+ *
+ * <p>These tests need no running Kafka cluster: constructing a {@link KafkaConsumer} performs no
+ * network I/O, and {@link KafkaPartitionSplitReader#fetch()} tolerates the {@code
+ * IllegalStateException}/{@code WakeupException} that polling without an assignment raises.
+ */
+class KafkaPartitionSplitReaderConsumerThreadTest {
```
Efrat19: Nit:
```suggestion
class KafkaPartitionSplitReaderLazyConsumerCreationTest {
```
or even move to `KafkaPartitionSplitReaderTest`

Efrat19 (commented):
Thanks, change seem LGTM % couple notes

Zhuoxi2000:
@Efrat19 Thanks for the review! All three addressed

Zhuoxi2000:
@AHeise  could you approve the CI run when you get a chance?

Efrat19 (approved):
LGTM, thanks

AHeise:
Thanks for tracking down FLINK-36434 — lazy creation on the fetcher thread is the right fix, and confining it to `KafkaPartitionSplitReader` is the right call for this ticket (I looked at pushing the fix into `SplitFetcherManager#createSplitFetcher`/`SplitFetcher` instead, but `getSplitReader()` is public and several callers, including `KafkaSourceFetcherManager#commitOffsets`, capture the reader reference before the fetcher thread ever runs — that's a much bigger, cross-connector change and belongs in its own ticket as you noted).

One simplification on the locking, though: I don't think `pendingWakeup`/`consumerLock` are reachable in practice, so we can drop them (and the lock) entirely.

`SplitReader#wakeUp()`'s contract is narrow: "wake up the split reader in case the fetcher thread is blocking in `fetch()`". In the whole of flink-connector-base + flink-connector-kafka, `splitReader.wakeUp()` has exactly one call site: `FetchTask#wakeUp()`, which only fires when `SplitFetcher`'s `runningTask == fetchTask`. `fetchTask` is only ever selected once `assignedSplits` is non-empty, and `assignedSplits` is populated in exactly one place — `AddSplitsTask#run()` — which unconditionally calls `handleSplitsChanges()` (i.e. `ensureConsumer()`) immediately after, on the same fetcher thread, before that thread ever loops back to pick a next task. So by construction, `wakeUp()` can never reach `KafkaPartitionSplitReader` before `ensureConsumer()` has already run once.

That means a plain `volatile` field is enough — no `AtomicReference`, no `synchronized`, no pending-wakeup bookkeeping:

```java
/** Created lazily on first use by the fetcher thread — see #ensureConsumer(). */
private volatile KafkaConsumer<byte[], byte[]> consumer;

private KafkaConsumer<byte[], byte[]> ensureConsumer() {
    KafkaConsumer<byte[], byte[]> c = consumer;
    if (c == null) {
        c = createConsumer(consumerProps);
        maybeRegisterKafkaConsumerMetrics(consumerProps, kafkaSourceReaderMetrics, c);
        kafkaSourceReaderMetrics.registerNumBytesIn(c);
        consumer = c;
    }
    return c;
}

@Override
public void wakeUp() {
    // wakeUp() only ever reaches a running fetch task (see SplitReader#wakeUp javadoc),
    // which the base fetcher can only schedule after handleSplitsChanges() has run at
    // least once, so consumer is always non-null here in practice.
    KafkaConsumer<byte[], byte[]> c = consumer;
    if (c != null) {
        c.wakeup();
    }
}

@Override
public void close() throws Exception {
    KafkaConsumer<byte[], byte[]> c = consumer;
    if (c != null) {
        c.close();
    }
}
```

`ensureConsumer()`'s creation branch needs no guard because it's single-writer (fetcher thread only calls it); `wakeUp()`'s cross-thread read just needs the `volatile` for visibility. Worth noting this leans on an implementation invariant of `SplitFetcher`/`AddSplitsTask` rather than a documented contract beyond `wakeUp()`'s own javadoc scoping — stable code, but that's why I'd keep the comment explaining it.

This also removes the exact thing @Efrat19 flagged as feeling weird (the `@GuardedBy` split between `ensureConsumer`/`createConsumer`).

The `KafkaPartitionSplitReaderLazyConsumerCreationTest` case for "wakeUp before creation is applied on creation" would need to flip to asserting that a pre-creation `wakeUp()` is safely dropped instead of deferred.

Zhuoxi2000:
> Thanks for tracking down [FLINK-36434](https://issues.apache.org/jira/browse/FLINK-36434) — lazy creation on the fetcher thread is the right fix, and confining it to `KafkaPartitionSplitReader` is the right call for this ticket (I looked at pushing the fix into `SplitFetcherManager#createSplitFetcher`/`SplitFetcher` instead, but `getSplitReader()` is public and several callers, including `KafkaSourceFetcherManager#commitOffsets`, capture the reader reference before the fetcher thread ever runs — that's a much bigger, cross-connector change and belongs in its own ticket as you noted).
> 
> One simplification on the locking, though: I don't think `pendingWakeup`/`consumerLock` are reachable in practice, so we can drop them (and the lock) entirely.
> 
> `SplitReader#wakeUp()`'s contract is narrow: "wake up the split reader in case the fetcher thread is blocking in `fetch()`". In the whole of [flink-connector-base](https://issues.apache.org/jira/browse/FLINK-connector-base) + [flink-connector-kafka](https://issues.apache.org/jira/browse/FLINK-connector-kafka), `splitReader.wakeUp()` has exactly one call site: `FetchTask#wakeUp()`, which only fires when `SplitFetcher`'s `runningTask == fetchTask`. `fetchTask` is only ever selected once `assignedSplits` is non-empty, and `assignedSplits` is populated in exactly one place — `AddSplitsTask#run()` — which unconditionally calls `handleSplitsChanges()` (i.e. `ensureConsumer()`) immediately after, on the same fetcher thread, before that thread ever loops back to pick a next task. So by construction, `wakeUp()` can never reach `KafkaPartitionSplitReader` before `ensureConsumer()` has already run once.
> 
> That means a plain `volatile` field is enough — no `AtomicReference`, no `synchronized`, no pending-wakeup bookkeeping:
> 
> ```java
> /** Created lazily on first use by the fetcher thread — see #ensureConsumer(). */
> private volatile KafkaConsumer<byte[], byte[]> consumer;
> 
> private KafkaConsumer<byte[], byte[]> ensureConsumer() {
>     KafkaConsumer<byte[], byte[]> c = consumer;
>     if (c == null) {
>         c = createConsumer(consumerProps);
>         maybeRegisterKafkaConsumerMetrics(consumerProps, kafkaSourceReaderMetrics, c);
>         kafkaSourceReaderMetrics.registerNumBytesIn(c);
>         consumer = c;
>     }
>     return c;
> }
> 
> @Override
> public void wakeUp() {
>     // wakeUp() only ever reaches a running fetch task (see SplitReader#wakeUp javadoc),
>     // which the base fetcher can only schedule after handleSplitsChanges() has run at
>     // least once, so consumer is always non-null here in practice.
>     KafkaConsumer<byte[], byte[]> c = consumer;
>     if (c != null) {
>         c.wakeup();
>     }
> }
> 
> @Override
> public void close() throws Exception {
>     KafkaConsumer<byte[], byte[]> c = consumer;
>     if (c != null) {
>         c.close();
>     }
> }
> ```
> 
> `ensureConsumer()`'s creation branch needs no guard because it's single-writer (fetcher thread only calls it); `wakeUp()`'s cross-thread read just needs the `volatile` for visibility. Worth noting this leans on an implementation invariant of `SplitFetcher`/`AddSplitsTask` rather than a documented contract beyond `wakeUp()`'s own javadoc scoping — stable code, but that's why I'd keep the comment explaining it.
> 
> This also removes the exact thing @Efrat19 flagged as feeling weird (the `@GuardedBy` split between `ensureConsumer`/`createConsumer`).
> 
> The `KafkaPartitionSplitReaderLazyConsumerCreationTest` case for "wakeUp before creation is applied on creation" would need to flip to asserting that a pre-creation `wakeUp()` is safely dropped instead of deferred.

@AHeise Thanks  agreed on both points, and I’ve updated it as suggested.

`consumerLock` and `pendingWakeup` are gone. `consumer` is now a plain `volatile`, only created by `ensureConsumer()` on the fetcher thread, while `wakeUp()` and `close()` just do a null-checked read. I kept the comment in `wakeUp()` explaining the `SplitFetcher` / `AddSplitsTask` invariant, including that we’re relying on current implementation behavior rather than a documented contract.

I updated the test as well: `testWakeUpBeforeConsumerCreationIsAppliedOnCreation` is now `testWakeUpBeforeConsumerCreationIsSafelyDropped`. A `wakeUp()` before consumer creation is simply ignored — it doesn’t throw, create the consumer, or get replayed later.

Also refreshed the PR description to remove the old pending-wakeup design.

Could you approve the CI run when you get a chance?

Efrat19:
> One simplification on the locking, though: I don't think `pendingWakeup`/`consumerLock` are reachable in practice, so we can drop them (and the lock) entirely.

cc @dalelane since https://github.com/apache/flink-connector-kafka/pull/293 introduces another consumer access from the main thread (to determine the right position to commit on checkpoint)
@AHeise wdyt

Zhuoxi2000:
The architecture test caught one issue on my side: the new `@VisibleForTesting` annotation on `createConsumer` trips the freezing ArchUnit rule, since it only allows the class’s existing violations.

I removed the annotation and kept the seam package-private, with the javadoc explaining why it exists. `ProductionCodeArchitectureTest` and the module tests pass locally now.

@AHeise could you approve the CI run once more when you get a chance?

AHeise:
> > One simplification on the locking, though: I don't think `pendingWakeup`/`consumerLock` are reachable in practice, so we can drop them (and the lock) entirely.
> 
> cc @dalelane since #293 introduces another consumer access from the main thread (to determine the right position to commit on checkpoint) @AHeise wdyt

Let's work it out with @dalelane before merging. Imho it should be fine: if there is no consumer, there are no offsets to commit. I'll also comment on that PR; I think that approach is unnecessarily complex.

Zhuoxi2000:
I looked at how #293 interacts with the lazy consumer, and I think the two changes compose cleanly:

* #293 avoids touching the consumer during offset reconciliation. It uses `lastFetchedOffsets` / `lastKnownPositions`, which are populated from `fetch()`. Both fetching and offset commits run on the split fetcher executor, so this doesn’t introduce any new cross-thread access.
* With lazy creation, a commit before the first fetch creates the consumer on that same fetcher thread, which is exactly what FLINK-36434 wants. Since nothing has been fetched yet, there are no known positions to reconcile, so the offsets pass through unchanged.
* The overlap is mostly mechanical. If #293 lands first, the rebase here should just change `consumer.commitAsync(...)` to `ensureConsumer().commitAsync(...)`.

Happy to rebase either way once the landing order is settled with @dalelane.

Efrat19 (approved):
No farther concerns from my side

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -94,17 +104,49 @@ public KafkaPartitionSplitReader(
...
+    private KafkaConsumer<byte[], byte[]> ensureConsumer() {
+        // Single-writer: only the fetcher thread calls this, so the creation branch needs no
+        // guard; the field is volatile solely for the cross-thread reads in wakeUp() and close().
+        KafkaConsumer<byte[], byte[]> currentConsumer = this.consumer;
+        if (currentConsumer == null) {
+            currentConsumer = createConsumer(consumerProps);
+            maybeRegisterKafkaConsumerMetrics(
+                    consumerProps, kafkaSourceReaderMetrics, currentConsumer);
+            kafkaSourceReaderMetrics.registerNumBytesIn(currentConsumer);
+            this.consumer = currentConsumer;
```
MartijnVisser: Assigning the field straight after `createConsumer` would be a bit safer I think: if metric registration throws, `close()` sees null and never closes the consumer.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -295,7 +355,7 @@ void setConsumerClientRack(Properties consumerProps, String rackId) {
     }
 
     long getConsumerPosition(TopicPartition tp, String msg) {
-        return retryOnWakeup(() -> consumer.position(tp), msg);
+        return retryOnWakeup(() -> ensureConsumer().position(tp), msg);
```
MartijnVisser: These helpers only run after `fetch()` or `handleSplitsChanges()` already ensured the consumer, and `fetch()` reaches this one per assigned partition per poll rather than once. Passing the consumer down keeps creation to the four entry points.

MartijnVisser (commented):
Checked the threading: `splitReader.wakeUp()` has a single call site in `FetchTask#wakeUp`, which can only fire once the fetch task is the running task, and that needs `AddSplitsTask` to have run `handleSplitsChanges()` on the fetcher thread first. Same in flink-connector-base 1.20.5 through 2.4-SNAPSHOT. I also probed the commit path from the ticket rather than only `fetch()`: `KafkaSourceReaderTest#testCommitOffsetsWithoutAliveFetchers` creates both consumers on `main` today and on `Source Data Fetcher for main` here. The new tests are red on eager creation and green on this branch, 50 runs clean.

On the FLINK-33484 overlap: #293's current head reconciles from `lastFetchedOffsets` and `lastKnownPositions`, both filled in `fetch()`, and only reads `consumer.assignment()` inside `notifyCheckpointComplete`, which the fetcher manager enqueues onto the fetcher, so nothing is left on the main thread. Applying both gives one conflict, on `notifyCheckpointComplete`; `ensureConsumer().commitAsync(reconcileOffsetsToCommit(...), callback)` resolves it and the reader suites stay green.

Left before merge: a rebase, main is 19 ahead and FLINK-40367 (#292) rewrote `fetch()` in the same file and left a comment about the constructor not leaking a consumer that no longer holds; the three review commits squashed into the first; and the `Generated-by:` trailer AGENTS.md now asks for. The flink-connector-base follow-up still has no ticket.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
Zhuoxi2000: Done the helpers now take the consumer directly, and `ensureConsumer()` is only used at the SplitReader entry points and in the test accessor.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
Zhuoxi2000: Done  the field is assigned before metric registration now, with a regression test covering the failure case.

Zhuoxi2000:
Thanks for the thorough review and for running the 50-run probe. I've addressed the feedback: rebased onto current `main`, fixed the consumer ownership around metric registration with a regression test, and cleaned up the `ensureConsumer()` usage so it's only called at the SplitReader entry points. I also squashed the changes and linked FLINK-40396 for the connector-base follow-up. The relevant unit tests pass locally; the Testcontainers suites will need CI.

MartijnVisser (approved):
Thanks, both points are in. The ownership fix has a regression test that is red on the old ordering, I ran it. Passing the consumer into the helpers also makes the description's "no" on the per-record path true: it is one volatile read per `fetch()` now, not one per assigned partition.

Re-ran the reader suites on c11c8f2d, all green, plus 50 runs of the new class. CI is green on all six jobs.

FLINK-33484's PR #293 still applies on top, with two small conflicts now. Keep `lastKnownPositions.put` after the reworked position call, and use `ensureConsumer().commitAsync(reconcileOffsetsToCommit(...), callback)` for the other. The reader suites stay green on the combination, so I'd like this one in first and #293 rebased on it.

Two lines in the PR description are out of date: the change log still offers a `@VisibleForTesting createConsumer` seam, and the connector-base follow-up is already filed as FLINK-40396.

Approving.

PR #291, opened 2026-08-10
[FLINK-40362][connector/kafka] Fix idle reader hang when no-more-splits precedes metadata update

slachiewicz:

An idle reader (a subtask with no assigned splits) that receives the no-more-splits signal before the `MetadataUpdateEvent` never forwards that signal to the sub-readers created by that update, so the reader returns `NOTHING_AVAILABLE` indefinitely and a bounded job never finishes.

## Brief change log

- `DynamicKafkaSourceReader.handleSourceEvents` guarded the no-more-splits re-notification with `!pendingSplits.isEmpty()`, which is false in an idle reader.
- Re-deliver the no-more-splits signal after processing the reader's first metadata update (`firstMetadataUpdate = !isActivelyConsumingSplits`). Replay is limited to the first update so that on later metadata changes, an active reader continues waiting for new assignments rather than finishing prematurely. Note that `MetadataUpdateEvent` is the reply to `GetMetadataUpdateEvent` (or periodic discovery), not sent on registration.
- Add `DynamicKafkaSourceReaderTest#testIdleReaderFinishesWhenNoMoreSplitsArrivesBeforeMetadata` and `DynamicKafkaSourceReaderTest#testActiveReaderWaitsForNewSplitsAfterMetadataChange`.

## Verifying this change

- `DynamicKafkaSourceReaderTest#testIdleReaderFinishesWhenNoMoreSplitsArrivesBeforeMetadata` fails on `main` with `expected: END_OF_INPUT but was: NOTHING_AVAILABLE` and passes with this change.
- `DynamicKafkaSourceReaderTest#testActiveReaderWaitsForNewSplitsAfterMetadataChange` fails (with early `END_OF_INPUT`) if replay is unconditional across all metadata updates, and passes with the `firstMetadataUpdate` guard.
- The hang reproduces in `SourceTestSuiteBase#testIdleReader` via `DynamicKafkaSourceITTest$IntegrationTests` whenever a run has an idle subtask, which the random topic suffix decides in roughly 1 run in 20: locally 2 hangs in 8 runs with main's reader, both the runs with an idle subtask; in CI jobs 99324744080 and 102185118941 on the #289 branch.

This change was split out of #285, where it had been committed alongside an unrelated test-flakiness fix.

  - The serializers: no
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Kafka connector, etc: yes, changes when `DynamicKafkaSourceReader` re-delivers the no-more-splits signal to sub-readers

## Documentation

---

Claude Code

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
@@ -378,6 +378,10 @@ public class DynamicKafkaSourceReader<T> implements SourceReader<T, DynamicKafka
             clustersProperties.putAll(newClustersProperties);
         }
 
+        // Captured before the flag flips below, so the no-more-splits replay can tell the
+        // reader's first metadata update from a later metadata change.
+        final boolean firstMetadataUpdate = !isActivelyConsumingSplits;
+
         // finally mark the reader as active, if not already and add pending splits
         if (!isActivelyConsumingSplits) {
             isActivelyConsumingSplits = true;
@@ -399,9 +403,15 @@ public class DynamicKafkaSourceReader<T> implements SourceReader<T, DynamicKafka
 
             addSplits(validPendingSplits);
             pendingSplits.clear();
-            if (isNoMoreSplits) {
-                notifyNoMoreSplits();
-            }
+        }
+
+        // Replay only on the first metadata update. On a later metadata change the reader must
+        // wait for the enumerator to signal again after the new assignments (that re-signal is
+        // currently missing for a sub-enumerator recreated with partitions already assigned; see
+        // FLINK-31006), so replaying here would finish an active reader before the new topic's
+        // splits arrive.
+        if (isNoMoreSplits && firstMetadataUpdate) {
+            notifyNoMoreSplits();
         }
 
         // Releasing the last split output can expose the runtime output as idle immediately. Keep
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReaderTest.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReaderTest.java
@@ -371,6 +371,67 @@ public class DynamicKafkaSourceReaderTest extends SourceReaderTestBase<DynamicKa
     }
 
     @Test
+    void testIdleReaderFinishesWhenNoMoreSplitsArrivesBeforeMetadata() throws Exception {
+        TestingReaderContext context = new TestingReaderContext();
+        try (DynamicKafkaSourceReader<Integer> reader = createReaderWithoutStart(context)) {
+            TrackingReaderOutput<Integer> readerOutput = new TrackingReaderOutput<>();
+            reader.start();
+
+            // The enumerator signals no more splits before the reader receives the metadata
+            // update event, e.g. when an idle reader registers after all bounded
+            // sub-enumerators have already finished split discovery.
+            reader.notifyNoMoreSplits();
+
+            MetadataUpdateEvent metadata =
+                    DynamicKafkaSourceTestHelper.getMetadataUpdateEvent(TOPIC);
+            CompletableFuture<Void> availableBeforeMetadata = reader.isAvailable();
+            reader.handleSourceEvents(metadata);
+
+            assertThat(availableBeforeMetadata)
+                    .as("the metadata update must wake up a task parked on the earlier future")
+                    .isDone();
+            assertThat(reader.pollNext(readerOutput))
+                    .as(
+                            "idle reader must reach END_OF_INPUT even when no-more-splits precedes the metadata update event")
+                    .isEqualTo(InputStatus.END_OF_INPUT);
+        }
+    }
+
+    @Test
+    void testActiveReaderWaitsForNewSplitsAfterMetadataChange() throws Exception {
+        TestingReaderContext context = new TestingReaderContext();
+        try (DynamicKafkaSourceReader<Integer> reader = createReaderWithoutStart(context)) {
+            TrackingReaderOutput<Integer> readerOutput = new TrackingReaderOutput<>();
+            reader.start();
+
+            // First metadata update: only cluster 0 is known, so the reader goes active with a
+            // single sub-reader.
+            KafkaStream clusterZeroOnly = DynamicKafkaSourceTestHelper.getKafkaStream(TOPIC);
+            clusterZeroOnly.getClusterMetadataMap().remove(kafkaClusterId1);
+            reader.handleSourceEvents(
+                    new MetadataUpdateEvent(Collections.singleton(clusterZeroOnly)));
+
+            // The enumerator finished discovery for that metadata epoch.
+            reader.notifyNoMoreSplits();
+
+            // Cluster 1 appears. The reader holds no splits, so the metadata change recreates every
+            // sub-reader and only the reader-level flag still remembers the earlier signal. Splits
+            // and a fresh no-more-splits signal for the new metadata arrive after this event.
+            reader.handleSourceEvents(DynamicKafkaSourceTestHelper.getMetadataUpdateEvent(TOPIC));
+
+            assertThat(reader.pollNext(readerOutput))
+                    .as(
+                            "reader must not finish on the stale no-more-splits signal while a newly added cluster still awaits its splits")
+                    .isEqualTo(InputStatus.NOTHING_AVAILABLE);
+
+            reader.notifyNoMoreSplits();
+            assertThat(reader.pollNext(readerOutput))
+                    .as("reader finishes once the enumerator signals again for the new metadata")
+                    .isEqualTo(InputStatus.END_OF_INPUT);
+        }
+    }
+
+    @Test
     void testAvailabilityFutureUpdates() throws Exception {
         TestingReaderContext context = new TestingReaderContext();
         try (DynamicKafkaSourceReader<Integer> reader = createReaderWithoutStart(context)) {


Efrat19:
> An idle reader (a subtask with no assigned splits) that receives the no-more-splits signal before the `MetadataUpdateEvent` never forwards that signal to the sub-readers created by the update, so the reader returns `NOTHING_AVAILABLE` indefinitely and a bounded job never finishes.

Was this ever reproduced in a real env or at least ITcase?
The code is 3 years old, it is surprising to learn that bounded jobs were never able to complete all this time

slachiewicz:
Thanks for the question, @Efrat19!

### Why bounded jobs worked in most cases (and why this went unnoticed)
Bounded jobs were able to complete in most scenarios because this issue requires three specific conditions to coincide simultaneously:

1. **A bounded stream / batch job** (where a `NoMoreSplitsEvent` is emitted after split discovery).
2. **`parallelism > total partitions`**, meaning at least one reader subtask is an **idle reader** receiving 0 splits (`pendingSplits.isEmpty() == true`).
3. **Specific event arrival order**: The idle reader receives `NoMoreSplitsEvent` (`notifyNoMoreSplits()`) **before** the initial `MetadataUpdateEvent` is handled.

In normal/happy-path executions:
- **Readers with assigned splits** were never affected because `!pendingSplits.isEmpty()` evaluated to `true`, which forwarded `notifyNoMoreSplits()` properly.
- **Idle readers where `MetadataUpdateEvent` arrived first** also completed normally because `clusterReaderMap` was already populated when `notifyNoMoreSplits()` was called.

The hang only occurs when an idle reader registers or handles events in the order `NoMoreSplitsEvent -> MetadataUpdateEvent` (e.g. when an idle subtask finishes registration slightly after the enumerator has already completed partition discovery and broadcasted the signal). In that specific order, `clusterReaderMap` was empty when `notifyNoMoreSplits()` was called, and when `MetadataUpdateEvent` later populated `clusterReaderMap`, the re-notification was skipped due to `pendingSplits.isEmpty()`.

### Reproduction in ITCase vs. Unit Test
- In real/integration environments, this depends on non-deterministic RPC/actor message scheduling. We observed intermittent hangs in `DynamicKafkaSourceITTest#testIdleReader` in CI under heavy load, which pointed to this race condition.
- Because reproducing this via full ITCases is subject to thread scheduling timing, we added the deterministic unit test `DynamicKafkaSourceReaderTest#testIdleReaderFinishesWhenNoMoreSplitsArrivesBeforeMetadata`. It isolates this exact sequence:
  1. Idle reader starts (0 splits).
  2. Receives `notifyNoMoreSplits()`.
  3. Receives `MetadataUpdateEvent`.
  4. On unfixed `main`, `pollNext()` returns `NOTHING_AVAILABLE` indefinitely; with this fix, it cleanly emits `END_OF_INPUT`.

### Impact & Safety
The fix is strictly additive and safe:
- `notifyNoMoreSplits()` is idempotent and only executes if `isNoMoreSplits == true`.
- `MetadataUpdateEvent` is only delivered upon initial reader registration or actual cluster/topic metadata changes (not on checkpoint intervals), so this introduces no recurring cross-component overhead.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
```diff
@@ -385,9 +385,10 @@ public void handleSourceEvents(SourceEvent sourceEvent) {
 
             addSplits(validPendingSplits);
             pendingSplits.clear();
-            if (isNoMoreSplits) {
-                notifyNoMoreSplits();
-            }
+        }
+
+        if (isNoMoreSplits) {
```
MartijnVisser: Replaying on every metadata update makes an active reader finish as soon as its current splits are done, before the splits of a newly added topic arrive. I added `testActiveReaderWaitsForNewSplitsAfterMetadataChange` to check this: it passes on main and fails here with `END_OF_INPUT`.

Limiting the replay to the reader's first metadata update keeps your fix and passes both that test and your `testIdleReaderFinishesWhenNoMoreSplitsArrivesBeforeMetadata`:

```java
final boolean firstMetadataUpdate = !isActivelyConsumingSplits;
...
if (isNoMoreSplits && firstMetadataUpdate) {
    notifyNoMoreSplits();
}
```

For later metadata changes the reader must wait for the enumerator to signal again after the new assignments. That signal is currently missing for a recreated sub-enumerator (FLINK-31006), so an unconditional replay would only mask that as an early finish. Happy to push the test to your branch if you want.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReaderTest.java
```diff
@@ -369,6 +369,34 @@ void testNotifyNoMoreSplits() throws Exception {
...
+            // sub-enumerators have already finished split discovery.
+            reader.notifyNoMoreSplits();
+
+            MetadataUpdateEvent metadata =
+                    DynamicKafkaSourceTestHelper.getMetadataUpdateEvent(TOPIC);
+            reader.handleSourceEvents(metadata);
+
+            long deadline = System.currentTimeMillis() + 10_000;
+            InputStatus status = reader.pollNext(readerOutput);
+            while (status != InputStatus.END_OF_INPUT && System.currentTimeMillis() < deadline) {
```
MartijnVisser: The loop is not needed. After the metadata update the availability future obtained before it is complete and a single `pollNext` returns `END_OF_INPUT`:

```java
CompletableFuture<Void> availableBeforeMetadata = reader.isAvailable();
reader.handleSourceEvents(metadata);
assertThat(availableBeforeMetadata).isDone();
assertThat(reader.pollNext(readerOutput)).isEqualTo(InputStatus.END_OF_INPUT);
```

This also asserts that the task gets woken up, which is what the ITCase depends on.

MartijnVisser (changes requested):
Thanks, the idle reader fix and its unit test are correct. Two changes before I merge this, both with tests I ran locally.

I want this in first, then #289, then #300 rebased on both. #300 needs the reader replay from this PR, otherwise the dedup there turns the idle reader race into a hang.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
slachiewicz: Applied in `4eded31c`, with the capture placed above the `isActivelyConsumingSplits = true` block so the flag still reads the pre-update state.

Verified: removing the `firstMetadataUpdate` guard fails `testActiveReaderWaitsForNewSplitsAfterMetadataChange` and leaves `testIdleReaderFinishesWhenNoMoreSplitsArrivesBeforeMetadata` passing, so the new test pins exactly the early finish you described. Wrote it on the branch rather than taking the push, thanks for the offer.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReaderTest.java
slachiewicz: Applied in `4eded31c` — loop replaced with the pre-update availability future plus a single `pollNext`.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
```diff
@@ -385,9 +389,13 @@ public void handleSourceEvents(SourceEvent sourceEvent) {
...
             addSplits(validPendingSplits);
             pendingSplits.clear();
-            if (isNoMoreSplits) {
-                notifyNoMoreSplits();
-            }
+        }
+
+        // Replay only on the first metadata update. On a later metadata change the reader must
+        // wait for the enumerator to signal again after the new assignments, so replaying here
+        // would finish an active reader before the new topic's splits arrive.
```
MartijnVisser: Nit: the re-signal this comment relies on is missing today for a sub-enumerator recreated with its partitions already assigned (FLINK-31006). Worth one mention so nobody removes the guard as redundant later.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReaderTest.java
```diff
@@ -369,6 +369,61 @@ void testNotifyNoMoreSplits() throws Exception {
...
+            KafkaStream clusterZeroOnly = DynamicKafkaSourceTestHelper.getKafkaStream(TOPIC);
+            clusterZeroOnly.getClusterMetadataMap().remove(kafkaClusterId1);
+            reader.handleSourceEvents(
+                    new MetadataUpdateEvent(Collections.singleton(clusterZeroOnly)));
+
+            // The enumerator finished discovery for that metadata epoch.
+            reader.notifyNoMoreSplits();
+
+            // Cluster 1 appears, which creates a sub-reader the earlier no-more-splits signal
+            // never covered. Its splits are assigned only after the enumerator signals again.
```
MartijnVisser: The enumerator assigns before it signals, and after this event every sub-reader is new, not only cluster 1's, because the reader holds no splits. That is also what makes the single `pollNext` below conclusive.

```suggestion
            // Cluster 1 appears. The reader holds no splits, so the metadata change recreates every
            // sub-reader and only the reader-level flag still remembers the earlier signal. Splits and
            // a fresh no-more-splits signal for the new metadata arrive after this event.
```

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReaderTest.java
```diff
@@ -369,6 +369,61 @@ void testNotifyNoMoreSplits() throws Exception {
...
+            reader.handleSourceEvents(DynamicKafkaSourceTestHelper.getMetadataUpdateEvent(TOPIC));
+
+            assertThat(reader.pollNext(readerOutput))
+                    .as(
+                            "reader must not finish on the stale no-more-splits signal while a newly added cluster still awaits its splits")
+                    .isNotEqualTo(InputStatus.END_OF_INPUT);
```
MartijnVisser: The result here is `NOTHING_AVAILABLE`, nothing else can come out of two fresh sub-readers. Assert that, and add the other half of the contract, that the next signal finishes the reader. Both pass on this branch.

```suggestion
                    .isEqualTo(InputStatus.NOTHING_AVAILABLE);

            reader.notifyNoMoreSplits();
            assertThat(reader.pollNext(readerOutput))
                    .as("reader finishes once the enumerator signals again for the new metadata")
                    .isEqualTo(InputStatus.END_OF_INPUT);
```

MartijnVisser (approved):
The hang reproduces in `SourceTestSuiteBase#testIdleReader` via `DynamicKafkaSourceITTest$IntegrationTests` whenever a run has an idle subtask, which the random topic suffix decides in roughly 1 run in 20: locally 2 hangs in 8 runs with main's reader, both the runs with an idle subtask; in CI jobs 99324744080 and 102185118941 on the #289 branch. Please replace the "not confirmed" paragraph with that.

The change is right, approving. Can you address the comments below and the commit message, then I merge:

The commit body and change log still describe the first revision. Please say the replay is limited to the reader's first metadata update and why, list `testActiveReaderWaitsForNewSplitsAfterMetadataChange` with its red/green, add the component tag to the subject (`[FLINK-40362][connector/kafka] ...`), and add the AI disclosure section the other PRs here carry. Bullet 2: the event is the reply to `GetMetadataUpdateEvent`, not sent on registration.

MartijnVisser:
@Efrat19 It is deterministic, not an RPC race: a subtask registers before its reader asks for metadata, so a subtask that registers after the bounded sub-enumerators finished discovery always gets `NoMoreSplitsEvent` before its first `MetadataUpdateEvent`. With splits that is fine, idle it hangs unless a later registration re-broadcasts. It needs an idle subtask, which is rare, and the reader-side shortcut that used to finish an idle reader on the signal alone went away in cdfa328b (kafka-3.1.0). Reproduced in `testIdleReader`: CI jobs 99324744080 and 102185118941 on the #289 branch, and 2 of 8 local runs with main's reader, exactly the runs with an idle subtask.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReader.java
slachiewicz: Added note regarding missing re-signal in FLINK-31006 in `5ae82e44`.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReaderTest.java
slachiewicz: Applied suggestion in `5ae82e44`.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReaderTest.java
slachiewicz: Applied suggestion in `5ae82e44`.

slachiewicz:
@MartijnVisser Addressed all review items in `5ae82e44` and updated the PR title and description:
- Added component tag `[FLINK-40362][connector/kafka]` to commit subject and PR title.
- Updated commit message and PR description to explain limiting replay to `firstMetadataUpdate`, noted that `MetadataUpdateEvent` is the reply to `GetMetadataUpdateEvent`, and included red/green status for both unit tests.
- Replaced the ITCase paragraph in the PR description with the reproduction details from `SourceTestSuiteBase#testIdleReader`.
- Added the AI disclosure section.
- Updated `DynamicKafkaSourceReader.java` comment with note on FLINK-31006.
- Applied suggested comment and assertion updates in `DynamicKafkaSourceReaderTest.java`.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/reader/DynamicKafkaSourceReaderTest.java
```diff
@@ -369,6 +369,67 @@ void testNotifyNoMoreSplits() throws Exception {
...
+            clusterZeroOnly.getClusterMetadataMap().remove(kafkaClusterId1);
+            reader.handleSourceEvents(
+                    new MetadataUpdateEvent(Collections.singleton(clusterZeroOnly)));
+
+            // The enumerator finished discovery for that metadata epoch.
+            reader.notifyNoMoreSplits();
+
+            // Cluster 1 appears. The reader holds no splits, so the metadata change recreates every
+            // sub-reader and only the reader-level flag still remembers the earlier signal. Splits and
+            // a fresh no-more-splits signal for the new metadata arrive after this event.
```
MartijnVisser:
```suggestion
            // Cluster 1 appears. The reader holds no splits, so the metadata change recreates every
            // sub-reader and only the reader-level flag still remembers the earlier signal. Splits
            // and a fresh no-more-splits signal for the new metadata arrive after this event.
```

MartijnVisser (commented):
Thanks. One more push: my suggested comment was one column too wide, `spotless:check` fails on line 417. Rewrapped below; tests are green on this head (15/15).

MartijnVisser:
@slachiewicz Can you run spotless for CI?

MartijnVisser:
@slachiewicz Any particular reason why the PR is now a draft?

slachiewicz:
@MartijnVisser Rewrapped the comment per your suggestion in `606fd393`, no other change.

I moved it to draft while I took a step back to work out the best merge order for the open PRs and to focus on resolving the main branch failures, which meant reviewing the strategy for that first. Marked ready again now.

MartijnVisser:
@slachiewicz Can you open backports to v5.0, v4.0 and v3.4? The same guard is on all three branch heads, so the hang is there too. You'll need to make an adjustment for v4.0 and v3.4 to use `TestingReaderOutput`

Efrat19:
> @Efrat19 It is deterministic, not an RPC race: a subtask registers before its reader asks for metadata, so a subtask that registers after the bounded sub-enumerators finished discovery always gets `NoMoreSplitsEvent` before its first `MetadataUpdateEvent`. With splits that is fine, idle it hangs unless a later registration re-broadcasts. It needs an idle subtask, which is rare, and the reader-side shortcut that used to finish an idle reader on the signal alone went away in [cdfa328](https://github.com/apache/flink-connector-kafka/commit/cdfa328b5ec34d711ae2c9e93de6de7565fd1db6) (kafka-3.1.0). Reproduced in `testIdleReader`: CI jobs 99324744080 and 102185118941 on the #289 branch, and 2 of 8 local runs with main's reader, exactly the runs with an idle subtask.

Understood, thanks. Not blocking here.

PR #292, opened 2026-08-11
[FLINK-40367][Connectors/Kafka] Make Kafka consumer poll timeout configurable

spuru9:

`KafkaPartitionSplitReader` polls the consumer with a hardcoded 10 second timeout. On idle partitions the split reader only reacts to split changes after the ongoing poll returns, so this can add up to 10 seconds of delay to checkpoints, which is especially painful for jobs using transactional Kafka producers.

This makes the timeout configurable via a new `poll.timeout.ms` source property, defaulting to the previous 10 seconds so existing jobs are unaffected.

## Brief change log

- Add `KafkaSourceOptions.POLL_TIMEOUT_MS` (`poll.timeout.ms`), default 10000 ms.
- `KafkaPartitionSplitReader` parses the property once in the constructor and uses it for `consumer.poll(...)` instead of the `POLL_TIMEOUT` constant. Parsing happens before the `KafkaConsumer` is created so an invalid value cannot leak a consumer, and non-positive values are rejected because they would turn the fetch loop into a busy loop.
- Document the option in the DataStream Kafka docs (English and Chinese). The dynamic Kafka docs already link to that list.

The property flows through the existing `Properties` bag, so it works for `KafkaSource.builder().setProperty(...)`, `DynamicKafkaSource` per-cluster properties, and SQL via `properties.poll.timeout.ms` without additional plumbing.

## Verifying this change

Added to `KafkaPartitionSplitReaderTest`:

- `testConfiguredPollTimeoutIsUsedForPolling`: assigns a split on an empty topic with a 500 ms poll timeout and asserts the fetch returns well below the 10 second default. Verified the assertion is meaningful by temporarily restoring the hardcoded timeout, which makes the test fail.
- `testNonPositivePollTimeoutIsRejected`: parameterized over `0` and `-1`.
- `testDefaultPollTimeoutIsTenSeconds`: guards the unchanged default.

- The serializers: no
- The runtime per-record code paths (performance sensitive): yes, the poll timeout in `fetch()` is now read from a field; behavior is unchanged at the default value

## Documentation

- Does this pull request introduce a new feature? yes, a new source configuration option
- If yes, how is the feature documented? docs

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSourceOptions.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSourceOptions.java
@@ -58,6 +58,18 @@ public class KafkaSourceOptions {
                     .defaultValue(true)
                     .withDescription("Whether to commit consuming offset on checkpoint.");
 
+    public static final ConfigOption<Long> POLL_TIMEOUT_MS =
+            ConfigOptions.key("poll.timeout.ms")
+                    .longType()
+                    .defaultValue(Duration.ofSeconds(10).toMillis())
+                    .withDescription(
+                            "The maximum time in milliseconds the Kafka consumer blocks in a single "
+                                    + "poll() call while waiting for records. Since the poll returns "
+                                    + "as soon as records are available, a smaller value mainly makes "
+                                    + "an idle split reader react faster to split changes, at the cost "
+                                    + "of polling more often. 0 polls without blocking. Must not be "
+                                    + "negative.");
+
     @SuppressWarnings("unchecked")
     public static <T> T getOption(
             Properties props, ConfigOption<?> configOption, Function<String, T> parser) {
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
@@ -64,12 +64,12 @@ import java.util.stream.Collectors;
 public class KafkaPartitionSplitReader
         implements SplitReader<ConsumerRecord<byte[], byte[]>, KafkaPartitionSplit> {
     private static final Logger LOG = LoggerFactory.getLogger(KafkaPartitionSplitReader.class);
-    private static final long POLL_TIMEOUT = 10000L;
 
     private final KafkaConsumer<byte[], byte[]> consumer;
     private final Map<TopicPartition, Long> stoppingOffsets;
     private final String groupId;
     private final int subtaskId;
+    private final Duration pollTimeout;
 
     private final KafkaSourceReaderMetrics kafkaSourceReaderMetrics;
 
@@ -90,6 +90,8 @@ public class KafkaPartitionSplitReader
             String rackIdSupplier) {
         this.subtaskId = context.getIndexOfSubtask();
         this.kafkaSourceReaderMetrics = kafkaSourceReaderMetrics;
+        // Parsed before creating the consumer so that an invalid value does not leak a consumer.
+        this.pollTimeout = parsePollTimeout(props);
         Properties consumerProps = new Properties();
         consumerProps.putAll(props);
         consumerProps.setProperty(ConsumerConfig.CLIENT_ID_CONFIG, createConsumerClientId(props));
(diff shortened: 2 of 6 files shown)


flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReaderTest.java
```diff
@@ -377,6 +379,58 @@ void testPauseOrResumeSplitsWithUnassignedPartition() {
...
+        reader.fetch();
+
+        final long startTime = System.nanoTime();
+        RecordsWithSplitIds<ConsumerRecord<byte[], byte[]>> records = reader.fetch();
+        final Duration fetchDuration = Duration.ofNanos(System.nanoTime() - startTime);
+
+        assertThat(records.nextSplit()).isNull();
+        // Generously below the 10 seconds default, but way above the configured 500ms, so that the
+        // assertion only fails if the configured timeout is ignored
+        assertThat(fetchDuration).isLessThan(Duration.ofSeconds(5));
```
Efrat19: Why 5 seconds and not the configured 500ms?
Also maybe its enough to have a visibleForTesting KafkaPartitionSplitReader#getPollTimeout() instead of testing kafka client behavior here, making the test lighter and more deterministic, wdyt?

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReaderTest.java
```diff
@@ -377,6 +379,58 @@ void testPauseOrResumeSplitsWithUnassignedPartition() {
                 Collections.singletonList(unassignedSplit));
     }
 
+    @Test
+    public void testDefaultPollTimeoutIsTenSeconds() {
```
Efrat19: IMO asserting the default value is used when no other is provided would make this test more valueable, wdyt?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSourceOptions.java
```diff
@@ -58,6 +58,18 @@ public class KafkaSourceOptions {
                     .defaultValue(true)
                     .withDescription("Whether to commit consuming offset on checkpoint.");
 
+    public static final ConfigOption<Long> POLL_TIMEOUT_MS =
+            ConfigOptions.key("poll.timeout.ms")
+                    .longType()
+                    .defaultValue(Duration.ofSeconds(10).toMillis())
+                    .withDescription(
```
Efrat19, line 65: It might be important to also mention that poll returns immediately if there are records available.
([docs](https://kafka.apache.org/25/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#poll-java.time.Duration-))

Efrat19 (commented):
Most needed change! thank you @spuru9 
LGTM % couple notes

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReaderTest.java
Efrat19: see https://github.com/apache/flink-connector-kafka/pull/292#discussion_r3764180777

spuru9:
@Efrat19 Have addressed the comments, can you recheck. Also can you rerun the pipeline if you have access.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReaderTest.java
```diff
@@ -377,6 +379,40 @@ void testPauseOrResumeSplitsWithUnassignedPartition() {
                 Collections.singletonList(unassignedSplit));
     }
 
+    @Test
+    public void testDefaultPollTimeout() {
+        // The default is part of the connector's behavior contract: changing it would silently
+        // change how long the split reader blocks on idle partitions for all existing jobs.
+        assertThat(createReader().getPollTimeout()).isEqualTo(Duration.ofSeconds(10));
```
Efrat19:
```suggestion
        assertThat(createReader().getPollTimeout()).isEqualTo(KafkaSourceOptions.POLL_TIMEOUT_MS.defaultValue());
```

Efrat19 (approved):
Thanks!

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -278,8 +280,25 @@ KafkaConsumer<byte[], byte[]> consumer() {
...
+
     // --------------- private helper method ----------------------
 
+    private static Duration parsePollTimeout(Properties props) {
+        long pollTimeoutMs =
+                KafkaSourceOptions.getOption(
+                        props, KafkaSourceOptions.POLL_TIMEOUT_MS, Long::parseLong);
+        Preconditions.checkArgument(
+                pollTimeoutMs > 0,
+                "Property %s should be positive, but is %s",
```
rmetzger: Why can't the poll timeout be 0? Kafka forbids only negative poll timeouts, and I think having a 0 timeout is a valid use-case

rmetzger (approved):
Thanks, will merge when CI is green

spuru9:
@rmetzger Have made the docs a little shorter. Can you retrigger.

PR #293, opened 2026-08-12
[FLINK-33484][Connectors/Kafka] Fix offsets committed to Kafka

dalelane:

Offsets that the Kafka source commits back to Kafka can lag the
consumer's true position, because entries that the consumer reads
but never delivers as records (e.g. transaction commit / abort
markers) were not accounted for. The result is that the committed
offset was stopping at the last delivered record, while the
consumer had advanced past that.

This wasn't impacting Flink processing guarantees, as restarts
restore from checkpointed offsets, not offsets committed to Kafka.
External tooling that uses the Kafka committed offset for
monitoring would give the appearance of consumer lag. On
low-traffic or bursty transactional topics, this would report lag
that does not exist until new events are received.

## Brief change log

This is now changed to reconcile the offset to commit with the
consumer's position:

- KafkaPartitionSplitReader now tracks, per partition, the offset
  of the last record handed to the reader, and the consumer
  position observed at the end of each fetch()
- At checkpoint commit, reconcileOffsetsToCommit advances the
  offset to commit to the consumer position when the reader is
  caught up to what it delivered, so the committed offset will
  reflects the non-record entries that the consumer read past.
  Reconcilation is skipped for partitions that aren't assigned and
  for bounded splits
- Per-partition tracking state is cleaned up when partitions are
  unassigned
- KafkaSourceReader records the committed-offset metric from the
  offsets that are committed (rather than those requested) so
  Kafka and Flink's committedOffsets metrics agree.

This also changes the split fetcher when enqueueing the
offset-commit task, consistent with what addSplits, removeSplits,
and pauseOrResumeSplits already do. Without this, a commit on an
idle partition cannot be sent until the in-flight consumer.poll()
has timed out.

## Verifying this change

- Added unit tests for the offset commit filter in `KafkaSourceReaderTest` that recreate scenarios and edge cases that could result in offset lag  
- Added unit tests for the split reader in `KafkaPartitionSplitReaderTest` to verify that state is cleared when partitions are unassigned

  - Dependencies (does it add or upgrade a dependency, including `kafka.version` or `flink.version`): no
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)` or `@Experimental`, or are the Table options or the PyFlink wrappers changed: no
  - Checkpointed state, its serializers, or exactly-once delivery (splits, enumerator state, writer state, committables, transactions): no
  - The per-record code paths (split reader, record emitter, writer, serialization schemas; performance sensitive): yes 

## Documentation

  - If the docs changed, are both `docs/content` and `docs/content.zh` updated? not applicable

---

Generated-by: Claude Code (Sonnet 5)

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
@@ -79,6 +79,9 @@ public class KafkaPartitionSplitReader
     // Offset of the last record that fetch() handed to the source reader, per partition
     private final Map<TopicPartition, Long> lastFetchedOffsets = new HashMap<>();
 
+    // Consumer position observed at the end of the most recent fetch(), per partition.
+    private final Map<TopicPartition, Long> lastKnownPositions = new HashMap<>();
+
     public KafkaPartitionSplitReader(
             Properties props,
             SourceReaderContext context,
@@ -130,6 +133,7 @@ public class KafkaPartitionSplitReader
         for (TopicPartition tp : consumer.assignment()) {
             long stoppingOffset = getStoppingOffset(tp);
             long consumerPosition = getConsumerPosition(tp, "retrieving consumer position");
+            lastKnownPositions.put(tp, consumerPosition);
             // Stop fetching when the consumer's position reaches the stoppingOffset.
             // Control messages may follow the last record; therefore, using the last record's
             // offset as a stopping condition could result in indefinite blocking.
@@ -343,6 +347,11 @@ public class KafkaPartitionSplitReader
      * the offset stops at the first entry that Kafka does not deliver, for as long as the partition
      * is idle. The consumer's own position accounts for those entries, so it is the offset that
      * external tooling expects to see.
+     *
+     * <p>This relies on {@link #lastKnownPositions}, populated as a side effect of the regular
+     * {@link #fetch()} poll loop, rather than querying the consumer for the position again here.
+     * {@link #notifyCheckpointComplete} runs on this same split fetcher thread, but at a point
+     * outside that poll loop, so this allows us to avoid a separate blocking call to the consumer.
      */
     private Map<TopicPartition, OffsetAndMetadata> reconcileOffsetsToCommit(
             Map<TopicPartition, OffsetAndMetadata> offsetsToCommit) {
@@ -358,8 +367,8 @@ public class KafkaPartitionSplitReader
                             || offsetAndMetadata.offset() != lastFetchedOffset + 1) {
                         return;
                     }
-                    long position = getConsumerPosition(tp, "reconciling the offset to commit");
-                    if (position > offsetAndMetadata.offset()) {
+                    Long position = lastKnownPositions.get(tp);
+                    if (position != null && position > offsetAndMetadata.offset()) {
                         LOG.debug(
                                 "Advancing offset to commit for {} from {} to the consumer position {}.",
                                 tp,
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
@@ -76,6 +76,9 @@ public class KafkaPartitionSplitReader
     // Tracking empty splits that has not been added to finished splits in fetch()
     private final Set<String> emptySplits = new HashSet<>();
 
+    // Offset of the last record that fetch() handed to the source reader, per partition
+    private final Map<TopicPartition, Long> lastFetchedOffsets = new HashMap<>();
+
     public KafkaPartitionSplitReader(
             Properties props,
             SourceReaderContext context,
@@ -149,6 +152,8 @@ public class KafkaPartitionSplitReader
                         trackTp -> {
                             kafkaSourceReaderMetrics.maybeAddRecordsLagMetric(consumer, trackTp);
                         });
+
+        trackLastFetchedRecordOffsets(consumerRecords);
 
         markEmptySplitsAsFinished(recordsBySplits);
 
@@ -272,7 +277,7 @@ public class KafkaPartitionSplitReader
     public void notifyCheckpointComplete(
             Map<TopicPartition, OffsetAndMetadata> offsetsToCommit,
             OffsetCommitCallback offsetCommitCallback) {
-        consumer.commitAsync(offsetsToCommit, offsetCommitCallback);
+        consumer.commitAsync(reconcileOffsetsToCommit(offsetsToCommit), offsetCommitCallback);
     }
 
     @VisibleForTesting
@@ -318,6 +323,53 @@ public class KafkaPartitionSplitReader
 
     long getConsumerPosition(TopicPartition tp, String msg) {
         return retryOnWakeup(() -> consumer.position(tp), msg);
+    }
+
+    private void trackLastFetchedRecordOffsets(ConsumerRecords<byte[], byte[]> consumerRecords) {
+        for (TopicPartition tp : consumerRecords.partitions()) {
+            List<ConsumerRecord<byte[], byte[]>> partitionRecords = consumerRecords.records(tp);
+            if (!partitionRecords.isEmpty()) {
+                lastFetchedOffsets.put(
+                        tp, partitionRecords.get(partitionRecords.size() - 1).offset());
+            }
+        }
+    }
+
+    /**
+     * Advances the offsets to commit over the entries that the Kafka consumer read but never
+     * delivered, such as transaction control markers and records of aborted transactions.
+     *
+     * <p>{@link KafkaRecordEmitter} derives the offset to commit from the records it receives, so
+     * the offset stops at the first entry that Kafka does not deliver, for as long as the partition
+     * is idle. The consumer's own position accounts for those entries, so it is the offset that
+     * external tooling expects to see.
+     */
+    private Map<TopicPartition, OffsetAndMetadata> reconcileOffsetsToCommit(
+            Map<TopicPartition, OffsetAndMetadata> offsetsToCommit) {
+        Map<TopicPartition, OffsetAndMetadata> reconciled = new HashMap<>(offsetsToCommit);
+        Set<TopicPartition> assignment = consumer.assignment();
+        offsetsToCommit.forEach(
+                (tp, offsetAndMetadata) -> {
+                    if (!assignment.contains(tp) || stoppingOffsets.containsKey(tp)) {
+                        return;
+                    }
+                    Long lastFetchedOffset = lastFetchedOffsets.get(tp);
+                    if (lastFetchedOffset == null
+                            || offsetAndMetadata.offset() != lastFetchedOffset + 1) {
+                        return;
+                    }
+                    long position = getConsumerPosition(tp, "reconciling the offset to commit");
+                    if (position > offsetAndMetadata.offset()) {
+                        LOG.debug(
+                                "Advancing offset to commit for {} from {} to the consumer position {}.",
+                                tp,
+                                offsetAndMetadata.offset(),
+                                position);
+                        reconciled.put(
+                                tp, new OffsetAndMetadata(position, offsetAndMetadata.metadata()));
+                    }
+                });
+        return reconciled;
     }
 
     private void parseStartingOffsets(
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
@@ -373,8 +373,8 @@ public class KafkaPartitionSplitReader
                         return;
                     }
                     Long lastFetchedOffset = lastFetchedOffsets.get(tp);
-                    if (lastFetchedOffset == null
-                            || offsetAndMetadata.offset() != lastFetchedOffset + 1) {
+                    if (lastFetchedOffset != null
+                            && offsetAndMetadata.offset() != lastFetchedOffset + 1) {
                         return;
                     }
                     Long position = lastKnownPositions.get(tp);
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
@@ -294,6 +294,16 @@ public class KafkaPartitionSplitReader
         return pollTimeout;
     }
 
+    @VisibleForTesting
+    Map<TopicPartition, Long> lastFetchedOffsets() {
+        return lastFetchedOffsets;
+    }
+
+    @VisibleForTesting
+    Map<TopicPartition, Long> lastKnownPositions() {
+        return lastKnownPositions;
+    }
+
     // --------------- private helper method ----------------------
 
     private static Duration parsePollTimeout(Properties props) {
@@ -520,6 +530,11 @@ public class KafkaPartitionSplitReader
         Collection<TopicPartition> newAssignment = new HashSet<>(consumer.assignment());
         newAssignment.removeAll(partitionsToUnassign);
         consumer.assign(newAssignment);
+        partitionsToUnassign.forEach(
+                tp -> {
+                    lastFetchedOffsets.remove(tp);
+                    lastKnownPositions.remove(tp);
+                });
     }
 
     private String createConsumerClientId(Properties props) {
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReader.java
@@ -159,8 +159,9 @@ public class KafkaSourceReader<T>
                                         "Successfully committed offsets for checkpoint {}",
                                         checkpointId);
                                 kafkaSourceReaderMetrics.recordSucceededCommit();
-                                // offsets committed to Kafka can differ from what was requested, 
-                                // because the split reader reconciles them with the consumer position
+                                // offsets committed to Kafka can differ from what was requested,
+                                // because the split reader reconciles them with the consumer
+                                // position
                                 committedOffsets.forEach(
                                         (tp, offset) ->
                                                 kafkaSourceReaderMetrics.recordCommittedOffset(
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReader.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReader.java
@@ -145,7 +145,7 @@ public class KafkaSourceReader<T>
         ((KafkaSourceFetcherManager) splitFetcherManager)
                 .commitOffsets(
                         committedPartitions,
-                        (ignored, e) -> {
+                        (committedOffsets, e) -> {
                             // The offset commit here is needed by the external monitoring. It won't
                             // break Flink job's correctness if we fail to commit the offset here.
                             if (e != null) {
@@ -159,9 +159,9 @@ public class KafkaSourceReader<T>
                                         "Successfully committed offsets for checkpoint {}",
                                         checkpointId);
                                 kafkaSourceReaderMetrics.recordSucceededCommit();
-                                // If the finished topic partition has been committed, we remove it
-                                // from the offsets of the finished splits map.
-                                committedPartitions.forEach(
+                                // offsets committed to Kafka can differ from what was requested, 
+                                // because the split reader reconciles them with the consumer position
+                                committedOffsets.forEach(
                                         (tp, offset) ->
                                                 kafkaSourceReaderMetrics.recordCommittedOffset(
                                                         tp, offset.offset()));
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReaderTest.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReaderTest.java
@@ -381,7 +381,20 @@ public class KafkaSourceReaderTest extends SourceReaderTestBase<KafkaPartitionSp
                         6L,
                         4,
                         10,
-                        (RecordProducer) topic -> produceInterleavedTransactions(topic, 0, 2)));
+                        (RecordProducer) topic -> produceInterleavedTransactions(topic, 0, 2)),
+                // a single aborted transaction 
+                // so the partition never delivers a record to the reader
+                // offset 0 = record
+                // offset 1 = record
+                // offset 2 = record
+                // offset 3 = abort marker
+                // offset 4 << log end offset
+                Arguments.of(
+                        "OnlyAbortedTransaction",
+                        4L,
+                        0,
+                        10,
+                        (RecordProducer) topic -> produceInTransaction(topic, 0, 3, false)));
     }
 
     @ParameterizedTest(name = "{0}")
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReaderTest.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReaderTest.java
@@ -457,6 +457,92 @@ public class KafkaSourceReaderTest extends SourceReaderTestBase<KafkaPartitionSp
                 .isEqualTo(2);
     }
 
+    @Test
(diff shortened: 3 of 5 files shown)


flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReaderTest.java
```diff
@@ -266,6 +273,340 @@ void testOffsetCommitOnCheckpointComplete() throws Exception {
         }
     }
 
+    @Test
+    void testCommittedOffsetConvergesWithLogEndOffsetForTransactionalTopic() throws Throwable {
```
Efrat19: Do you think we could consider a matrix / reusable method for the `await-stable-offset → assert-readable-count → read-and-commit → assert-convergence` for these tests to improve readability?
```
- testCommittedOffsetConvergesWithLogEndOffsetForTransactionalTopic
- testCommittedOffsetConvergesWithLogEndOffsetForNonTransactionalTopic
- testCommittedOffsetConvergesWithLogEndOffsetForMultiRecordTransaction
- testCommittedOffsetConvergesWithLogEndOffsetAfterAbortedTransaction
- testCommittedOffsetConvergesWithLogEndOffsetForMultipleTransactions
- testCommittedOffsetConvergesWithLogEndOffsetForInterleavedTransactions
```
They seem to follow the same pattern
claude example suggestion:
```
@FunctionalInterface
private interface RecordProducer {
    void produce(String topic) throws Exception;
}

private static Stream<Arguments> convergenceScenarios() {
    return Stream.of(
            // scenarioName, expectedStableOffset, expectedReadable, maxCheckpoints, producer
            Arguments.of(
                    "NonTransactional", 1L, 1, 1,
                    (RecordProducer)
                            topic ->
                                    KafkaSourceTestEnv.produceToKafka(
                                            Collections.singletonList(
                                                    new ProducerRecord<>(
                                                            topic, 0, topic + "-key", 0)))),
            Arguments.of(
                    "Transactional", 2L, 1, 1,
                    (RecordProducer)
                            topic ->
                                    KafkaSourceTestEnv.produceToKafka(
                                            Collections.singletonList(
                                                    new ProducerRecord<>(
                                                            topic, 0, topic
                                            kafkaServer.getTransactionalProducerConfig())),
            Arguments.of(
                    "MultiRecordTransaction", 4L, 3, 10, 
                    (RecordProducer) topic -> produceInTransaction(topic, 0,
            Arguments.of(
                    "AbortedTransaction", 7L, 2, 10,
                    (RecordProducer)
                            topic -> {
                                produceInTransaction(topic, 0, 2, true);
                                produceInTransaction(topic, 0, 3, false);
                            }),
            Arguments.of(
                    "MultipleTransactions", 6L, 3, 10,
                    (RecordProducer)
                            topic -> {
                                produceInTransaction(topic, 0, 1, true);
                                produceInTransaction(topic, 0, 1, true);
                                produceInTransaction(topic, 0, 1, true);
                            }),
            Arguments.of(
                    "InterleavedTransactions", 6L, 4, 10,
                    (RecordProducer) topic -> produceInterleavedTransactions
}

@ParameterizedTest(name = "{0}")
@MethodSource("convergenceScenarios")
void testCommittedOffsetConvergesWithLogEndOffset(
        String scenarioName,
        long expectedStableOffset,
        int expectedReadable,
        int maxCheckpoints,
        RecordProducer produce)
        throws Throwable {
    final String topic = TOPIC + scenarioName;
    final String groupId = scenarioName + "OffsetCommitGroup";
    final TopicPartition tp = new TopicPartition(topic, 0);
    KafkaSourceTestEnv.createTestTopic(topic, 1, 1);

    produce.produce(topic);

    final long lastStableOffset = awaitLastStableOffset(tp, expectedStableOffset);
    assertThat(getReadCommittedRecordsCount(tp))
            .as("Expected readable record count")
            .isEqualTo(expectedReadable);

    final long committedOffset =
            readAndCommitOffset(tp, groupId, expectedReadable, maxCheckpoints, lastStableOffset);
    assertThat(committedOffset)
            .as("The committed offset should converge with the log end offset")              .isEqualTo(lastStableOffset);
}
```

Efrat19 (commented):
LGTM % a note about tests readability

Efrat19:
Also FYI https://github.com/apache/flink-connector-kafka/pull/290 - not a blocker but would require synchronization here as well

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReaderTest.java
dalelane: thanks for the review, @Efrat19 

That does reduce the copy/pasted logic duplication across my tests - applied in ce28b75b7f7a9b288bb1b31ffd975fae0255be27

Efrat19 (approved):
LGTM

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -125,6 +131,7 @@ public RecordsWithSplitIds<ConsumerRecord<byte[], byte[]>> fetch() throws IOExce
         for (TopicPartition tp : consumer.assignment()) {
             long stoppingOffset = getStoppingOffset(tp);
             long consumerPosition = getConsumerPosition(tp, "retrieving consumer position");
+            lastKnownPositions.put(tp, consumerPosition);
```
AHeise, line 136: Sorry for being late to the party. Have you considered adding this information simply to KafkaPartitionSplitRecords? Then we can use it in KafkaSourceReader#offsetsToCommit without having another flow of data.

AHeise (commented):
Quick idea on an alternative solution.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
dalelane: @AHeise - Thanks for the review.

I went with the split reader to keep track of the consumer's position as it polls, so I could correct offsets as they're committed to Kafka. By leaving it to the last minute like this, when the checkpoint completes the split reader can take the offset the reader has reached, and correct it using the most recent position the consumer reported. 

I think what you're describing would identify the position earlier, as records are emitted. By the time the checkpoint starts, the value would be fixed, so the commit will send what has already been decided. 

The difference comes down to when we get the offset right. Deciding late means the committed offset will always reflect the most recent poll and will always be right at every checkpoint. Deciding early means it could be one poll out of date, as corrections will land once the reader has read past the records it already has, so it needs another record to arrive to carry an updated position through.  

For busy topics, no-one will notice the difference. But what prompted me to pick up this issue was a problem report on a low-traffic / bursty transactional topic, where the offset effectively reported an incorrect lag "forever". (In hindsight, I should've written a test for this scenario, then I could point at it now! That's my bad, and even if we keep my design I should add a test for that.) 

The question is whether solving that issue is worth the cost/complexity of my approach. IMHO it is, but I'm open to debate :-) 

(_@AHeise - I'm aware I'm making a bunch of assumptions from your description, so it's entirely possible I've misunderstood what you're suggesting. Would a quick agent-driven spike to give it a try and see what an implementation would look like help?_)

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -270,7 +279,7 @@ public void pauseOrResumeSplits(
     public void notifyCheckpointComplete(
             Map<TopicPartition, OffsetAndMetadata> offsetsToCommit,
             OffsetCommitCallback offsetCommitCallback) {
-        consumer.commitAsync(offsetsToCommit, offsetCommitCallback);
+        consumer.commitAsync(reconcileOffsetsToCommit(offsetsToCommit), offsetCommitCallback);
```
Savonitar, line 284: reconcileOffsetsToCommit creates a new map, so the offsets that we send to Kafka here can differ from the original committedPartitions map from [KafkaSourceReader](https://github.com/apache/flink-connector-kafka/blob/main/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReader.java#L126) (that callback [ignores](https://github.com/apache/flink-connector-kafka/blob/main/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReader.java#L148)  its first argument and updates offsets from original map). That means Kafka will report 2 while Flink's committed-offset gauge will report 1. Which is a contradiction. 

Could we update the callback to record metrics from its offsets argument?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -298,6 +307,58 @@ long getConsumerPosition(TopicPartition tp, String msg) {
...
+            Map<TopicPartition, OffsetAndMetadata> offsetsToCommit) {
+        Map<TopicPartition, OffsetAndMetadata> reconciled = new HashMap<>(offsetsToCommit);
+        Set<TopicPartition> assignment = consumer.assignment();
+        offsetsToCommit.forEach(
+                (tp, offsetAndMetadata) -> {
+                    if (!assignment.contains(tp) || stoppingOffsets.containsKey(tp)) {
+                        return;
+                    }
+                    Long lastFetchedOffset = lastFetchedOffsets.get(tp);
+                    if (lastFetchedOffset == null
```
Savonitar: Can lastFetchedOffset be legimately null when the consumer position has advanced? e.g. read_committed consumer position moves past entries without delivering anything. 
example:
partition has: 0-5 offsets as aborted records, 6 offset abort marker, 7 offset LSO.
consumer position: 7
lastFetchedOffsets will be null (nothing was delivered)
offset that will be commited: 0

As a result: The reader is fully caught up (position 7 equals the LSO) meanwhile every checkpoint commits 0. Monitoring shows a lag of 7, forever, until some producer finally writes a committed record.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -76,6 +76,12 @@ public class KafkaPartitionSplitReader
     // Tracking empty splits that has not been added to finished splits in fetch()
     private final Set<String> emptySplits = new HashSet<>();
 
+    // Offset of the last record that fetch() handed to the source reader, per partition
+    private final Map<TopicPartition, Long> lastFetchedOffsets = new HashMap<>();
+
+    // Consumer position observed at the end of the most recent fetch(), per partition.
+    private final Map<TopicPartition, Long> lastKnownPositions = new HashMap<>();
```
Savonitar, line 83: Do (when?) we cleanup these entries? Or we never cleanup them?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -298,6 +307,58 @@ long getConsumerPosition(TopicPartition tp, String msg) {
...
+     * {@link #notifyCheckpointComplete} runs on this same split fetcher thread, but at a point
+     * outside that poll loop, so this allows us to avoid a separate blocking call to the consumer.
+     */
+    private Map<TopicPartition, OffsetAndMetadata> reconcileOffsetsToCommit(
+            Map<TopicPartition, OffsetAndMetadata> offsetsToCommit) {
+        Map<TopicPartition, OffsetAndMetadata> reconciled = new HashMap<>(offsetsToCommit);
+        Set<TopicPartition> assignment = consumer.assignment();
+        offsetsToCommit.forEach(
+                (tp, offsetAndMetadata) -> {
+                    if (!assignment.contains(tp) || stoppingOffsets.containsKey(tp)) {
```
Savonitar, line 372: Is it safe to skip reconcilation for splits with stopped offsets? Could you please clarify why?

Savonitar (commented):
Hi. thanks for the PR. 
I left some comments, could you please take a look?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
dalelane: That's a good point, sorry - added cleanup in 7e962f7ea1d4297206453d35c6a9f93ac5d1a6e4

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
dalelane: of course! I hadn't thought of that - changed the check to allow null lastFetchedOffset values 
3121818a8b0ee1ae7a8793083e268a2f256a0bbe

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
dalelane: Thanks for the fix - that does the trick. Added in ae32fafcbd4d9c77e4c4f4bbb011bfaf52762f3a

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
dalelane: I think we need it to protect against the offset going backwards, but you've got me doubting myself now :) 

I'll see if I can come up with a test scenario that properly depends on the check

dalelane: added an example test in 984d50b6d905a1883bb99676a6fd3b6cde5b99a3 
It's a bit contrived - had to resort to asking Claude to set it up

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
dalelane: > I should've written a test for this scenario, then I could point
> at it now! That's my bad, and even if we keep my design I 
> should add a test for that.

I've added that now

https://github.com/apache/flink-connector-kafka/pull/293/changes/9e524f851a03b6c47fde001a637201d036dabe87#diff-912e48531349dc6ecdd8a8a2c25c5b65d4f3511bdd028a0e1cdb5605350759f3R460-R501

dalelane:
I don't think the failing test in the https://github.com/apache/flink-connector-kafka/actions/runs/34335588824/job/102416409039?pr=293 workflow is related to my change. I've created a separate pull request at https://github.com/apache/flink-connector-kafka/pull/308 for that

dalelane:
@AHeise @Savonitar Would you mind having another look at this one please? I think I've addressed all the outstanding issues, but please let me know if you feel that I've missed something

dalelane:
plus if someone could re-run the checks please, I think the tests in main are working now so hopefully that'll get rid of the unrelated red cross

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
```diff
@@ -285,6 +294,16 @@ Duration getPollTimeout() {
         return pollTimeout;
     }
 
+    @VisibleForTesting
+    Map<TopicPartition, Long> lastFetchedOffsets() {
```
MartijnVisser: AGENTS.md rules out adding lines to the violation stores. Package-private fields without these two accessors keep it unchanged, like `getConsumerPosition` below.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReader.java
```diff
@@ -159,9 +159,10 @@ public void notifyCheckpointComplete(long checkpointId) throws Exception {
                                         "Successfully committed offsets for checkpoint {}",
                                         checkpointId);
                                 kafkaSourceReaderMetrics.recordSucceededCommit();
-                                // If the finished topic partition has been committed, we remove it
-                                // from the offsets of the finished splits map.
-                                committedPartitions.forEach(
+                                // offsets committed to Kafka can differ from what was requested,
```
MartijnVisser, line 162: The comment you replaced explained the `removeIf` below, not the metric loop. Worth keeping that one too.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReaderTest.java
```diff
@@ -695,6 +1156,72 @@ private long getCommittedOffsetMetric(TopicPartition tp, MetricListener listener
...
+     * transaction coordinator propagates it to partition leaders asynchronously, so LSO can briefly
+     * lag behind a committed transaction. To avoid introducing a test race condition, this method
+     * checks again after a brief wait.
+     */
+    private static long awaitLastStableOffset(TopicPartition tp, long expectedOffset)
+            throws Exception {
+        try (KafkaConsumer<String, String> probe = createReadCommittedProbe()) {
+            final List<TopicPartition> partitions = Collections.singletonList(tp);
+            long lastStableOffset = probe.endOffsets(partitions).get(tp);
+            final long deadline = System.currentTimeMillis() + 3_000L;
```
MartijnVisser: Three seconds is tight for marker propagation on a loaded runner, and this asserts rather than retries. Thirty seconds costs nothing here.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReaderTest.java
```diff
@@ -266,6 +277,456 @@ void testOffsetCommitOnCheckpointComplete() throws Exception {
         }
     }
 
+    /** Writes the records that a {@link #offsetConvergenceScenarios} scenario needs. */
+    @FunctionalInterface
+    private interface RecordProducer {
+        void produce(String topic) throws Throwable;
+    }
+
+    private static Stream<Arguments> offsetConvergenceScenarios() {
```
MartijnVisser, line 286: 16 tests in 34s becomes 26 in 173s here. Do all seven earn that? `MultiRecordTransaction`, `MultipleTransactions` and `InterleavedTransactions` exercise the same skip as `Transactional`.

MartijnVisser (commented):
I took each production hunk out in turn on this head and reran the two suites. Without the reconciliation, six of the seven scenarios and the idle-partition test fail. Without the `lastFetchedOffset + 1` guard, `testCommittedOffsetDoesNotOverrunRecordsInATransaction` fails. Without the `stoppingOffsets` guard, `testCommittedOffsetForBoundedSplitDoesNotGoBackwards` fails. Without the cleanup in `unassignPartitions`, the split reader test fails. Without `committedOffsets` in the commit callback, six scenarios fail on the metric assertion. Every hunk is covered, and `NonTransactional` passes on the base as the control. All 45 tests pass here, on a rebase onto current main as well, and the new ones came through eight repeats clean.

On threading: `KafkaSourceFetcherManager.commitOffsets` enqueues `notifyCheckpointComplete` onto the split fetcher, so the two new maps and `consumer.assignment()` are only ever touched from the thread that runs `fetch()`. Nothing new crosses a thread boundary.

The red cross is `SourceTopicIntegrityTest`, not this change, and FLINK-40618 has since moved that class to the integration-test phase on main. A rebase clears it and re-triggers CI by itself, so there is no re-run to ask for.

Left before merge: the ten commits squashed into one `[FLINK-33484][Connectors/Kafka]` commit, the `Co-Authored-By: Claude` trailer replaced by `Generated-by: <tool and version>` (https://www.apache.org/legal/generative-tooling.html), and the PR template that landed on main last week filled in, including the AI disclosure checkbox. The Jira needs a release note too, since the committed offset is what people read as lag.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
MartijnVisser: `KafkaPartitionSplitRecords.nextSplit()` walks `consumerRecords.partitions()`, so a partition with no records in a poll never reaches the emitter and its state never moves. That is the idle case here.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
MartijnVisser: The guard is needed. Without it `testCommittedOffsetForBoundedSplitDoesNotGoBackwards` goes red: once the split finishes the offset comes from `offsetsOfFinishedSplits`, unreconciled, so it drops back.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReader.java
dalelane: Sorry about that - restored in 6d1130708ae3e37cbb05ce9c30b7235f7534a7e5

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReaderTest.java
dalelane: Increased in 620d6988deb45493fd73e855f5d694d8b9a48e04

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/reader/KafkaPartitionSplitReader.java
dalelane: That's cleaner, thanks - added in ef50104e963038267fd7040e2e8d13870ff0c03c

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReaderTest.java
dalelane: My initial reaction was that I would like to keep all seven tests. They were written with a view to describing scenarios that felt likely to regress when this area of code is rewritten or refactored in future - even if today they go through the same code. 

That duration is still hard to defend, and is a bad smell, so I spent a bit of time tracing through to work out why these tests were taking so long. 

The delay was coming from `KafkaSourceFetcherManager.enqueueOffsetsCommitTask` which was enqueuing the commit task. It introduced a poll.timeout.ms delay per commit whenever the split is idle, because the fetcher was blocked in the consumer.poll. Across the test scenarios the 10 secs every time added up to a very slow batch of tests.

I think this was highlighting a real issue, not just a test problem. Looking at other places where tasks are enqueued onto live fetchers, this is paired with a wakeup. I've added a wakeup call in e44287bedc002e0b88379beba44fcac7afbdf4ab and that does bring the time for KafkaSourceReaderTest back down to ~49s for me, which feels like a more reasonable increase.

dalelane:
@MartijnVisser Thanks very much for the thorough review - that was helpful and clarified a few things I wasn't certain about.

re: housekeeping stuff

I've squashed the commits into bfa41ed378ba7e7266d3a31845e2abd3d2bea723 and added the `Generated-by` line to reflect the test commit contribution.

I've rewritten the PR description above using the new template. 

I don't see a way to add a release note in JIRA - possibly I'm just looking in the wrong place, but my guess is that I don't have the right permissions to do this. I've added a comment with proposed release note text to the JIRA issue instead.

dalelane:
rebased to reflect new upstream changes - rebase delta in 

```
git range-diff bfa41ed3~1..bfa41ed3 4076a1ab~1..4076a1ab
```

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/source/reader/KafkaSourceReaderTest.java
dalelane: Frustratingly, my wakeUp addition broke the CI run (an error I can't reproduce locally). 

My gut says that it's probably a problem with my test being overly simplistic, but I'll try and set up a local mini-CI in a tiny constrained VM and reproduce the CI results before I try and fix it.

dalelane: running in a very small VM, I was able to reproduce the CI failure

KafkaSourceReaderTest.completeCheckpoint and readAndCommitOffset were waiting for an offset commit to reach Kafka without polling the reader. I was getting away with this while each fetcher cycle cost a full poll.timeout.ms, but by adding the wakeUp I gave myself a race. 

I've added a poll into both loops and now the tests consistently pass even in a constrained VM.

PR #295, opened 2026-08-25
[FLINK-40469] Checkpoint during recovery window silently loses partitions

Savonitar:

Under `SupportsSplitReassignmentOnRecovery`, restored readers hold no splits and report them to `DynamicKafkaSourceEnumerator`, which defers reassignment until the first metadata discovery completes and all readers register. A checkpoint completing inside that window captured no record of the reported splits: the readers' state was empty and the sub enumerator states still marked the partitions assigned. Restoring from such a checkpoint silently stopped consuming those partitions.

This change checkpoints the pending reported splits so a restore from such a checkpoint behaves as if the readers had just reported them.

Might be relevant to #278.

## Brief change log

- Add KafkaSinkRecoveryITCase with controlled record emission and manually triggered checkpoints.
- Complete checkpoint C1, keep C2 and C3 incomplete, and verify through Kafka that C1’s transactional ID has been reused.
- Cancel the job and restore C1 with the same transactional ID prefix and operator identities.
- Assert the committed output after recovery and verify that the final records appear exactly once.

## Verifying this change

- Recovery tests cover checkpoints taken before reassignment, preserving reader-reported offsets, rescaling, retained splits, and metadata changes.
- Gate tests verify report merging and snapshot isolation.
- Serializer tests cover V4 round-trips and compatibility with V1–V3.
- Regression tests fail when the fix is removed and pass with it applied.

  - Dependencies (does it add or upgrade a dependency, including `kafka.version` or `flink.version`): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)` or `@Experimental`, or are the Table options or the PyFlink wrappers changed: (no)
  - Checkpointed state, its serializers, or exactly-once delivery (splits, enumerator state, writer state, committables, transactions): (yes,  persists pending reported splits in enumerator state and upgrades DynamicKafkaSourceEnumStateSerializer to V4. V1–V3 remain readable, older connector versions cannot read V4 checkpoints.)
  - The per-record code paths (split reader, record emitter, writer, serialization schemas; performance sensitive): (no)

## Documentation

  - If the docs changed, are both `docs/content` and `docs/content.zh` updated? (not applicable)

---

Generated-by: Codex CLI 0.155.0-alpha.2.6


Savonitar:
@bowenli86 Could you take a look when you get a chance? This fixes a silent data-loss case in the DynamicKafkaSource recovery flow (might be relevant to https://github.com/apache/flink-connector-kafka/pull/278)
A checkpoint that completes while reported splits are still pending reassignment loses those splits on restore. You seem to be the most active maintainer of the dynamic module, so your review would be much appreciated.

slachiewicz:
Heads-up on overlap with #289 ([FLINK-40137](https://issues.apache.org/jira/browse/FLINK-40137), approved and ready to merge): it extracts the reader-recovery gating state from `DynamicKafkaSourceEnumerator` — including `pendingReportedSplitsByReader` — into a dedicated `ReaderRecoveryGate` class. This PR writes that field directly (constructor restore, `addReader` merge, snapshot getter), so the two will conflict.

Once #289 lands, the rebase here should be mechanical: restore becomes `readerRecoveryGate.recordReportedSplits(...)` per restored reader, and the snapshot path needs a non-draining accessor on the gate (it currently only exposes `drainReportedSplits`). The extraction should actually make the merge-on-downscale semantics here easier to unit-test.

Also worth noting for coordination: #279 adds a new `DynamicKafkaSourceEnumeratorRecoveryTest.java` with the same filename this PR modifies.

gaoxin-oai (approved):
LGTM

lnbest0707 (approved):
LGTM

bowenli86:
kicked off a new round of CI

bowenli86:
Thanks for fixing this—the pending-state approach looks sound. Could we also clarify the reader-only restart case: after a reader is unregistered but before its restored-split report arrives, does Flink’s runtime ordering prevent a checkpoint from completing? The new state covers reports already received. A test or documented runtime guarantee for the pre-registration window would help establish the boundary of this fix.

Savonitar:
Hi @bowenli86 . 

> Could we also clarify the reader-only restart case: after a reader is unregistered but before its restored-split report arrives, does Flink’s runtime ordering prevent a checkpoint from completing? The new state covers reports already received. A test or documented runtime guarantee for the pre-registration window would help establish the boundary of this fix.

Good question, FLIP-537 explicitly guarantees that the SourceCoordinator receives restored-split reports before any subsequent checkpoint, and I verified the path in Flink 2.2.1: SourceOperator submits the report during restore before the task reports RUNNING, checkpoint planning rejects non-running tasks, and registration and snapshot actions are serialized by the coordinator.

Since a connector unit test would only "model" this runtime ordering, I documented that in https://github.com/apache/flink-connector-kafka/pull/295/changes/bd222e5299310ec1b37ea5db877ae04f80762609  a proper ordering test belongs in Flink’s runtime layer.

Please tell me wdyt?

bowenli86:
thanks lgtm

bowenli86 (approved):

bowenli86:
running CI again

Savonitar:
@bowenli86  thanks for retriggering. Looks like the RC of flaky failure is same as fixed in https://github.com/apache/flink-connector-kafka/pull/285

bowenli86:
rerunning failed CI

MartijnVisser (commented):
Verified locally: the six recovery tests and both gate tests fail with the restore and snapshot hunks taken out and pass on this branch, and the V3 fixture bytes match what the current serializer writes. The CI failure is FLINK-40136, not this change.

Two things before I approve. This changes the checkpoint format, so please fill in the remaining template sections, the affects part and the AI disclosure in particular. AGENTS.md also lists the serializer versions, that line needs 4 now.

I think this should go in before #300 and #323, both need a rebase on it anyway.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/ReaderRecoveryGate.java
```diff
@@ -72,11 +75,42 @@ class ReaderRecoveryGate {
...
+    Map<Integer, List<DynamicKafkaSourceSplit>> snapshotReportedSplits() {
+        Map<Integer, List<DynamicKafkaSourceSplit>> snapshot = new HashMap<>();
+        pendingReportedSplitsByReader.forEach(
+                (readerId, splits) -> snapshot.put(readerId, new ArrayList<>(splits)));
+        return snapshot;
+    }
+
+    /**
+     * Pending entries can come from earlier registrations or remapped checkpoint state. Merging by
+     * split id, preferring the current report, avoids both losing and duplicating splits.
```
MartijnVisser: Can this see two different non-empty reports for one reader? A reader with a pending entry holds no splits, so its next report is empty or the same set. If it is only defensive, say so here.

Savonitar: There is a concrete case where the sets differ: a restored sub-enumerator can discover a new partition of an existing topic before the first asynchronous dynamic-metadata refresh completes. The reader buffers the new split B while previously reported split A still awaits reassignment, so a checkpoint can contain A in the enumerator’s pending map and B in reader state. On restore, both sets must be preserved. I verified this with a deterministic connector-level diagnostic and checked the runtime checkpoint ordering against Flink 2.2.1. Updated the comment in 4ee2c4a1 to explain this distinction: merging the sets is necessary, preferring the current report for overlapping split IDs is defensive.

MartijnVisser: You are right, I reproduced it. With the metadata refresh outstanding, the restored sub enumerator pushed a newly discovered partition to the reader whose report was still in the pending map, so the merge is needed and not just defensive.

MartijnVisser (approved):

PR #305, opened 2026-09-09
[FLINK-40615][Connectors/Kafka] Fix duplicate partition assignment in KafkaSourceEnumerator

slachiewicz:

`KafkaSourceEnumerator` can assign the same partition to two readers when the discovery interval is shorter than the time `initializePartitionSplits` takes on the worker thread ([FLINK-40615](https://issues.apache.org/jira/browse/FLINK-40615)).

`getPartitionChange()` deduplicates against `assignedSplits` and `pendingPartitionSplitAssignment` only. Neither covers partitions whose split initialization is still in flight, so a discovery pass that fires before `handlePartitionSplitChanges` runs sees those partitions as new, dispatches a second `initializePartitionSplits`, and assigns them twice.

Impact: a job whose `partition.discovery.interval.ms` is below the broker metadata response time can silently consume duplicate records.

Observed in CI: `SourceTopicIntegrityTest.testTopicIntegritySuccess[PATTERN]` read 120 records instead of 100, with partitions 7 and 8 assigned twice ([run 34200551453](https://github.com/apache/flink-connector-kafka/actions/runs/34200551453/job/101978105599)). The test is now named `SourceTopicIntegrityITCase` (FLINK-40618).

## Brief change log

- `KafkaSourceEnumerator.java`: adds a `partitionsBeingInitialized` set. `checkPartitionChanges` fills it before dispatching `callAsync`, the handler wrapper clears it on the success and failure paths, and `getPartitionChange` deduplicates against it with `removeAll`. The migration path for `isMigrated()` splits skips partitions that are already in flight.
- `KafkaSourceEnumeratorTest.java`: adds `testInFlightPartitionsNotRediscoveredAsNew` and `testMigratedSplitInFlightNotRediscovered`. Each test runs two discovery cycles without executing the pending `initializePartitionSplits` callable between them, then asserts a single dispatch. The migration test fails without the guard (`Expected size: 1 but was: 2`).
- Trims long comments and Javadoc in both files; the trimming changes no behavior.

`SourceTopicIntegrityITCase` is unchanged. Its 50 ms discovery interval is what exposed the race, so it stays as the regression guard rather than being tuned away.

## Verifying this change

The new unit tests reproduce the race directly: each drives two periodic discovery passes while the first initialization is still pending, which is the ordering the enumerator previously mishandled. Existing `KafkaSourceEnumeratorTest` cases pass unmodified.

Verified on the rebased branch: `mvn -pl flink-connector-kafka verify -Dtest='KafkaSourceEnumeratorTest,SourceTopicIntegrityITCase'` → 33 tests, 0 failures in each surefire execution, including the nine `SourceTopicIntegrityITCase` cases at the unchanged 50 ms interval. That is one local run; the original failure was intermittent, so CI is the stronger signal.

  - The serializers: no

## Documentation

---

Generated-by: Claude Code

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
@@ -133,6 +133,12 @@ public class KafkaSourceEnumerator
     private final Map<Integer, Set<KafkaPartitionSplit>> pendingPartitionSplitAssignment =
             new HashMap<>();
 
+    /**
+     * Partitions with split initialization in flight on the worker thread. Discovery skips these
+     * partitions to avoid duplicate assignments. Accessed only from the coordinator thread.
+     */
+    private final Set<TopicPartition> partitionsBeingInitialized = new HashSet<>();
+
     private final SplitOwnerSelector splitOwnerSelector;
 
     /** The consumer group id used for this KafkaSource. */
@@ -368,9 +374,19 @@ public class KafkaSourceEnumerator
         if (partitionChange.isEmpty()) {
             return;
         }
+        // Track in-flight partitions to avoid duplicate discovery.
+        final Set<TopicPartition> partitionsInThisBatch = new HashSet<>();
+        partitionsInThisBatch.addAll(partitionChange.getInitialPartitions());
+        partitionsInThisBatch.addAll(partitionChange.getNewPartitions());
+        partitionsBeingInitialized.addAll(partitionsInThisBatch);
+
         context.callAsync(
                 () -> initializePartitionSplits(partitionChange),
-                this::handlePartitionSplitChanges);
+                (result, error) -> {
+                    // Clear in-flight state on the success and failure paths.
+                    partitionsBeingInitialized.removeAll(partitionsInThisBatch);
+                    handlePartitionSplitChanges(result, error);
+                });
     }
 
     /**
@@ -565,6 +581,8 @@ public class KafkaSourceEnumerator
                 (reader, splits) ->
                         splits.forEach(
                                 split -> dedupOrMarkAsRemoved.accept(split.getTopicPartition())));
+        // Skip partitions with initialization in flight to avoid duplicate assignments.
+        newPartitions.removeAll(partitionsBeingInitialized);
 
         if (!newPartitions.isEmpty()) {
             LOG.info("Discovered new partitions: {}", newPartitions);
@@ -580,7 +598,8 @@ public class KafkaSourceEnumerator
         }
         // migration path, ensure that partitions without offset are properly initialized
         for (KafkaPartitionSplit split : unassignedSplits.values()) {
-            if (split.isMigrated()) {
+            if (split.isMigrated()
+                    && !partitionsBeingInitialized.contains(split.getTopicPartition())) {
                 initialPartitions.add(split.getTopicPartition());
             }
         }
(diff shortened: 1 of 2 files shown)


slachiewicz:
Should fix current master branch issue

spuru9:
There are some spotless error. Can you fix.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
```diff
@@ -565,6 +588,14 @@ PartitionChange getPartitionChange(
...
                         splits.forEach(
                                 split -> dedupOrMarkAsRemoved.accept(split.getTopicPartition())));
+        // Exclude partitions whose split initialization is currently in-flight on the
+        // worker thread. Without this, a discovery cycle that fires before
+        // handlePartitionSplitChanges runs would re-dispatch initializePartitionSplits
+        // for the same partitions, leading to duplicate split assignments.
+        partitionsBeingInitialized.forEach(
+                (tp) -> {
+                    newPartitions.remove(tp);
+                });
```
spuru9: **nit:** Can just be `removeAll`, no need to loop.

```suggestion
        newPartitions.removeAll(partitionsBeingInitialized);
```

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
spuru9, line 605: **major:** Same race, unguarded here. A migrated split stays `isMigrated()` in `unassignedSplits` until `handlePartitionSplitChanges` overwrites it, so if discovery fires again while its `initializePartitionSplits` is in flight, this loop re-adds it to `initialPartitions` and dispatches a second time. I reproduced the double-dispatch in a test; the guard below fixes it.

```        
        // migration path, ensure that partitions without offset are properly initialized
        for (KafkaPartitionSplit split : unassignedSplits.values()) {
            if (split.isMigrated()
                    && !partitionsBeingInitialized.contains(split.getTopicPartition())) {
                initialPartitions.add(split.getTopicPartition());
            }
        }
```

spuru9 (commented):
Optional: I personally don't prefer long inline comments especially in the test suite. If there is scope to cut down would be more readable.

spuru9 (approved):
LGTM

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
slachiewicz: Taken; the guard is at line 601–602.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
```diff
@@ -368,9 +374,19 @@ private void checkPartitionChanges(Set<TopicPartition> fetchedPartitions, Throwa
...
+        partitionsInThisBatch.addAll(partitionChange.getInitialPartitions());
+        partitionsInThisBatch.addAll(partitionChange.getNewPartitions());
+        partitionsBeingInitialized.addAll(partitionsInThisBatch);
+
         context.callAsync(
                 () -> initializePartitionSplits(partitionChange),
-                this::handlePartitionSplitChanges);
+                (result, error) -> {
+                    // Clear in-flight state on the success and failure paths.
+                    partitionsBeingInitialized.removeAll(partitionsInThisBatch);
```
Savonitar: Should we also clear the in-flight state when `StoppableKafkaEnumContextProxy` suppresses an initialization failure?
Wouldnt this leave the partition unassigned until the enumerator is recreated? 
Could we cover this recovery sequence with a regression test and ensure the suppressed-failure path releases the partition?

PR #308, opened 2026-09-09
[FLINK-40620][source/tests] Split SourceTopicIntegrityTest by guarantee

dalelane:
SourceTopicIntegrityTest.testTopicIntegritySuccess has been intermittently failing in CI workflows, reporting that it has got more records than it was expecting.

The test now runs each subscription mode twice — once with the sink set to at-least-once delivery and once to exactly-once where previously it only ever ran at-least-once.

In the at-least-once runs it checks that no records were lost rather than insisting on an exact count, so a harmless retry that writes a record twice no longer fails the test.

In the exactly-once runs it checks for an exact count because that mode promises no duplicates.


MartijnVisser:
@dalelane I think this duplicates https://github.com/apache/flink-connector-kafka/pull/305

dalelane:
@MartijnVisser Yeah, I think you're right. Thanks for the heads-up - apologies for the noise!

PR #300, opened 2026-09-07
[FLINK-39712][Connectors/Kafka] Avoid duplicate bounded-source completion notifications

Savonitar:
## What is the purpose of the change

Fixes FLINK-39712: a bounded `DynamicKafkaSource` could send `NoMoreSplitsEvent` more than once to a reader. A later registration or returned-split callback could signal a task that had already finished, causing the runtime to fail the job.

The observed failure includes:

```text
An OperatorEvent from an OperatorCoordinator to a task was lost
TaskNotRunningException: Task is not running, but in state FINISHED
```

Signal each reader once per registration, after all split assignments for the operation are complete. Deduplication and deferral are needed together: a cluster callback must not consume the reader's completion signal before the remaining clusters or retained recovery splits have been assigned.

## Brief change log

- Track readers that have received completion and reset a reader's entry when it registers again.
- Defer completion during pending recovery and across returned-split assignment, reader registration, and reported-split recovery.
- Centralize the synchronous assignment guard in a private `Runnable` helper, resetting it in `finally`.
- Recheck completion after each successful assignment operation, preserving the ordering before deferred metadata-event delivery.
- Wire the completion callback into the enumerator test context and assert the assigned split IDs at each completion signal.

## Verifying this change

Five new tests in `DynamicKafkaSourceEnumeratorTest` cover signal counts and assignment-before-completion ordering:

- `testBoundedSourceSignalsNoMoreSplitsOncePerReader`: initial completion is sent once per reader.
- `testBoundedSourceAssignsAllClustersBeforeCompletingLateReaders`: all clusters assign their splits before a late reader completes.
- `testBoundedSourceCompletesRestartedReaderAfterReturnedSplits`: a restarted reader receives its returned splits and a fresh completion signal without re-signalling other readers.
- `testBoundedSourceReassignsReportedSplitsBeforeCompletingReader`: reported splits are reassigned before completion.
- `testBoundedSourceCompletesReadersAfterActiveAndRetainedReassignment`: active and retained splits are assigned before recovered readers complete.

The branch is rebased onto Apache `main` at `9c0ccf3b1b52ccd98ccd9cd714330a522b19850b`, preserving the pending-recovery restoration from #295 and the upstream enumerator-test concurrency fix.

For current head `0f703b136bb28540ba529eef607f99bf565fa262`:

- Static inspection confirmed assignment/completion ordering and preservation of all five regression tests.
- `git diff --check` passed, and GitHub reports no merge conflicts.
[...]

PR #309, opened 2026-09-10
[FLINK-40586][Connectors/Kafka] Signal no more splits to each reader once per registration

vbhanuchander-lang:

Fixes [FLINK-40586](https://issues.apache.org/jira/browse/FLINK-40586).

`DynamicKafkaSourceEnumerator.handleNoMoreSplits()` signalled **every** currently registered reader on **every** invocation, with no record of which readers had already been signalled:

```java
enumContext.registeredReaders().keySet().forEach(enumContext::signalNoMoreSplits);
```

It is reached from `addReader`, `addSplitsBack` and `tryCompletePendingReaderRegistration`, so a reader that finished on an earlier signal is signalled again when another reader registers — or when any reader is reset, since `SourceCoordinator.subtaskReset` calls `addSplitsBack` for every reset subtask, passing an empty list for a reader that owned no splits. `NoMoreSplitsEvent` is not loss tolerant, so delivering it to a task already in `FINISHED` fails the job:

```
org.apache.flink.util.FlinkException: An OperatorEvent from an OperatorCoordinator to a task was
lost. Triggering task failover to ensure consistency. Event: '[NoMoreSplitEvent]'
Caused by: TaskNotRunningException: Task is not running, but in state FINISHED
```

That is the `DynamicKafkaSourceITTest$IntegrationTests.testIdleReader` failure seen in weekly CI, where parallelism is one greater than the split count so the idle reader finishes on the first signal.

## Brief change log

- `handleNoMoreSplits()` signals a registered reader only if `readersSignalledNoMoreSplits.add(readerId)` succeeds, so each reader is signalled at most once per registration.
- `addReader()` clears the reader's entry. **Registration is the point at which a new attempt of a reader becomes addressable**, so it is the event that has to re-arm the signal; without it a restarted reader would never be told and the job would not finish.

A metadata change recreates the sub-enumerators in `onHandleSubscribedStreamsFetch` and their new assignments would need a fresh signal too, but that path does not signal at all today (FLINK-31006), so it is left alone here and noted in a comment.

## Verifying this change

Added `DynamicKafkaSourceEnumeratorTest#testNoMoreSplitsIsSignalledOncePerReaderRegistration`. It registers three readers and asserts each is signalled exactly once; then unregisters one reader and hands its splits back — the order the coordinator actually produces, since `executionAttemptFailed` unregisters before `subtaskReset` — and asserts nobody is re-signalled; then re-registers that reader and asserts its count reaches 2 while the others stay at 1.

`MockSplitEnumeratorContext` only latches a `boolean[] subtaskHasNoMoreSplits` and exposes `hasNoMoreSplits(int)`, so a repeated signal is invisible to it. The test uses a `CountingSignalNoMoreSplitsContext` subclass that counts calls per subtask.

The test fails on `main` and passes here. Locally: `Tests run: 31, Failures: 0` for the whole class, `spotless:check` and `checkstyle:check` clean.

**Ordering:** this should land after #291 (FLINK-40362). Per @MartijnVisser's measurements of `testIdleReader`, `main` hung 2 of 24 runs, `main` + this PR 1 of 24, and `main` + #291 + this PR 0 of 20 — the reader-side hang is a separate cause that this change does not address.

- Dependencies (does it add or upgrade a dependency): **no**
- The public API: **no**
- The serializers: **no**
- The runtime per-record code paths: **no**
- Anything that affects deployment or recovery: **yes** — it changes when `NoMoreSplitsEvent` is delivered to readers on registration and failover, for bounded sources only.
- The S3 file system connector: **no**

## Documentation

- Does this pull request introduce a new feature? **no**
- If yes, how is the feature documented? **not applicable**

Generated-by: Claude Code (Opus 5)

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
@@ -115,6 +115,7 @@ public class DynamicKafkaSourceEnumerator
             retainedClusterEnumeratorStates;
     private boolean firstDiscoveryComplete;
     private final ReaderRecoveryGate readerRecoveryGate;
+    private final Set<Integer> readersSignalledNoMoreSplits;
 
     public DynamicKafkaSourceEnumerator(
             KafkaStreamSubscriber kafkaStreamSubscriber,
@@ -218,6 +219,7 @@ public class DynamicKafkaSourceEnumerator
         this.splitAssignmentStrategy = createSplitAssignmentStrategy(properties);
         this.readerRecoveryGate =
                 new ReaderRecoveryGate(hasRestoredEnumeratorState(dynamicKafkaSourceEnumState));
+        this.readersSignalledNoMoreSplits = new HashSet<>();
 
         if (!dynamicKafkaSourceEnumState.getClusterEnumeratorStates().isEmpty()) {
             logger.info("Dynamic Kafka source restored from checkpointed enumerator state");
@@ -390,10 +392,17 @@ public class DynamicKafkaSourceEnumerator
             }
 
             if (firstDiscoveryComplete && allEnumeratorsHaveSignalledNoMoreSplits) {
-                logger.info(
-                        "Signal no more splits to all readers: {}",
-                        enumContext.registeredReaders().keySet());
-                enumContext.registeredReaders().keySet().forEach(enumContext::signalNoMoreSplits);
+                // a reader that has already been signalled may have finished, and
+                // NoMoreSplitsEvent is not loss tolerant: re-sending it to a FINISHED task fails
+                // the job. Signal each reader once per generation of this set, which is cleared
+                // when the reader registers again -- the point at which a new attempt becomes
+                // addressable -- and when a metadata change recreates the sub enumerators.
+                for (Integer readerId : enumContext.registeredReaders().keySet()) {
+                    if (readersSignalledNoMoreSplits.add(readerId)) {
+                        logger.info("Signal no more splits to reader: {}", readerId);
+                        enumContext.signalNoMoreSplits(readerId);
+                    }
+                }
             } else {
                 logger.info("Not ready to notify no more splits to readers.");
             }
@@ -478,6 +487,11 @@ public class DynamicKafkaSourceEnumerator
         logger.info("Closing enumerators due to metadata change");
 
         closeAllEnumeratorsAndContexts();
+        // every sub enumerator is about to be recreated and will signal no more splits again for
+        // its new assignments, and the reader has closed and recreated its sub readers, so the
+        // signals sent for the previous generation no longer hold. Without clearing this, a bounded
+        // job would hang after a metadata change once FLINK-31006 restores that re-signal.
+        readersSignalledNoMoreSplits.clear();
         retainRemovedClusterEnumeratorStates(
                 dynamicKafkaSourceEnumState.getClusterEnumeratorStates(),
                 latestClusterTopicsMap.keySet());
@@ -763,6 +777,10 @@ public class DynamicKafkaSourceEnumerator
     @Override
     public void addReader(int subtaskId) {
         logger.debug("Adding reader {}", subtaskId);
+        // registration is the point at which a new attempt of this reader becomes addressable, so
+        // a signal sent to a previous attempt does not count: it must be signalled again or a
+        // restarted reader would never finish.
+        readersSignalledNoMoreSplits.remove(subtaskId);
         ReaderInfo readerInfo = enumContext.registeredReaders().get(subtaskId);
         if (readerInfo != null) {
             readerRecoveryGate.recordReportedSplits(
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
@@ -2024,9 +2024,31 @@ public class DynamicKafkaSourceEnumeratorTest {
     }
 
     private DynamicKafkaSourceEnumerator createEnumerator(
+            SplitEnumeratorContext<DynamicKafkaSourceSplit> context, Boundedness boundedness) {
+        return createEnumerator(
+                context,
+                new MockKafkaMetadataService(
+                        Collections.singleton(DynamicKafkaSourceTestHelper.getKafkaStream(TOPIC))),
+                (properties) -> {},
+                boundedness);
+    }
+
+    private DynamicKafkaSourceEnumerator createEnumerator(
             SplitEnumeratorContext<DynamicKafkaSourceSplit> context,
             KafkaMetadataService kafkaMetadataService,
             Consumer<Properties> applyPropertiesConsumer) {
+        return createEnumerator(
+                context,
+                kafkaMetadataService,
+                applyPropertiesConsumer,
+                Boundedness.CONTINUOUS_UNBOUNDED);
+    }
+
+    private DynamicKafkaSourceEnumerator createEnumerator(
+            SplitEnumeratorContext<DynamicKafkaSourceSplit> context,
+            KafkaMetadataService kafkaMetadataService,
+            Consumer<Properties> applyPropertiesConsumer,
+            Boundedness boundedness) {
         Properties properties = new Properties();
         applyPropertiesConsumer.accept(properties);
         properties.putIfAbsent(KafkaSourceOptions.PARTITION_DISCOVERY_INTERVAL_MS.key(), "0");
@@ -2037,9 +2059,11 @@ public class DynamicKafkaSourceEnumeratorTest {
                 kafkaMetadataService,
                 context,
                 OffsetsInitializer.earliest(),
-                new NoStoppingOffsetsInitializer(),
+                Boundedness.BOUNDED.equals(boundedness)
+                        ? OffsetsInitializer.latest()
+                        : new NoStoppingOffsetsInitializer(),
                 properties,
-                Boundedness.CONTINUOUS_UNBOUNDED,
+                boundedness,
                 new DynamicKafkaSourceEnumState(),
                 new TestKafkaEnumContextProxyFactory());
     }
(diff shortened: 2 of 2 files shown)


vbhanuchander-lang:
@MartijnVisser tagging you since you reported FLINK-40586 — this is the enumerator-side fix for the `testIdleReader` failover you saw in the weekly CI.

One thing worth flagging: `Flink Connector Kafka CI` is sitting at `action_required`, since this is my first PR in this repo, so it needs a maintainer to approve the workflow run before anything actually compiles or tests here. Only `Title Validator` has run so far. I verified locally against Flink 2.2.1 on **JDK 21 only** — `DynamicKafkaSourceEnumeratorTest` is 31/31 with the change, and the new test fails on unfixed `main` with the first reader signalled 3 times rather than once — so the JDK 11 and 17 legs are still unverified.

If you or another Kafka committer could also assign the ticket to me, happy to keep it tidy. No rush on either; mainly noting the CI gate so the missing checks don't read as checks that passed.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -767,6 +778,9 @@ private void addSplitsBackToClusterEnumerators(
     @Override
     public void addReader(int subtaskId) {
         logger.debug("Adding reader {}", subtaskId);
+        // this reader is (re-)registering, so a no more splits signal sent to a previous attempt
+        // does not count: it must be signalled again or a restarted reader would never finish.
+        readersSignalledNoMoreSplits.remove(subtaskId);
```
MartijnVisser, line 783: Resetting here is right, but not for the reason given in the description and the commit message. `SourceCoordinator.subtaskReset` calls `addSplitsBack` for every reset subtask, with an empty list for a reader that owned nothing; on `main` that empty call re-signals everyone (checked with a test, reader 0 ends at 4). Please reword both to what holds: the new attempt's registration is the event that makes the reader addressable again.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -394,10 +396,19 @@ private void handleNoMoreSplits() {
...
-                        "Signal no more splits to all readers: {}",
-                        enumContext.registeredReaders().keySet());
-                enumContext.registeredReaders().keySet().forEach(enumContext::signalNoMoreSplits);
+                // a reader that has already been signalled may have finished, and
+                // NoMoreSplitsEvent is not loss tolerant: re-sending it to a FINISHED task fails
+                // the job. Signal each registered reader once, until it registers again.
+                Set<Integer> readersToSignal =
+                        new HashSet<>(enumContext.registeredReaders().keySet());
+                readersToSignal.removeAll(readersSignalledNoMoreSplits);
+                if (readersToSignal.isEmpty()) {
```
MartijnVisser: Nit: `if (readersSignalledNoMoreSplits.add(readerId))` per reader does the same without the temporary set and the early `return`.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -394,10 +396,19 @@ private void handleNoMoreSplits() {
             }
 
             if (firstDiscoveryComplete && allEnumeratorsHaveSignalledNoMoreSplits) {
-                logger.info(
-                        "Signal no more splits to all readers: {}",
-                        enumContext.registeredReaders().keySet());
-                enumContext.registeredReaders().keySet().forEach(enumContext::signalNoMoreSplits);
+                // a reader that has already been signalled may have finished, and
+                // NoMoreSplitsEvent is not loss tolerant: re-sending it to a FINISHED task fails
+                // the job. Signal each registered reader once, until it registers again.
```
MartijnVisser: Nit: "until it registers again" is one of two resets. A metadata change recreates all sub-enumerators in `onHandleSubscribedStreamsFetch`, and their new assignments need a fresh signal; today that path never signals anyway (FLINK-31006), so clearing the set there changes nothing observable, but it keeps this set from being the thing that blocks the re-signal once 31006 is fixed. Fine as a follow-up if you prefer.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
```diff
@@ -2455,6 +2455,93 @@ private ClusterMetadata copyClusterMetadataWithOverrides(
...
+            for (SplitsAssignment<DynamicKafkaSourceSplit> assignment :
+                    context.getSplitsAssignmentSequence()) {
+                List<DynamicKafkaSourceSplit> splits = assignment.assignment().get(failingReader);
+                if (splits != null) {
+                    splitsOfFailingReader.addAll(splits);
+                }
+            }
+            assertThat(splitsOfFailingReader).as("precondition: reader had splits").isNotEmpty();
+
+            enumerator.addSplitsBack(splitsOfFailingReader, failingReader);
```
MartijnVisser, line 2521: The second half of the contract, that a re-registered reader is signalled again, is not asserted, and the reset reader is still registered here, which is not what the coordinator produces (`executionAttemptFailed` unregisters before `subtaskReset`). Please unregister it first, then re-register after `addSplitsBack` and assert its count goes to 2 while the others stay at 1. Passes on this branch, fails on `main`.

```java
            context.unregisterReader(failingReader);
            enumerator.addSplitsBack(splitsOfFailingReader, failingReader);
            for (int reader = 0; reader < NUM_SUBTASKS; reader++) {
                assertThat(context.getSignalCount(reader)).isEqualTo(1);
            }

            mockRegisterReaderAndSendReaderStartupEvent(context, enumerator, failingReader);
            runAllOneTimeCallables(context);
            assertThat(context.getSignalCount(failingReader)).isEqualTo(2);
            for (int reader = 0; reader < failingReader; reader++) {
                assertThat(context.getSignalCount(reader)).isEqualTo(1);
            }
```

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
```diff
@@ -2455,6 +2455,93 @@ private ClusterMetadata copyClusterMetadataWithOverrides(
...
+
+            for (int reader = 0; reader < failingReader; reader++) {
+                assertThat(context.getSignalCount(reader))
+                        .as("reader %s must not be signalled again after it was already told", reader)
```
MartijnVisser: `spotless:check` fails on this line, CI will stop here.

```suggestion
                        .as(
                                "reader %s must not be signalled again after it was already told",
                                reader)
```

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
```diff
@@ -2455,6 +2455,93 @@ private ClusterMetadata copyClusterMetadataWithOverrides(
...
+                        .isEqualTo(1);
+            }
+        }
+    }
+
+    private DynamicKafkaSourceEnumerator createBoundedEnumerator(
```
MartijnVisser: Nit: this copies `createEnumerator` with two arguments changed; a `Boundedness` parameter on the existing helper is enough.

MartijnVisser (changes requested):
Thanks, the fix is right: the new test fails on `main` and passes here, also on JDK 11/17/21 and on a rebase over #289. Three things before it can go in. Please rebase over `main`, #289 moved the pending-reader fields into `ReaderRecoveryGate` and the conflict is only the field and its initialiser. `spotless:check` fails on this branch (inline). And this lands after FLINK-40362's #291: on `main` an idle reader that gets the signal before its `MetadataUpdateEvent` only finishes through a later duplicate signal, which this removes. `testIdleReader` here: `main` hung 2 of 24 runs (idle reader, three signals all before its metadata, no end of input), `main` + this PR 1 of 24, `main` + #291 + this PR 0 of 20. One more: this repo has no PR template, but please follow the apache/flink one for the AI disclosure, a `Generated-by: <tool and version>` line in the description and the same trailer on the commit, or state that no tooling was used (https://www.apache.org/legal/generative-tooling.html). Smaller points inline.

vbhanuchander-lang:
Thanks @MartijnVisser — that is a thorough review, and the correction on the rationale is the important one. All six points addressed, force-pushed on top of `main` (`91e01a5`, so #289's `ReaderRecoveryGate` is in).

**Rationale reworded in both the description and the commit message.** You are right and I was wrong: I had claimed `addSplitsBack` is never called for a reader that owned nothing, so resetting there would strand an idle reader. `SourceCoordinator.subtaskReset` calls it for *every* reset subtask, with an empty list in that case, which is exactly why the empty call re-signals everyone on `main`. Both now say what actually holds — registration is the event that makes a new attempt addressable, so it is what has to re-arm the signal.

**`handleNoMoreSplits` simplified** to `if (readersSignalledNoMoreSplits.add(readerId))` per reader; the temporary set and the early `return` are gone.

**Comment on the metadata path** now records that `onHandleSubscribedStreamsFetch` recreates the sub-enumerators and their assignments would need a fresh signal, that FLINK-31006 means that path does not signal at all today, and that clearing the set there is therefore not observable yet. Left as a follow-up as you suggested, so this set does not become the thing that blocks it once 31006 is fixed.

**Test strengthened** using your ordering — `unregisterReader` before `addSplitsBack`, assert nobody moves off 1, then re-register and assert the reset reader reaches 2 while the others stay at 1. That asserts the half of the contract I had left uncovered.

**`spotless`** fixed. That was my mistake in a way worth owning: I ran `spotless:check` before opening the PR and read the exit code of a `tail` in the pipeline rather than Maven's, so I reported it clean when it was not, and every CI leg died on it. Now verified properly — `spotless:check` and `checkstyle:check` both exit 0, and the class is 31/31 locally.

**Duplicate helper removed** — `createEnumerator` takes a `Boundedness` parameter and picks the matching stopping offsets initializer; the near-copy is gone.

**Ordering** noted in the description: this should go in after #291. Your numbers make the point clearly — 1 of 24 with this change alone versus 0 of 20 with both — so this closes the re-signal cause and #291 closes the reader-side hang. Happy to wait for #291 and rebase, or to have them go in together, whichever you prefer.

On the generative tooling disclosure: added `Generated-by: Claude Code (Opus 5)` to both the description and the commit trailer, per https://www.apache.org/legal/generative-tooling.html. Thanks for pointing at the apache/flink template — I had not realised this repo inherits it.

MartijnVisser:
@vbhanuchander-lang CI is failing, please take a look

MartijnVisser:
Correcting my own nit on `readersSignalledNoMoreSplits`: I said clearing the set when sub-enumerators are recreated was fine as a follow-up. It is not, once FLINK-31006 lands.

The chain is #291 plus this set. #291 deliberately replays no-more-splits only on the reader's first metadata update and leaves a later metadata change to the enumerator signalling again, which is the re-signal FLINK-31006 restores. On that metadata change `onHandleSubscribedStreamsFetch` recreates every sub-enumerator, each one signals, but the readers never re-registered, so the set still holds them, `readersToSignal` is empty and `handleNoMoreSplits` returns without signalling. `DynamicKafkaSourceReader` has meanwhile closed and recreated every sub-reader, each back to `noMoreSplitsAssignment == false`. A bounded job hangs after a metadata change.

The set is keyed on the reader alone; it needs the enumerator generation as well, or clearing wherever the sub-enumerators are recreated. Please take it in this PR rather than as a follow-up.

vbhanuchander-lang:
@MartijnVisser you were right to retract that nit — taken in this PR, and thanks for working the chain through. Pushed.

**Metadata change now clears the set**, in `onHandleSubscribedStreamsFetch` immediately after `closeAllEnumeratorsAndContexts()`, which is the point where the previous generation stops existing: every sub enumerator is about to be recreated and will signal for its new assignments, and the reader has closed and recreated its sub readers back to `noMoreSplitsAssignment == false`. I went with clearing at the recreation point rather than keying the set on a generation counter, since the clear sits exactly where the generation changes and needs no extra state; happy to switch to an explicit generation key if you would rather have it be impossible to forget at a future recreation site. The commit message now describes both clears as the two points that define a generation.

**I could not write a test that proves it, and would rather say so than fake it.** I wrote one — change metadata mid-run, then drive any path back into `handleNoMoreSplits` and assert the readers are signalled a second time. It fails at `expected: 2, but was: 1`, because after the recreation the new sub enumerators have not signalled, so `allEnumeratorsHaveSignalledNoMoreSplits` is false and `handleNoMoreSplits` takes the "not ready" branch — the FLINK-31006 gap you described, from the other side. The only ways I found to make it green were asserting the private set through `Whitebox` or contriving the precondition, both of which test the harness rather than the behaviour, so I dropped the test. If you want coverage now I am happy to add whichever form you prefer, or to add the real one to the 31006 change so it lands with the behaviour it asserts.

**On the CI failure — that was my mistake twice over and worth naming, since it explains why a "fixed" formatting failure came back.** `flink-connector-parent` skips spotless on Java 21 ("Current google format does not run on Java 21"), and I was running JDK 21, so both `spotless:apply` and `spotless:check` were no-ops and reported success while changing nothing. Re-run under JDK 17 it reformatted four wrappings immediately. I now check that the goal actually executed rather than trusting the exit code: `spotless:check` and `checkstyle:check` exit 0 under JDK 17 with no "Spotless check skipped" lines, and the class is 31/31 locally.

Still happy to sequence this behind #291 — say the word if you would rather I rebase on it once it lands.

vbhanuchander-lang:
@MartijnVisser #291 landed yesterday, so I have rebased this onto current `main` (`6b99bb3`, which includes it) — the sequencing you asked for is now satisfied and this is no longer waiting on anything. Clean rebase, no conflicts.

Re-verified on the rebased tree:

- `DynamicKafkaSourceEnumeratorTest` 31/31 and `DynamicKafkaSourceReaderTest` 15/15, 46 tests, 0 failures. I ran the reader class too since #291 changed that side and this change interacts with it.
- `spotless:check` and `checkstyle:check` clean **on JDK 17**, confirmed by checking the goal actually executed rather than the exit code — `flink-connector-parent` sets `<skip>true</skip>` for spotless on Java 21, which is what made my earlier runs report success while doing nothing. That was the cause of both red CI rounds.

Two things still open from your side rather than mine:

1. **The workflow run is gated.** Every run on this PR has ended at `action_required` because it is my first PR in this repo, so the JDK 11/17/21 legs have never actually executed here — only `Title Validator`. If you or another committer can approve the run, that is the last unverified piece.
2. **Coverage for the metadata-change clear.** Still no test, for the reason in my previous comment: the recreated sub enumerators do not reach the signalling precondition, so the behaviour is unreachable until FLINK-31006. Happy to add a `Whitebox` assertion on the set if you want something now, or to put the real test on the 31006 change so it lands with the behaviour it asserts — your call, and I did not want to pick for you.

MartijnVisser:
@vbhanuchander-lang #300 carries the same dedup and has been open since 7 September under FLINK-39712. I filed FLINK-40586 an hour later and assigned it to you without spotting the overlap. That is my mistake.

The dedup can't land on its own. `handleNoMoreSplits` is reached from the proxy callback part way through `addReaderToClusterEnumerators`, so the first signal goes out before the other clusters have assigned. On `main` a later one corrects it; once the set makes it the only signal, the reader finishes holding one cluster's splits. Four of #300's five tests fail on this branch for exactly that.

So I want #300 in first, with its deferral, and this PR reduced to the clear in `onHandleSubscribedStreamsFetch` that #300 doesn't have. Please rebase once it lands and drop the set, its initialiser and the reset in `addReader`. #300 also wires `signalNoMoreSplitsCallback` in `TestKafkaEnumContextProxyFactory`, so the test you couldn't write works on that base. I'll re-scope FLINK-40586 and it stays assigned to you.

PR #310, opened 2026-09-10
[FLINK-40621][Connectors/Kafka] Flush before fencing in producer reset test

slachiewicz:

`FlinkKafkaInternalProducerITCase.testResetInnerTransactionIfFinalizingTransactionFailed` is flaky ([FLINK-40621](https://issues.apache.org/jira/browse/FLINK-40621)). It sends a record without flushing, then fences that producer with a second one using the same transactional id, and asserts that the finalizer throws `ProducerFencedException`.

Which exception arrives depends on where the first producer's `ProduceRequest` lands relative to the epoch bump:

- The record is accepted first, so the fencing surfaces on the `EndTxn` path as `PRODUCER_FENCED`, and the assertion holds.
- The `ProduceRequest` arrives after the bump, so the broker rejects it with `INVALID_PRODUCER_EPOCH` and the finalizer throws `InvalidProducerEpochException`.

In kafka-clients 4.2.0 both are siblings under `ApplicationRecoverableException`, neither a subtype of the other, so the second ordering fails the assertion. The reported failure carries the produce-path message, `attempted to produce with an old epoch`.

## Brief change log

- `FlinkKafkaInternalProducerITCase`: flush the fenced producer before the second producer starts, so the failure is always discovered on the `EndTxn` path. A comment records why the flush must stay.

No production code and no assertion change: widening the assertion to the shared supertype would also be green, but would stop pinning which path failed.

## Verifying this change

Verified on the rebased branch: `mvn -pl flink-connector-kafka verify -Dtest='FlinkKafkaInternalProducerITCase#testResetInnerTransactionIfFinalizingTransactionFailed'` → 2 tests, 0 failures, both parameterizations.

That is one local run against a passing ordering, so it does not by itself prove the race is gone; the argument for that is the flush removing the interleaving rather than the run.

  - The serializers: no

## Documentation

---

Generated-by: Claude Code (Claude Opus 5)

### DIFF
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducerITCase.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/FlinkKafkaInternalProducerITCase.java
@@ -159,6 +159,12 @@ class FlinkKafkaInternalProducerITCase {
             fenced.initTransactions();
             fenced.beginTransaction();
             fenced.send(new ProducerRecord<>(topic, "test-value"));
+            // Await the produce before fencing. An in-flight ProduceRequest that reaches the
+            // broker after the epoch bump below is rejected with INVALID_PRODUCER_EPOCH, and the
+            // finalizer then throws InvalidProducerEpochException, a sibling of
+            // ProducerFencedException rather than a subtype. Flushing keeps the fencing on the
+            // EndTxn path, which is the case this test pins.
+            fenced.flush();
             // Start a second producer that fences the first one
             try (FlinkKafkaInternalProducer<String, String> producer =
                     new FlinkKafkaInternalProducer<>(getProperties(), "dummy")) {


slachiewicz:
The red here is [FLINK-40622](https://issues.apache.org/jira/browse/FLINK-40622), not this change: [`SourceTopicIntegrityTest.testTopicIntegrityFailure[1]`](https://github.com/apache/flink-connector-kafka/actions/runs/34455212027/job/102799893439) times out on JDK 17 with the same message as on `main`. JDK 11 and 21 were cancelled by matrix fail-fast, so it is one failure rather than three.

Worth flagging for review: `FlinkKafkaInternalProducerITCase` did not run in this build at all. `SourceTopicIntegrityTest` matches `**/*Test.*` and so fails in the `default-test` execution, which ends the module before the `integration-test` phase this ITCase runs in. CI has therefore not exercised the change in either direction.

Local run against a broker: `FlinkKafkaInternalProducerITCase#testResetInnerTransactionIfFinalizingTransactionFailed` gives 2 tests, 0 failures, both parameterizations.

PR #313, opened 2026-09-10
[FLINK-40585][Connectors/Kafka] Wait for the last stable offset before draining committed records

slachiewicz:
Under `read_committed`, `endOffsets` is the last stable offset, which trails the high watermark until the broker has written the markers of a transaction it already acknowledged. The helper snapshotted it once, right after the job finished, and cut off the final transaction in `rescaleListing` (run 31916828021, `[2->5]`). It now waits up to 10 s for the LSO to meet the `read_uncommitted` end offsets and logs the pinned offsets if it does not, so a truncated drain names its cause. Call sites that drain with transactions legitimately open (`testAbortTransactionsAfterScaleInBeforeFirstCheckpoint`, `ExactlyOnceKafkaWriterITCase`) keep their result and pay the bounded wait.

The other failure mode in FLINK-40585 (run 30181683657, `[3->5]`) is a product bug, tracked separately as FLINK-40626.

Verified: `KafkaSinkITCase#rescaleListing` → 10/10 locally.

### DIFF
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/testutils/KafkaUtil.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/testutils/KafkaUtil.java
@@ -55,6 +55,8 @@ public class KafkaUtil {
 
     private static final Logger LOG = LoggerFactory.getLogger(KafkaUtil.class);
     private static final Duration CONSUMER_POLL_DURATION = Duration.ofSeconds(1);
+    private static final Duration OPEN_TRANSACTION_SETTLE_TIMEOUT = Duration.ofSeconds(10);
+    private static final Duration OPEN_TRANSACTION_POLL_INTERVAL = Duration.ofMillis(100);
     private static final int REQUEST_TIMEOUT_SECONDS = 30;
 
     private KafkaUtil() {}
@@ -157,7 +159,8 @@ public class KafkaUtil {
         consumerConfig.put("value.deserializer", ByteArrayDeserializer.class.getName());
         try (KafkaConsumer<byte[], byte[]> consumer = new KafkaConsumer<>(consumerConfig)) {
             Set<TopicPartition> topicPartitions = getAllPartitions(consumer, topic);
-            Map<TopicPartition, Long> endOffsets = consumer.endOffsets(topicPartitions);
+            Map<TopicPartition, Long> endOffsets =
+                    getSettledEndOffsets(topic, consumer, consumerConfig, topicPartitions);
             consumer.assign(topicPartitions);
             consumer.seekToBeginning(topicPartitions);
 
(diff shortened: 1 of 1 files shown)


flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/testutils/KafkaUtil.java
```diff
@@ -55,6 +55,8 @@ public class KafkaUtil {
 
     private static final Logger LOG = LoggerFactory.getLogger(KafkaUtil.class);
     private static final Duration CONSUMER_POLL_DURATION = Duration.ofSeconds(1);
+    private static final Duration OPEN_TRANSACTION_SETTLE_TIMEOUT = Duration.ofSeconds(10);
```
MartijnVisser, line 58: The drains that assert while transactions are open pay this in full: `testAbortOnClose` 6.7s -> 16.6s, `testAbortTransactionsAfterScaleInBeforeFirstCheckpoint` 56.2s -> 96.2s, so +52.7s per cell across the two classes, times three JDK cells on PR CI and seven on the weekly. Please add a variant of `drainAllRecordsFromTopic` that skips the wait when transactions are expected to be open.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/testutils/KafkaUtil.java
```diff
@@ -192,6 +195,55 @@ public static List<ConsumerRecord<byte[], byte[]>> drainAllRecordsFromTopic(
...
+     * logs which offsets are pinned, so a truncated drain names its cause.
+     */
+    private static Map<TopicPartition, Long> getSettledEndOffsets(
+            String topic,
+            KafkaConsumer<byte[], byte[]> consumer,
+            Properties consumerConfig,
+            Set<TopicPartition> topicPartitions) {
+        Map<TopicPartition, Long> endOffsets = consumer.endOffsets(topicPartitions);
+        if (!"read_committed"
+                .equals(consumerConfig.getProperty(ConsumerConfig.ISOLATION_LEVEL_CONFIG))) {
```
MartijnVisser, line 213: Nit: `getProperty` returns `null` for a non-`String` value, and callers here already mix the two (`enable.auto.commit` is stored as a `Boolean`). Works today because only the three-arg overload writes `isolation.level`; `String.valueOf(consumerConfig.get(...))` makes it safe.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/testutils/KafkaUtil.java
```diff
@@ -192,6 +195,55 @@ public static List<ConsumerRecord<byte[], byte[]>> drainAllRecordsFromTopic(
...
+            return endOffsets;
+        }
+        final Properties uncommittedConfig = new Properties();
+        uncommittedConfig.putAll(consumerConfig);
+        uncommittedConfig.put(ConsumerConfig.ISOLATION_LEVEL_CONFIG, "read_uncommitted");
+        try (KafkaConsumer<byte[], byte[]> uncommittedConsumer =
+                new KafkaConsumer<>(uncommittedConfig)) {
+            final Map<TopicPartition, Long> highWatermarks =
```
MartijnVisser, line 221: Nit: the high watermarks are read once and the loop compares a moving last stable offset against that fixed target. Correct here because the drain runs after the job finished, but worth a line in the javadoc since the method reads as reusable.

MartijnVisser (commented):
Nothing in the PR fails without this. "10/10" is one pass over `rescaleListing`'s ten `@CsvSource` cases, and it's green on `main` too. A direct test of `getSettledEndOffsets` would cover it: open a transaction, assert you get the last stable offset; commit it, assert the wait picks up the higher offsets.

It also won't fix FLINK-40585 on its own. `rescaleListing` changes the prefix per job, so FLINK-40626's orphan is never reused and the 10 s just expires. I'd take #314 first.

slachiewicz:
Agreed on all points. Reworking it as: an opt-in variant of `drainAllRecordsFromTopic` for drains that expect open transactions, so the 10 s wait only applies to settled-topic drains; a direct test of `getSettledEndOffsets` for the open, warn, commit, advance sequence; `String.valueOf` for the isolation level; a javadoc line on the fixed high-watermark target. And yes, the "10/10" was one pass over the parameter sets, not ten runs. This PR waits for #314, which is what makes `rescaleListing` pass.

MartijnVisser:
This isn't fixing FLINK-40585. Runs 34473915461 and 30181683657 both show a precommitted id still ONGOING, so they're FLINK-40626 and #314 is what closes them. Run 31916828021 is the only one without that signature, and it's the v4.0 cell: v4.0 still has `ThrottleUntilFirstCheckpointStrategy`, which FLINK-38696's `a90e8f48` replaced on main with `perCheckpoint(10)`, so the third job writes the rest in one transaction there and loses 966 records where main would lose ten. I can't tell from the log whether that last transaction was cut off by a late marker or hadn't finished committing.

The direction is right, the condition isn't. Wait on transaction state, not on the LSO catching up with the high watermark. I'd take #314 first and then move this to v4.0 under its own ticket, and I wouldn't carry it on main: the marker lands inside 2 ms and the drain spends over 3 ms building the consumer before it reads the offsets, so I couldn't hit the race here at all.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/testutils/KafkaUtil.java
MartijnVisser: Filter on transaction state instead. A commit in flight is `PrepareCommit`, a transaction the test left open is `Ongoing`, and `filterStates` is broker-side, so the wait never sees the second. On a prototype `testAbortOnClose` and `testAbortTransactionsAfterScaleInBeforeFirstCheckpoint` are back to their base times.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/testutils/KafkaUtil.java
MartijnVisser: This is worse than a nit. The commit marker takes an offset, so when it's the marker that's pending the watermark reads X and the LSO settles at X+1, and `endOffsets.equals(highWatermarks)` never holds. The wait then always runs the full 10 s and warns about open transactions while printing an LSO above the watermark.

slachiewicz:
Thanks for measuring this, Martijn. You're right on both counts: the two runs with the ONGOING id are FLINK-40626, and the v4.0 cell needs its own look at whether that last transaction was cut off by a late marker or never finished committing. With the window not opening on main, there is nothing for this change to fix here, so I'm closing it. Appreciate the careful review and the pointer to `a90e8f48`.

PR #314, opened 2026-09-10
[FLINK-40626][Connectors/Kafka] Abort superseded precommitted transactions on recovery

slachiewicz:

With `TransactionNamingStrategy.POOLING`, a committed transactional id is reused for a later checkpoint and `initTransactions()` bumps its epoch. A recovery from the earlier checkpoint lists the id as precommitted, so `TransactionAbortStrategyImpl.LISTING` skips it, the committer's re-commit is fenced, and the newer transaction under the id stays ONGOING on the broker, pinning the last stable offset for `read_committed` consumers.

With an unchanged prefix and parallelism the orphan is short-lived: POOLING takes the id back at the next checkpoint and `initTransactions` aborts it. It survives until `transaction.timeout.ms` whenever the id is never recycled again: after a downscale, because new transactions only use `ownedSubtaskIds[0]` and an orphan under a secondary owned subtask id is never reused, and after a change of `transactionalIdPrefix` (how `rescaleListing` hits it, FLINK-40585). The downscale case needs no user error.

The fix aborts that transaction at recovery, the one point where it is provably ownerless: before the new attempt holds any id, a transaction the broker reports under a precommitted id with a newer epoch than the snapshotted one (or a different producer id) can only belong to the failed attempt or to a zombie, and the newest attempt is the side entitled to fence. The committer is unchanged, so a fenced commit from an older attempt can never fence a live owner.

The aborted id stays reserved in the producer pool. The committer's fenced commit of the restored transaction reports the id as finished through the backchannel, and the writer drains that only at its next snapshot; if POOLING took the id for the first new transaction in between, the stale report would close the new producer after `prepareCommit` handed it to the committer and lose that checkpoint. `KafkaSinkRecoveryITCase` (#322) pins this. The id is released by that report, or by the existing sweep once a later checkpoint's transaction finishes.

## Brief change log

- `CheckpointTransaction` records producer id and epoch; `KafkaWriterStateSerializer` is at v3. v2 state reads back with unknown epoch and keeps the previous skip.
- `ProducerPoolImpl.getOngoingTransactions()` reports the live producer's id and epoch; new `abortRestoredTransaction` fences the broker-side transaction through an unregistered producer and keeps the restored entry, so the id is not reused until `recycleByTransactionId` releases it.
- `AdminUtils.describeTransactions` (broker 3.0+, `Describe` on the transactional id, implied by `Write`).
- `LISTING` aborts a precommitted id when the context reports it superseded; `TransactionAbortStrategyContextImpl` describes the precommitted ids once, compares, and logs one WARN with the recorded and reported producer id, epoch and state.

## Verifying this change

- `KafkaSinkRecoveryITCase` (from #322, unchanged): fails on the previous revision of this branch rebased on `main` (`[0L]` instead of `[0L, 1L]`), passes with this change, 5 consecutive runs.
- `ExactlyOnceKafkaWriterITCase#shouldAbortSupersededPrecommittedTransactionOnRecovery`: checkpoint 1 committed, its id reused for checkpoint 3 under a bumped epoch, recovery from checkpoint 1 with the same prefix; no ONGOING transaction survives, the first new transaction does not take the reserved id, and a fenced report for the restored transaction leaves the new committable committable. The last two assertions fail without the reservation.
- `ProducerPoolImplITCase`: id stays reserved with the recorded epoch after the abort, the broker no longer reports it ONGOING, the abort producer returns to the pool, release through `recycleByTransactionId` and through the sweep.
- `TransactionAbortStrategyImplTest`: superseded abort goes through the context only, the generic aborter is not invoked, equal-epoch skip. `KafkaWriterStateSerializerTest`: v3 round trip and v2 bytes read with unknown epoch. These are red by compilation only.

Verified: `./mvnw test -pl flink-connector-kafka -Dtest='KafkaSinkRecoveryITCase,KafkaSinkITCase,ExactlyOnceKafkaWriterITCase,ProducerPoolImplITCase,TransactionAbortStrategyImplTest,KafkaWriterStateSerializerTest,KafkaCommitterTest,*ArchitectureTest'` → 0 failures; `spotless:check checkstyle:check` and `verify -DskipTests` (japicmp) clean under JDK 17.

- Dependencies: no
- Public API: no; all touched types are `@Internal`
- Checkpointed state: yes, `KafkaWriterStateSerializer` 2 → 3, backwards compatible on read
- Per-record code paths: no
- Anything that affects deployment or recovery: yes, recovery now issues `describeTransactions` for precommitted ids

## Documentation

- New feature: no

Backports: v5.0 and v4.0 (POOLING is not on v3.4).

Generated-by: Claude Code (Opus 5, Sonnet 5)

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriterStateSerializer.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriterStateSerializer.java
@@ -1,12 +1,13 @@
 /*
- * Licensed to the Apache Software Foundation (ASF) under one or more
- * contributor license agreements.  See the NOTICE file distributed with
- * this work for additional information regarding copyright ownership.
- * The ASF licenses this file to You under the Apache License, Version 2.0
- * (the "License"); you may not use this file except in compliance with
- * the License.  You may obtain a copy of the License at
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
  *
- *    http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -31,11 +32,17 @@ import java.util.Collection;
 
 import static org.apache.flink.connector.kafka.sink.KafkaWriterState.UNKNOWN;
 
-/** A serializer used to serialize {@link KafkaWriterState}. */
+/**
+ * A serializer used to serialize {@link KafkaWriterState}.
+ *
+ * <p>Version 3 adds the producer id and epoch of every precommitted transaction, so that a recovery
+ * can tell a transaction the committer still has to commit from a later transaction that reused the
+ * same transactional id. State written by version 2 is read with unknown producer id and epoch.
+ */
 class KafkaWriterStateSerializer implements SimpleVersionedSerializer<KafkaWriterState> {
     @Override
     public int getVersion() {
-        return 2;
+        return 3;
     }
 
     @Override
@@ -50,6 +57,8 @@ class KafkaWriterStateSerializer implements SimpleVersionedSerializer<KafkaWrite
             for (CheckpointTransaction transaction : state.getPrecommittedTransactionalIds()) {
                 out.writeUTF(transaction.getTransactionalId());
                 out.writeLong(transaction.getCheckpointId());
+                out.writeLong(transaction.getProducerId());
+                out.writeShort(transaction.getEpoch());
             }
             out.flush();
             return baos.toByteArray();
@@ -58,7 +67,7 @@ class KafkaWriterStateSerializer implements SimpleVersionedSerializer<KafkaWrite
 
     @Override
     public KafkaWriterState deserialize(int version, byte[] serialized) throws IOException {
-        if (version > 2) {
+        if (version > 3) {
             throw new IOException("Unknown version: " + version);
         }
 
@@ -69,14 +78,25 @@ class KafkaWriterStateSerializer implements SimpleVersionedSerializer<KafkaWrite
             int totalNumberOfOwnedSubtasks = UNKNOWN;
             TransactionOwnership transactionOwnership = TransactionOwnership.IMPLICIT_BY_SUBTASK_ID;
             final Collection<CheckpointTransaction> precommitted = new ArrayList<>();
-            if (version == 2) {
+            if (version >= 2) {
                 ownedSubtaskId = in.readInt();
                 totalNumberOfOwnedSubtasks = in.readInt();
                 transactionOwnership = TransactionOwnership.values()[in.readInt()];
 
                 final int usedTransactionIdsSize = in.readInt();
                 for (int i = 0; i < usedTransactionIdsSize; i++) {
-                    precommitted.add(new CheckpointTransaction(in.readUTF(), in.readLong()));
+                    final String transactionalId = in.readUTF();
+                    final long checkpointId = in.readLong();
+                    if (version >= 3) {
+                        precommitted.add(
+                                new CheckpointTransaction(
+                                        transactionalId,
+                                        checkpointId,
+                                        in.readLong(),
+                                        in.readShort()));
+                    } else {
+                        precommitted.add(new CheckpointTransaction(transactionalId, checkpointId));
+                    }
                 }
             }
             return new KafkaWriterState(
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/ProducerPool.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/ProducerPool.java
@@ -40,8 +40,22 @@ public interface ProducerPool extends AutoCloseable {
     FlinkKafkaInternalProducer<byte[], byte[]> getTransactionalProducer(
             String transactionalId, long checkpointId);
 
-    /** Returns a snapshot of all ongoing transactions. */
+    /**
+     * Returns a snapshot of all ongoing transactions. Transactions opened by this pool carry the
+     * producer id and epoch of their producer; transactions restored from state keep whatever the
+     * state recorded.
+     */
     Collection<CheckpointTransaction> getOngoingTransactions();
+
+    /**
+     * Stops tracking a transaction that was restored from state without touching a producer. Used
+     * on recovery when the broker no longer holds the restored transaction under its transactional
+     * id because the id was reused for a later checkpoint, so that the id can be aborted and
+     * reused.
+     *
+     * @throws IllegalStateException if the id is not a restored transaction or has a live producer
+     */
+    void abandonTransaction(String transactionalId);
 
     /**
      * Explicitly recycle a producer. This is useful when the producer has not been passed to the
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/ProducerPoolImpl.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/ProducerPoolImpl.java
@@ -34,6 +34,7 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.Deque;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.NavigableMap;
 import java.util.Objects;
@@ -254,7 +255,39 @@ public class ProducerPoolImpl implements ProducerPool {
 
     @Override
     public Collection<CheckpointTransaction> getOngoingTransactions() {
-        return new ArrayList<>(transactionalIdsByCheckpoint.keySet());
+        List<CheckpointTransaction> ongoing = new ArrayList<>(transactionalIdsByCheckpoint.size());
+        for (Map.Entry<CheckpointTransaction, String> entry :
+                transactionalIdsByCheckpoint.entrySet()) {
+            CheckpointTransaction transaction = entry.getKey();
+            ProducerEntry producerEntry = producerByTransactionalId.get(entry.getValue());
+            FlinkKafkaInternalProducer<byte[], byte[]> producer =
+                    producerEntry == null ? null : producerEntry.getProducer();
+            if (producer == null) {
+                // restored from state; keep what the state knows
+                ongoing.add(transaction);
+            } else {
+                ongoing.add(
+                        new CheckpointTransaction(
+                                transaction.getTransactionalId(),
+                                transaction.getCheckpointId(),
+                                producer.getProducerId(),
+                                producer.getEpoch()));
+            }
+        }
+        return ongoing;
+    }
+
+    @Override
+    public void abandonTransaction(String transactionalId) {
+        ProducerEntry producerEntry = producerByTransactionalId.get(transactionalId);
+        checkState(
+                producerEntry != null && producerEntry.getProducer() == null,
+                "Transaction %s is not a restored transaction without a producer: %s",
+                transactionalId,
+                producerEntry);
+        producerByTransactionalId.remove(transactionalId);
+        transactionalIdsByCheckpoint.remove(producerEntry.getCheckpointedTransaction());
+        LOG.debug("Abandoned restored transaction {}", transactionalId);
     }
 
     @VisibleForTesting
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyContextImpl.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyContextImpl.java
@@ -23,12 +23,21 @@ import org.apache.flink.connector.kafka.sink.internal.TransactionAbortStrategyIm
 import org.apache.flink.connector.kafka.util.AdminUtils;
 
 import org.apache.kafka.clients.admin.Admin;
+import org.apache.kafka.clients.admin.TransactionDescription;
 import org.apache.kafka.clients.admin.TransactionListing;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
 
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
@@ -38,6 +47,9 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
 /** Implementation of {@link TransactionAbortStrategyImpl.Context}. */
 @Internal
 public class TransactionAbortStrategyContextImpl implements TransactionAbortStrategyImpl.Context {
+    private static final Logger LOG =
+            LoggerFactory.getLogger(TransactionAbortStrategyContextImpl.class);
+
     private final Supplier<Admin> adminSupplier;
     private final Supplier<Collection<String>> topicNames;
     private final int currentSubtaskId;
@@ -48,8 +60,14 @@ public class TransactionAbortStrategyContextImpl implements TransactionAbortStra
     private final long startCheckpointId;
     private final TransactionAborter transactionAborter;
 
-    /** Transactional ids that mustn't be aborted. */
-    private final Set<String> precommittedTransactionIds;
+    /** Transactions that mustn't be aborted unless the broker no longer holds them. */
+    private final Map<String, CheckpointTransaction> precommittedTransactions;
+
+    /** Drops a superseded precommitted transaction from the writer's bookkeeping. */
+    private final Consumer<String> precommittedTransactionAbandoner;
+
+    /** Broker-side descriptions of the precommitted transactional ids, fetched on first use. */
+    @Nullable private Map<String, TransactionDescription> precommittedDescriptions;
 
     /** Creates a new {@link TransactionAbortStrategyContextImpl}. */
     public TransactionAbortStrategyContextImpl(
@@ -62,7 +80,8 @@ public class TransactionAbortStrategyContextImpl implements TransactionAbortStra
             long startCheckpointId,
             TransactionAborter transactionAborter,
             Supplier<Admin> adminSupplier,
-            Set<String> precommittedTransactionIds) {
+            Collection<CheckpointTransaction> precommittedTransactions,
+            Consumer<String> precommittedTransactionAbandoner) {
         this.topicNames = checkNotNull(topicNames, "topicNames must not be null");
         this.currentSubtaskId = currentSubtaskId;
         this.currentParallelism = currentParallelism;
@@ -73,9 +92,15 @@ public class TransactionAbortStrategyContextImpl implements TransactionAbortStra
         this.transactionAborter =
                 checkNotNull(transactionAborter, "transactionAborter must not be null");
         this.adminSupplier = checkNotNull(adminSupplier, "adminSupplier must not be null");
-        this.precommittedTransactionIds =
+        checkNotNull(precommittedTransactions, "precommittedTransactions must not be null");
+        this.precommittedTransactions = new HashMap<>();
+        for (CheckpointTransaction transaction : precommittedTransactions) {
+            this.precommittedTransactions.put(transaction.getTransactionalId(), transaction);
+        }
+        this.precommittedTransactionAbandoner =
                 checkNotNull(
-                        precommittedTransactionIds, "transactionsToBeCommitted must not be null");
+                        precommittedTransactionAbandoner,
+                        "precommittedTransactionAbandoner must not be null");
     }
 
     @Override
@@ -111,7 +136,50 @@ public class TransactionAbortStrategyContextImpl implements TransactionAbortStra
 
     @Override
     public Set<String> getPrecommittedTransactionalIds() {
-        return precommittedTransactionIds;
+        return Collections.unmodifiableSet(precommittedTransactions.keySet());
+    }
+
+    @Override
+    public boolean isPrecommittedTransactionSuperseded(String transactionalId) {
+        CheckpointTransaction transaction = precommittedTransactions.get(transactionalId);
+        if (transaction == null || !transaction.hasKnownEpoch()) {
+            // state written before v3 did not record the epoch; keep the old behavior
+            return false;
+        }
+        TransactionDescription description = describePrecommitted().get(transactionalId);
+        if (description == null) {
+            // the broker does not know the id any more; nothing to abort
+            return false;
+        }
+        boolean superseded =
+                description.producerId() != transaction.getProducerId()
+                        || description.producerEpoch() > transaction.getEpoch();
+        if (superseded) {
+            LOG.info(
+                    "Recovered transaction {} was opened with producer id {} and epoch {}, but the broker now holds producer id {} and epoch {} in state {}",
+                    transactionalId,
+                    transaction.getProducerId(),
+                    transaction.getEpoch(),
+                    description.producerId(),
+                    description.producerEpoch(),
+                    description.state());
+        }
+        return superseded;
+    }
+
+    @Override
+    public void abandonPrecommittedTransaction(String transactionalId) {
+        precommittedTransactions.remove(transactionalId);
+        precommittedTransactionAbandoner.accept(transactionalId);
+    }
+
+    private Map<String, TransactionDescription> describePrecommitted() {
+        if (precommittedDescriptions == null) {
+            precommittedDescriptions =
+                    AdminUtils.describeTransactions(
+                            adminSupplier.get(), precommittedTransactions.keySet());
+        }
+        return precommittedDescriptions;
     }
 
     @Override
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java
@@ -143,6 +143,18 @@ public enum TransactionAbortStrategyImpl {
             TransactionAborter transactionAborter = context.getTransactionAborter();
             for (String name : openTransactionsForSubtask) {
                 if (context.getPrecommittedTransactionalIds().contains(name)) {
+                    if (context.isPrecommittedTransactionSuperseded(name)) {
+                        // The broker holds a later transaction under this id than the one the
+                        // committer is about to commit. That commit will be fenced, and nobody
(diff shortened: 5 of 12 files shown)


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitter.java
```diff
@@ -184,6 +188,33 @@ private void logFencedRequest(
...
+        } catch (KafkaException e) {
+            LOG.warn(
+                    "Could not abort the open transaction under {}; it stays open until '{}' expires.",
+                    transactionalId,
+                    ProducerConfig.TRANSACTION_TIMEOUT_CONFIG,
+                    e);
+        } finally {
+            producer.close();
+        }
+    }
```
spuru9: major: `catch (KafkaException)` only covers `initTransactions()`. `producerFactory.apply(...)` (before the `try`) and `close()` (in `finally`) can throw too, and since this runs inside the `ProducerFencedException` catch, an escape skips `handleFailedTransaction` and `signalFailedWithKnownReason` — best-effort cleanup failing the commit path. try-with-resources guards both the acquisition and the close:

```suggestion
    private void abortNewerTransaction(String transactionalId) {
        try (FlinkKafkaInternalProducer<?, ?> producer =
                producerFactory.apply(kafkaProducerConfig, transactionalId)) {
            producer.initTransactions();
            LOG.info(
                    "Bumped the epoch under {} to abort any lingering transaction of a newer epoch after its commit was fenced.",
                    transactionalId);
        } catch (KafkaException e) {
            LOG.warn(
                    "Could not abort the open transaction under {}; it stays open until '{}' expires.",
                    transactionalId,
                    ProducerConfig.TRANSACTION_TIMEOUT_CONFIG,
                    e);
        }
    }
```

Also reworded the `LOG.info` (init bumps the epoch even when nothing was open, so "Aborted…" overclaims). Widen to `catch (RuntimeException)` if you want it airtight against non-`KafkaException` throws.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitter.java
```diff
@@ -108,6 +109,9 @@ public void commit(Collection<CommitRequest<KafkaCommittable>> requests)
                 request.retryLater();
             } catch (ProducerFencedException e) {
                 logFencedRequest(request, e);
+                if (reusesTransactionalIds) {
+                    abortNewerTransaction(transactionalId);
```
MartijnVisser: A fenced commit also means we might be the stale side: an unreachable-but-alive TM whose committer arrives after the new attempt took the id over. Then this bumps the epoch and fences the healthy attempt. I checked both directions against a broker. On `main` a live owner of the id still commits, here it gets `ProducerFencedException`. Can you show that window is closed, or bump only when the transaction is provably orphaned?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitter.java
```diff
@@ -184,6 +188,33 @@ private void logFencedRequest(
...
+     * whose transaction is still open on the broker under a newer epoch. Nobody owns that
+     * transaction any more: the writer skips the id as precommitted and this committer cannot
+     * commit it. Bumping the epoch aborts it, so that it does not block read_committed consumers
+     * until the transaction timeout expires.
+     */
+    private void abortNewerTransaction(String transactionalId) {
+        FlinkKafkaInternalProducer<?, ?> producer =
+                producerFactory.apply(kafkaProducerConfig, transactionalId);
+        try {
+            producer.initTransactions();
```
MartijnVisser: `max.block.ms` isn't set by `KafkaSinkBuilder`, so this blocks up to the 60s default per fenced request, serially. `KafkaCommitterTest` pins it to 100 so it doesn't show. Please set a short one on a copy of the config.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitter.java
```diff
@@ -184,6 +188,33 @@ private void logFencedRequest(
...
+            LOG.info(
+                    "Aborted open transaction of a newer epoch under {} after its commit was fenced.",
+                    transactionalId);
+        } catch (KafkaException e) {
```
MartijnVisser: `InterruptException extends KafkaException`, so a cancel here is swallowed and only surfaces on the next request. `commit()` converts it a few lines up for that reason. Skip the abort when the thread is already interrupted.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitterTest.java
```diff
@@ -219,6 +221,49 @@ private AtomicBoolean interruptOnMessage(Thread mainThread, ServerSocket serverS
...
+            assertThat(backchannel).has(transactionFinished(false));
+            assertThat(committer.getCommittingProducer()).isNull();
+            assertThat(createdProducers).allMatch(FlinkKafkaInternalProducer::isClosed);
+            // with reused ids the fenced id may hold a newer, orphaned transaction; bumping the
+            // epoch through initTransactions aborts it
+            long abortedIds =
+                    createdProducers.stream()
+                            .filter(MockProducer::isTransactionsInitialized)
+                            .count();
+            assertThat(abortedIds).isEqualTo(reusesTransactionalIds ? 1 : 0);
```
MartijnVisser: `MockProducer` ignores the id the factory is handed (`super(properties, TRANS_ID)`, line 316) and `initTransactions()` just sets a flag, so this passes even with the wrong id. Assert the captured id, and add a broker-level case; `ExactlyOnceKafkaWriterITCase` has the fixtures.

MartijnVisser (changes requested):
Reproduced against a broker: the orphan stays `Ongoing` on `main` and is aborted here, and the new `KafkaCommitterTest` case is red/green as claimed. Blocking question on line 113.

Note "10/10" is one pass over `rescaleListing`'s ten `@CsvSource` cases, not ten runs. Backports v5.0 and v4.0, `POOLING` isn't on v3.4.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitter.java
slachiewicz: You are right, and the window is not closable from here. The orphan and the zombie look identical to the broker: same producer id, epoch plus one, ONGOING. In the orphan case `getTransactionalProducer` re-initialised the id for a later checkpoint; in your case the new attempt's committer committed at the old epoch, the backchannel freed the id, pooling reused it and `initTransactions` bumped it. So `describeTransactions` in the committer cannot prove anything either.

Moved the abort to where it is provable: `ExactlyOnceKafkaWriter.initialize()` at recovery, inside `TransactionAbortStrategyImpl.LISTING`, before the new attempt holds any id. There, an open transaction under a precommitted id whose broker epoch is newer than the one the writer snapshotted (or whose producer id differs) can only belong to the dead attempt or to a zombie, and the newest attempt is the side that is supposed to fence. This also covers the changed-prefix case from FLINK-40585, which the committer-side bump would not have.

Cost: the writer state did not carry the epoch, so `CheckpointTransaction` now records producer id and epoch and `KafkaWriterStateSerializer` is at v3; v2 state reads back with unknown epoch and keeps the previous skip. `describeTransactions` needs broker 3.0+ like `listTransactions` that LISTING already uses, and `Describe` on the transactional id, which `Write` implies.

Regression guard is a changed-prefix recovery in `ExactlyOnceKafkaWriterITCase` asserting no ONGOING transaction survives; it fails without the strategy change. Trailer and Release Note added. The inline threads below are outdated by the move; I have replied on each.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitter.java
slachiewicz: Superseded: the committer-side bump is gone in 05b8847f; the abort now happens at recovery in `TransactionAbortStrategyImpl.LISTING` (see the thread on line 113), so this code no longer exists.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitter.java
slachiewicz: Superseded: the committer-side bump is gone in 05b8847f; the abort now happens at recovery in `TransactionAbortStrategyImpl.LISTING` (see the thread on line 113), so this code no longer exists.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitter.java
slachiewicz: Superseded: the committer-side bump is gone in 05b8847f; the abort now happens at recovery in `TransactionAbortStrategyImpl.LISTING` (see the thread on line 113), so this code no longer exists.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/internal/KafkaCommitterTest.java
slachiewicz: Superseded: that test went with the committer change in 05b8847f. The guard is now `ExactlyOnceKafkaWriterITCase#shouldAbortSupersededPrecommittedTransactionOnRecovery` against the broker: checkpoint 1 committed, its id reused for checkpoint 3 under a bumped epoch, recovery from checkpoint 1 with a new prefix, then `AdminUtils.getOpenTransactionsForTopics` must be empty. Without the strategy change it fails with the reused id still ONGOING.

MartijnVisser:
Concept is right and I would like this to land. Three things from looking at the surrounding design.

Measured against a broker: with an unchanged prefix the orphan is cleared at the first checkpoint after the fenced commit, because POOLING takes the id back and `initTransactions` aborts it. It only survives when the id is never recycled: a changed `transactionalIdPrefix`, as `rescaleListing` does, or a downscale, where new transactions only use `ownedSubtaskIds[0]` so an orphan under a secondary owned subtask id is never reused. The downscale case needs no user error and is the strongest justification for the change. Worth putting in the description and in the FLINK-40626 release note in place of the "~1 hour" framing.

On the live owner question above: the window is narrow, but it ends in silent data loss rather than a failover, since a fenced commit is `signalFailedWithKnownReason` and the committable is discarded. One `describeTransactions` before the bump closes it: abort only when the open transaction started before this committer recovered.

FLINK-39218 (#296) adds the same epoch bump as a CLI abort, with a plain `KafkaProducer` and `max.block.ms` at 10s. One shared helper in the connector beats two implementations, and it settles the `max.block.ms` comment above. @Savonitar I think I'd like you to also review this PR in full

slachiewicz:
Thanks. All three taken, with two of them turning out differently because of the rewrite in eb13af06, which I think crossed with your comment:

1. Downscale: agreed, and it is the strongest case. Description and release note now lead with it, and `testAbortsSupersededPrecommittedTransactionOfSecondaryOwnedSubtask` covers it: `ownsTransactionalId` includes the secondary owned subtask ids and their precommitted transactions come from the same restored states, so the recovery aborts them.
2. Live owner: the committer no longer bumps at all. The abort runs in `LISTING` at recovery, before the new attempt holds any id, on a precommitted id whose broker epoch is newer than the snapshotted one or whose producer id differs. The only possible holders at that point are the failed attempt and a zombie, and fencing a zombie from the newest attempt is the intended direction, so no start-time comparison is needed. `describeTransactions` is what decides it, in `AdminUtils` and `TransactionAbortStrategyContextImpl`.
3. FLINK-39218: this version adds no epoch-bump helper; the abort goes through the existing `TransactionAborter`, the same pool-and-`initTransactions` path `LISTING` already uses for non-precommitted ids. `AdminUtils.describeTransactions` is reusable by #296 if the tool wants producer id, epoch and state per id. Happy to have @Savonitar review.

Local runs on eb13af06: the module unit suite (441), and every ITCase in the sink package against a Kafka container (`KafkaSinkITCase` 59, `ExactlyOnceKafkaWriterITCase` 14, `KafkaWriterITCase` 12, `FlinkKafkaInternalProducerITCase` 9, `ProducerPoolImplITCase` 8, `KafkaWriterFaultToleranceITCase` 4, `KafkaTransactionLogITCase` 1), all green.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java
```diff
@@ -143,6 +143,18 @@ public void abortTransactions(Context context) {
...
                 if (context.getPrecommittedTransactionalIds().contains(name)) {
+                    if (context.isPrecommittedTransactionSuperseded(name)) {
+                        // The broker holds a later transaction under this id than the one the
+                        // committer is about to commit. That commit will be fenced, and nobody
+                        // owns the open transaction; abort it so that it does not pin the last
+                        // stable offset until the transaction timeout.
+                        LOG.warn(
+                                "Aborting open transaction {}: the recovered transaction under this id was superseded by a newer epoch",
+                                name);
+                        context.abandonPrecommittedTransaction(name);
```
Savonitar, line 154: I think here we have a tricky issue:
With unchanged prefix (the common/default case), this method frees the ID and `POOLING` reuses it immediately, however the restored committer's fenced commit still can send notification `TransactionFinished.erroneously (ID)` and `recycleByTransactionId` that matches by transactional id.
 Can that stale notification land on the newly reused transaction? If yes, it means in that case we will have a dataloss.

To improve the robustness of this repository, I raised a separate **independent** PR with an additional test coverage https://github.com/apache/flink-connector-kafka/pull/322 
It is a general test (not dedicated for this PR/ticket). But it catches this issue too.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriterITCase.java
```diff
@@ -275,6 +275,61 @@ void shouldNotAbortPrecommittedTransactions(int numCheckpointed) throws Exceptio
...
+                    .isEqualTo(precommitted.getTransactionalId());
+            assertThat(checkpoint3.getEpoch()).isGreaterThan(precommitted.getEpoch());
+            // the job fails here; the transactions of checkpoints 2 and 3 linger on the broker
+        }
+
+        try (AdminClient admin = AdminClient.create(getKafkaClientConfiguration())) {
+            assertThat(AdminUtils.getOpenTransactionsForTopics(admin, Collections.singleton(topic)))
+                    .hasSize(2);
+
+            // recovery from checkpoint 1; the new writer gets a new prefix, so the old id is never
```
Savonitar, line 318: >  the new writer gets a new prefix,

Do we test only changed prefix here? So, the reused prefix is never excerciced. Should we add a same-prefix test case with the restored commiter?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyContextImpl.java
```diff
@@ -111,7 +136,50 @@ public Set<String> getPrefixesToAbort() {
 
     @Override
     public Set<String> getPrecommittedTransactionalIds() {
-        return precommittedTransactionIds;
+        return Collections.unmodifiableSet(precommittedTransactions.keySet());
+    }
+
+    @Override
+    public boolean isPrecommittedTransactionSuperseded(String transactionalId) {
```
Savonitar, line 143: Is this method tested only by IT case tests?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyContextImpl.java
```diff
@@ -111,7 +136,50 @@ public Set<String> getPrefixesToAbort() {
...
+    @Override
+    public void abandonPrecommittedTransaction(String transactionalId) {
+        precommittedTransactions.remove(transactionalId);
+        precommittedTransactionAbandoner.accept(transactionalId);
+    }
+
+    private Map<String, TransactionDescription> describePrecommitted() {
+        if (precommittedDescriptions == null) {
+            precommittedDescriptions =
+                    AdminUtils.describeTransactions(
```
Savonitar, line 179: just to confirm, the frequency of calling this method is not high? (according to PR description it is once). Is it once per subtask restart?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriterStateSerializer.java
```diff
@@ -31,11 +32,17 @@
 
 import static org.apache.flink.connector.kafka.sink.KafkaWriterState.UNKNOWN;
 
-/** A serializer used to serialize {@link KafkaWriterState}. */
+/**
+ * A serializer used to serialize {@link KafkaWriterState}.
+ *
+ * <p>Version 3 adds the producer id and epoch of every precommitted transaction, so that a recovery
```
Savonitar, line 38: Is a downgrade/rollback path considered or we expect forward-only for this bump? (I see PR says "backwards compatible on read").

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/ProducerPoolImpl.java
```diff
@@ -254,7 +255,39 @@ public FlinkKafkaInternalProducer<byte[], byte[]> getTransactionalProducer(
...
+
+    @Override
+    public void abandonTransaction(String transactionalId) {
+        ProducerEntry producerEntry = producerByTransactionalId.get(transactionalId);
+        checkState(
+                producerEntry != null && producerEntry.getProducer() == null,
+                "Transaction %s is not a restored transaction without a producer: %s",
+                transactionalId,
+                producerEntry);
+        producerByTransactionalId.remove(transactionalId);
```
Savonitar, line 288: could we keep the "superseeded" transactionalId reserved after abandoning? Removing this entry allows `POOLING` to reuse it during writer initialization, meanwhile a completion notification for the old transaction may still be queued.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java
```diff
@@ -143,6 +143,18 @@ public void abortTransactions(Context context) {
             TransactionAborter transactionAborter = context.getTransactionAborter();
             for (String name : openTransactionsForSubtask) {
                 if (context.getPrecommittedTransactionalIds().contains(name)) {
+                    if (context.isPrecommittedTransactionSuperseded(name)) {
+                        // The broker holds a later transaction under this id than the one the
+                        // committer is about to commit. That commit will be fenced, and nobody
+                        // owns the open transaction; abort it so that it does not pin the last
+                        // stable offset until the transaction timeout.
+                        LOG.warn(
```
Savonitar, line 151: we have this WARN log and INFO log from [public boolean isPrecommittedTransactionSuperseded(String transactionalId) {](https://github.com/apache/flink-connector-kafka/pull/314/changes#diff-2b2784802b7feb204128c1f72e9523037a202f417cb74ff3698cd2fe6ce45017R143)  
> LOG.info(
                    "Recovered transaction {} was opened with producer id {} and epoch {}, but the broker now holds producer id {} and epoch {} in state {}",

do we need to keep both?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyContextImpl.java
```diff
@@ -111,7 +136,50 @@ public Set<String> getPrefixesToAbort() {
...
+        if (description == null) {
+            // the broker does not know the id any more; nothing to abort
+            return false;
+        }
+        boolean superseded =
+                description.producerId() != transaction.getProducerId()
+                        || description.producerEpoch() > transaction.getEpoch();
+        if (superseded) {
+            LOG.info(
+                    "Recovered transaction {} was opened with producer id {} and epoch {}, but the broker now holds producer id {} and epoch {} in state {}",
```
Savonitar, line 159: Should we use "the broker reported" instead of "the broker now holds" here and avoid calling the transaction "open" in the subsequent warning? 

The `ONGOING` filter applies to the earlier `listTransactions()` result, while `describeTransactions()` is a separate call whose results are cached. The coordinator could abort the transaction between these observations or before cleanup, so **neither log guarantees** it is still **open**.
Or am I missing something?

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/ProducerPool.java
```diff
@@ -40,9 +40,23 @@ public interface ProducerPool extends AutoCloseable {
...
+     * producer id and epoch of their producer; transactions restored from state keep whatever the
+     * state recorded.
+     */
     Collection<CheckpointTransaction> getOngoingTransactions();
 
+    /**
+     * Stops tracking a transaction that was restored from state without touching a producer. Used
+     * on recovery when the broker no longer holds the restored transaction under its transactional
+     * id because the id was reused for a later checkpoint, so that the id can be aborted and
+     * reused.
```
Savonitar, line 54: > reused

Maybe this comment should be clarified a bit after you check https://github.com/apache/flink-connector-kafka/pull/314/changes#r4039525893 
(because then it can not be reused ASAP).

Savonitar (changes requested):
Hi. Thanks for the PR!
I left some comments. I think this PR introduces a rare and a bit tricky dataloss.
Could you please take a look?

slachiewicz:
Reproduced, thanks. With `KafkaSinkRecoveryITCase` from #322 on main, this branch merged onto main fails: record 1 never becomes visible, because POOLING reuses the abandoned id for the first post-restore transaction before the committer's fenced notification for the old transaction is drained, and that notification then closes the new producer. Working on keeping the superseded id reserved until the committer releases it, plus the wording comments. Converting to draft until that is in.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/ProducerPoolImpl.java
slachiewicz: Yes, and it was more than a race: with `KafkaSinkRecoveryITCase` from #322 on top of this branch, POOLING took the id for the first post-restore transaction and the fenced report then closed that producer, losing record 1. `abandonTransaction` is now `abortRestoredTransaction`: it fences the broker-side transaction through a producer that is never registered under the id and keeps the restored entry, so the id stays in `getOngoingTransactions()` until `recycleByTransactionId` releases it (the fenced report, or the sweep once a later checkpoint finishes). `ProducerPoolImplITCase#testAbortRestoredTransaction` and `#testAbortRestoredTransactionReleasedBySweep` cover both release paths; the recovery ITCase is green on 5 runs.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java
slachiewicz: One line now: the WARN in the strategy is gone and the context logs a single WARN with the recorded and reported producer id, epoch and state, where the values are known.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyContextImpl.java
slachiewicz: You are right, they are two separate observations. The message now says "the broker reported ... in state {}" and nothing in the diff calls the transaction open any more; the Context javadoc and the strategy comment were changed the same way.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/ProducerPool.java
slachiewicz: Rewritten with the reserved semantics: the id stays tracked as ongoing and is not reused until the committer reports the outcome of the restored transaction or a later checkpoint's transaction is reported finished.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyImpl.java
slachiewicz: Yes, it could, and #322 catches exactly that: on the previous revision only record 0 survived the restore. Fixed by keeping the id reserved; details in the thread on `ProducerPoolImpl` below.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriterITCase.java
slachiewicz: Changed: the test now restores with the same prefix, which is what makes POOLING contend for the reserved id, and it asserts that the first new transaction does not take it and that the fenced report leaves the new committable intact. The restored committer end to end is `KafkaSinkRecoveryITCase` from #322.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyContextImpl.java
slachiewicz: Yes. `TransactionAbortStrategyImplTest` drives the strategy through a fake context, so this method itself runs only in `ExactlyOnceKafkaWriterITCase` and `KafkaSinkRecoveryITCase`, because the decision needs `Admin.describeTransactions`. Those cover the newer-epoch case; the "state has no epoch" branch is covered through `KafkaWriterStateSerializerTest` reading v2 bytes as unknown. If you want the full matrix (id unknown to the broker, equal epoch, different producer id) without a broker, I can add a unit test with a stub `Admin`.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/internal/TransactionAbortStrategyContextImpl.java
slachiewicz: Once per writer initialization, so once per subtask start or restart: `abortLingeringTransactions` creates the context in `initialize()`, the first superseded check issues one batched `describeTransactions` for all precommitted ids, and the result is cached for the lifetime of the context, which ends with that call. Nothing on the checkpoint or record path.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/sink/KafkaWriterStateSerializer.java
slachiewicz: Forward-only, like the 1 to 2 bump: a connector without this change throws `IOException("Unknown version: 3")` on restore, so a rollback needs a savepoint taken by the older version. Backports to v5.0 and v4.0 keep the version consistent across the supported lines.

PR #315, opened 2026-09-10
[BP-5.0][FLINK-40362][connector/kafka] Fix idle reader hang when no-more-splits precedes metadata update

slachiewicz:
Backport of #291 to v5.0, requested in https://github.com/apache/flink-connector-kafka/pull/291#issuecomment-5618687965.


MartijnVisser (approved):
Range-diff against 48ddbf1f is identical apart from the cherry-pick line. Ran DynamicKafkaSourceReaderTest on v5.0: the idle test fails with the production file reverted to the branch, the active-reader test fails without the first-update guard, both green on this head.

PR #317, opened 2026-09-10
[BP-3.4][FLINK-40362][connector/kafka] Fix idle reader hang when no-more-splits precedes metadata update

slachiewicz:
Backport of #291 to v3.4, requested in https://github.com/apache/flink-connector-kafka/pull/291#issuecomment-5618687965.

This branch's test file has no `TrackingReaderOutput` helper, so the two cherry-picked tests use `TestingReaderOutput` directly, matching the other tests already in this file.


MartijnVisser (approved):
Range-diff against 48ddbf1f is identical apart from the cherry-pick line and the TestingReaderOutput swap, which is inert here: the added tests assert nothing about released outputs and this branch's reader never releases them. Same red/green on v3.4: idle test red with the production file reverted, active-reader test red without the first-update guard, both green on this head.

PR #318, opened 2026-09-10
[FLINK-31006][Connectors/Kafka] Signal no more splits after a restored bounded enumerator finds no partition changes

MartijnVisser:

A bounded `KafkaSource` never finishes once the job has been restored from a savepoint or a retained checkpoint, or after a JobManager failover. All three recreate the enumerator from state. The recreated enumerator runs its one-time discovery, finds every subscribed partition already in `assignedSplits`, and `checkPartitionChanges` returns on the empty `PartitionChange` before it reaches `handlePartitionSplitChanges`, which is the only place that sets `noMoreNewPartitionSplits`. The restored readers consume up to their stopping offset and then wait forever for a `NoMoreSplitsEvent`, which the runtime never re-sends on its own. The same early return hangs a bounded source that subscribes to no existing partition on a fresh start.

The early return came in with `0b38d1a9` (FLINK-22147). Before that refactor, discovery and split initialisation shared one callable whose callback always ran, so a restored bounded source did signal correctly. The Pulsar enumerator does not have this bug, because its split assigner marks the discovery finished on every discovery rather than only on a change.

## Brief change log

  - `KafkaSourceEnumerator.checkPartitionChanges` lets an empty partition change through when a bounded source has run its only discovery, so that it reaches `handlePartitionSplitChanges` and signals the registered readers
  - Periodic discovery and unbounded sources keep returning early; neither can act on an empty change, and treating one as a finished discovery would make the partitions a later discovery finds resolve against the earliest offset rather than the configured one
  - `[hotfix][tests]` the `createEnumerator` helper in `KafkaSourceEnumeratorTest` dropped its `initialDiscoveryFinished` argument
  - `[hotfix][tests]` the context proxy in `DynamicKafkaSourceEnumeratorTest` passed `null` where production passes `DynamicKafkaSourceEnumerator::handleNoMoreSplits`, so no test could observe that wiring

Signalling from `addReader` on the strength of the restored `initialDiscoveryFinished` flag would need no new state, but it finishes a reader before the post-restore discovery assigns it a partition created while the job was down. That is what apache/flink#21909 was rejected for. The reasoning is written up on FLINK-31006, answering the question left open on FLINK-33466 in 2023.

## Verifying this change

  - Added `testRestoredBoundedEnumeratorSignalsNoMoreSplitsWithoutPartitionChanges` and `testBoundedSourceWithoutPartitionsSignalsNoMoreSplits` to `KafkaSourceEnumeratorTest` for the restore and the empty-subscription paths
  - Added `testBoundedRestoreSignalsNoMoreSplitsWithoutPartitionChanges` to `DynamicKafkaSourceEnumeratorTest` for the same restore through the sub-enumerators
  - Added `KafkaSourceBoundedRestoreITCase`, which produces to a single partition, takes a savepoint with `stopWithSavepoint`, produces more records and restores; no existing test restored a bounded `KafkaSource`
  - Each of those four fails without the production change. The ITCase times out after two minutes on the base and passes in about six seconds with the fix
  - Added `testUnboundedSourceKeepsSkippingAnEmptyPartitionChange` and `testRestoredBoundedEnumeratorAssignsBeforeItSignals`, which pass on both sides by design: the first fails if the condition is ever widened past bounded sources, the second fails under the `addReader` variant described above
  - `mvn verify` on the module with no skips: 443 unit tests and 326 integration tests, no failures, spotless, checkstyle, enforcer and japicmp clean, plus `apache-rat:check` at the root
  - Ran the new ITCase 20 consecutive times, all passing

  - Dependencies (does it add or upgrade a dependency, including `kafka.version` or `flink.version`): no
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)` or `@Experimental`, or are the Table options or the PyFlink wrappers changed: no
  - Checkpointed state, its serializers, or exactly-once delivery (splits, enumerator state, writer state, committables, transactions): yes. There is no state format change and no serializer change, and `noMoreNewPartitionSplits` stays derived rather than checkpointed. What does change is the value a bounded source snapshots for the existing `initialDiscoveryFinished` field when its only discovery finds nothing: previously it stayed `false`, now it becomes `true`, which is what the readers are signalled from. On the restore path the restored value is already `true`, so nothing moves there.
  - The per-record code paths (split reader, record emitter, writer, serialization schemas; performance sensitive): no

## Documentation

  - If yes, how is the feature documented? not applicable. The documented behaviour, that a bounded source exits once all partitions reach their stopping offsets, is what this restores.
  - If the docs changed, are both `docs/content` and `docs/content.zh` updated? not applicable

---

Generated-by: Claude Code 2.1.267 (Claude Opus 5)

### DIFF
--- a/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
+++ b/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/KafkaSourceEnumerator.java
@@ -365,7 +365,18 @@ public class KafkaSourceEnumerator
 
         final PartitionChange partitionChange =
                 getPartitionChange(fetchedPartitions, !initialDiscoveryFinished);
-        if (partitionChange.isEmpty()) {
+        // A bounded source that has just run its only discovery must reach
+        // handlePartitionSplitChanges even when nothing changed: that is the only place that sets
+        // noMoreNewPartitionSplits, and so the only path that signals the readers. After a restore
+        // every subscribed partition is already assigned, which makes the change empty
+        // (FLINK-31006). initializePartitionSplits performs no broker I/O for an empty change.
+        // Periodic discovery and unbounded sources keep returning early. Neither can act on an
+        // empty change, and treating it as a finished discovery would make the partitions that a
+        // later discovery finds resolve against the earliest offset rather than the configured
+        // one.
+        final boolean boundedSourceFinishedDiscovery =
+                boundedness == Boundedness.BOUNDED && partitionDiscoveryIntervalMs <= 0;
+        if (partitionChange.isEmpty() && !boundedSourceFinishedDiscovery) {
             return;
         }
         context.callAsync(
--- a/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
+++ b/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumeratorTest.java
@@ -2467,7 +2467,8 @@ public class DynamicKafkaSourceEnumeratorTest {
             return new TestKafkaEnumContextProxy(
                     kafkaClusterId,
                     kafkaMetadataService,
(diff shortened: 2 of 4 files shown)


MartijnVisser:
@Efrat19 @Savonitar If you have some time, I'd appreciate a look from you

PR #322, opened 2026-09-17
[FLINK-40703][tests] Cover KafkaSink recovery after pooled transaction ID reuse

Savonitar:

Adds deterministic integration coverage for exactly-once KafkaSink recovery with pooled transactional IDs. The test verifies that restoring an earlier checkpoint after a transaction ID has been reused preserves committed records and successfully commits replayed records without loss or duplication.
It is a usefull test that will prevent further regressions in this area. 

## Brief change log

- Add KafkaSinkRecoveryITCase with controlled record emission and manually triggered checkpoints.
- Complete checkpoint C1, keep C2 and C3 incomplete, and verify through Kafka that C1’s transactional ID has been reused.
- Cancel the job and restore C1 with the same transactional ID prefix and operator identities.
- Assert the committed output after recovery and verify that the final records appear exactly once.

## Verifying this change

- Added KafkaSinkRecoveryITCase#recoversPooledTransactionsWithoutLosingNewRecords.
- Passed many (20+) consecutive runs using the standard Maven/Testcontainers lifecycle with JDK 17 and Flink 2.2.1:./mvnw test -pl flink-connector-kafka -Dtest=KafkaSinkRecoveryITCase
- Spotless and Checkstyle checks passed.

  - Dependencies (does it add or upgrade a dependency, including `kafka.version` or `flink.version`): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)` or `@Experimental`, or are the Table options or the PyFlink wrappers changed: (no)
  - Checkpointed state, its serializers, or exactly-once delivery (splits, enumerator state, writer state, committables, transactions): (no, it adds **additional test coverage for exactly-once delivery**)
  - The per-record code paths (split reader, record emitter, writer, serialization schemas; performance sensitive): (no)

## Documentation

  - If the docs changed, are both `docs/content` and `docs/content.zh` updated? (not applicable)

---

Generated-by: Claude Fable 5.1


flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaSinkRecoveryITCase.java
```diff
@@ -0,0 +1,537 @@
...
+/**
+ * Exercises transactional id reuse across checkpoint recovery with a real sink pipeline.
+ *
+ * <p>C1 completes and commits its transaction. C2's asynchronous snapshot is blocked rather than
+ * failed so processing can continue through C3, which precommits a transaction reusing C1's
+ * transactional id, while C1 remains the latest completed checkpoint. The job is then cancelled and
+ * restored from C1 with the same transactional id prefix and operator identities.
+ *
+ * <p>This exercises the recovery fencing scenario described in {@link
+ * KafkaCommitter#logFencedRequest}. The restored job must replay the uncheckpointed records and
```
MartijnVisser, line 106: Nit: `logFencedRequest` is private, so javadoc cannot resolve this link. `{@code}` as in the previous revision, or link `KafkaCommitter#commit`.

MartijnVisser (approved):
Approving. This also fails on #314's current head, only record 0 is visible after the restore, so it pins the point you raised there. I'll merge this first and #314 rebases on it.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaSinkRecoveryITCase.java
Savonitar: I switched to `@link` because, with `@code`, IntelliJ treated the method name as plain text: I couldn’t navigate to it via hotkeys/mouse. My intention was to make the reference navigable. It is convenient if people read java docs "manually". Maybe a little bit out of dated approach in AI era though :-) 

linking to KafkaCommitter#commit sounds like a good alternative, agree, will fix it in a hotfix commit.

Savonitar:
Thanks for review and merge.

flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/KafkaSinkRecoveryITCase.java
Savonitar: @MartijnVisser I raised [PR](https://github.com/apache/flink-connector-kafka/pull/324) with your proposed improvement 👍  
Could you please take a look?

PR #323, opened 2026-09-18
[FLINK-40710][Connectors/Kafka] Preserve Dynamic Kafka partition ownership during reader recovery

bowenli86:

Fix [FLINK-40710](https://issues.apache.org/jira/browse/FLINK-40710): after checkpoint recovery redistributes Dynamic Kafka partitions, a reader can fail before the next checkpoint and report its former assignments. Reconciling those reports against pending assignments without preserving current ownership can assign one physical partition to multiple readers.

This complements the discovery-race fix in #305 by reconciling checkpoint reports with current reader ownership. It prevents recurrence on valid checkpoint state. It does not normalize checkpoints that already contain duplicate partition reports.

## Brief change log

- Preserve current physical-partition ownership during local reader recovery, including reactivated retained splits.
- Reconcile checkpoint reports before flushing assignments; replace pending copies by physical identity and preserve recovered offsets against late discovery results.
- Send bounded-source completion only after all restored assignments have been delivered.

## Verifying this change

- Reproduced against Apache `main` at `61dee320`: the same three regression classes run 41 cases with 31 assertion failures and no execution errors before the fix; all 41 pass with it.
- 72 focused tests pass on public Flink 2.2.1 / Java 17, including recovery, pending ownership, delayed discovery, bounded completion, serializers, proxy behavior and architecture rules.
- Dynamic recovery tests exercise the coordinator callback ordering with broker discovery stubbed. Delayed-discovery tests run the real offset initializer and delay its completion callback.
- `./mvnw clean verify`: 328 tests passed; 15 Docker-dependent setups errored because no valid Docker environment was detected. Full integration verification remains pending in CI.

- Dependencies: no.
- Public API, Table options or PyFlink wrappers: no.
- Checkpointed state, serializers or exactly-once delivery: recovery assignment behavior changes; checkpoint formats and serializers are unchanged.
- Per-record code paths: no.

## Documentation

- New feature: no; recovery correctness fix.
- Feature documentation: not applicable.
- English/Chinese documentation changes: not applicable.

---

- [X] Yes

Generated-by: OpenAI Codex (GPT-6)


flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/ReaderRecoveryGate.java
```diff
@@ -65,32 +55,31 @@ class ReaderRecoveryGate {
...
 
     ReaderRecoveryGate(boolean restoredFromCheckpoint) {
         this.initialReaderRegistrationPending = restoredFromCheckpoint;
     }
 
-    /** Records splits a reader reported on registration; an empty report is ignored. */
+    /** Records the current report during initial recovery, including an empty replacement. */
     void recordReportedSplits(int subtaskId, List<DynamicKafkaSourceSplit> reportedSplits) {
-        if (!reportedSplits.isEmpty()) {
+        if (initialReaderRegistrationPending) {
```
MartijnVisser, line 70: An empty re-registration inside the recovery window now wipes the earlier report. That is what a reader restoring from a checkpoint taken in that window sends, and its splits then stay marked assigned and are never handed out. Two-split restore: main assigns both, this branch none; test in a comment below. Please keep ignoring empty reports, as #295 does.

flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/dynamic/source/enumerator/DynamicKafkaSourceEnumerator.java
```diff
@@ -801,16 +828,28 @@ private boolean allReadersRegistered() {
...
     private void flushPendingSplitAssignmentsForRegisteredReaders() {
         List<Integer> registeredReaders = new ArrayList<>(enumContext.registeredReaders().keySet());
         Collections.sort(registeredReaders);
-        for (int readerId : registeredReaders) {
-            addReaderToClusterEnumerators(readerId);
+        addReadersToClusterEnumerators(registeredReaders);
+    }
+
+    private void addReadersToClusterEnumerators(List<Integer> readers) {
+        boolean previousFlushInProgress = flushingPendingSplitAssignments;
```
MartijnVisser, line 841: Same deferral as FLINK-39712's #300; after the rebase this part goes.

MartijnVisser (changes requested):
I have one blocker, see the in-line comment. I also think that we need to make sure we're doing things in the right order, so FLINK-40469's #295 first, then FLINK-39712's #300, then this rebased on both.

MartijnVisser:
The test behind the blocker, using the helpers this PR adds to `DynamicKafkaSourceEnumeratorRecoveryTest`. Restored state has two splits on `cluster-0`, parallelism 2:

```java
enumerator.start();
// all readers register before the first discovery completes; reader 0 reports both splits
registerReaderAndReportRestoredSplits(context, enumerator, 0, splits);
registerReaderAndReportRestoredSplits(context, enumerator, 1, Collections.emptyList());
assertThat(context.getSplitsAssignmentSequence()).isEmpty();
// a checkpoint completes in this window: reader state is empty, the sub-enumerator state
// still marks both splits assigned. Reader 0 fails, restores from it and re-registers empty.
context.unregisterReader(0);
enumerator.addSplitsBack(Collections.emptyList(), 0);
registerReaderAndReportRestoredSplits(context, enumerator, 0, Collections.emptyList());
context.runNextOneTimeCallable();
assertExactCheckpointOffsets(context, 0, splits);
```

Passes on main. On this branch the last assertion sees no assignment at all, and nothing rediscovers the two partitions because the sub-enumerator still holds them as assigned.

